PR #6, opened 2017-05-01
Rewrite to match suggestions from issues and feedback.

KarlSchimpf:
Cleans up the syntax of try blocks. Catch blocks are now part of try blocks and not nested.

Uses type section to define signatures of exceptions, with the view that the type signature is defining the constructor.

The exception section is defining exceptions (not constructors). Each entry defines the type signature of the   corresponding exception constructor.

### DIFF
--- a/proposals/Exceptions.md
+++ b/proposals/Exceptions.md
@@ -1,6 +1,6 @@
 # Exception handling
 
-There are three sections to this proposal
+There are two sections to this proposal
 
 1. A overview of the proposed extension to WebAssembly
 
@@ -18,13 +18,14 @@ may be an unknown exception thrown by an imported call.
 
 Thrown exceptions are handled as follows:
 
-a) They can be caught by a catch in an enclosing try block of a function body.
+a) They can be caught by a catch block in an enclosing try block of a function
+   body.
 
 b) Throws not caught within a function body continue up the call chain until an
    enclosing try block is found.
    
-c) If the call stack is exhausted without any enclosing try blocks, it terminates
-   the application.
+c) If the call stack is exhausted without any enclosing try blocks, it
+   terminates the application.
 
 This proposal looks at the changes needed to incorporate these concepts into the
 portable binary encoding of WebAssembly modules.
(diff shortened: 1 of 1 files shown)


proposals/Exceptions.md
```diff
@@ -240,19 +234,21 @@ If the `kind` is `Exception`:
 
 | Field | Type | Description |
 |-------|------|-------------|
-| type  | `varuint32` | exception index to table in the *exception type section* |
+| tag   | varuint32 | index into exception section |
+| type  | `varuint32` | type index of the function signature |
```
KarlSchimpf: We need to also add the exception tag as an additional field.

KarlSchimpf: Never mind. The 'tag' field is already there.

proposals/Exceptions.md
```diff
@@ -38,125 +39,136 @@ and the instructions *throw* and *rethrow*.
...
-a different exception type (they must have different exception type indices), and
-the exception types are defined in the exception type section of the module. It
-also checks that the block type of the catch blocks is the same as the enclosing
-try block. Further, it verifies that there aren't any instructions between the
-catch blocks of the try block, and the last catch block is at the end of the try
-block.
-
-### Throws
+Exception tags are used by throw and catch instructions. The throw instruction
+uses the tag to call the appropriate exception constructor, which gets values
```
rossberg: Nit: I would avoid OO terminology like "calling a constructor" -- there is nothing being called, it simply allocates a data structure and initializes it.

proposals/Exceptions.md
```diff
@@ -178,58 +190,40 @@ there to query.
...
+[binary encoding design document](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md).
 
 ### Exception type signatures
 
-An exception type is described by its exception type signature as follows:
-
-`except_type`
+An exception is described by its exception type signature, which corresponds to
+the function signature of the exception constructor. Hence, an exception _type
+signature_ is a `func_type` that returns zero values.
```
rossberg: I would prefer not to reuse function types for exceptions, since they don't really fit. Exceptions don't "execute" and they don't have a return type, not even an empty one. It rather is a tuple type.

I think it's fine to define exception types inline as simply a vector of value types, instead of an indirection through the type section.

proposals/Exceptions.md
```diff
@@ -240,19 +234,21 @@ If the `kind` is `Exception`:
 
 | Field | Type | Description |
 |-------|------|-------------|
-| type  | `varuint32` | exception index to table in the *exception type section* |
+| tag   | varuint32 | index into exception section |
```
rossberg: This still looks wrong. Imports are disjoint from locally defined entities, they neither can nor should reference the exception section. I think this field can just be eliminated.

proposals/Exceptions.md
```diff
@@ -38,125 +39,136 @@ and the instructions *throw* and *rethrow*.
...
-If the throw appears within a try block, the value stack is popped back to
-the size the value stack had when the try block was initially entered.
+### Throws
 
-If a throw occurs within a function body, and it doesn't have an enclosing try
-block, the throw continues up the call stack until either an enclosing try block
-is found, or the call stack is flushed. If the call stack is flushed, execution
-is terminated.
+The _throw_ instruction has a single immediate argument, an exception tag.  The
+exception tag is used to define the constructed exception. The arguments for the
```
KarlSchimpf: Reworded this sentence to "When an exception is thrown, the exception is allocated and the values on the stack (corresponding to the type signature) are popped off and assigned to the allocated exception.

proposals/Exceptions.md
```diff
@@ -38,125 +39,136 @@ and the instructions *throw* and *rethrow*.
...
+
+A throw inside a the body of a catch block is never caught by the corresponding
+try block of the catch block, since instructions in the body of the catch block
+are not in the body of the try block.
+
+Once a catching try block is found for the throw, the value stack is popped back
+to the size the value stack had when the try block was entered. Then, tagged
+catch blocks are tried in the order they appear in the catching try block, until
+one matches. A tagged catch block matches if the corresponding `catch`
+instruction is applicable, i.e. the corresponding exception was generated using
```
KarlSchimpf: Replaced this sentence with "Then, tagged catch blocks are tried in the order they appear in the catching try block, until one matches.

proposals/Exceptions.md
```diff
@@ -38,125 +39,136 @@ and the instructions *throw* and *rethrow*.
...
+the constructor defined by the tag of the `catch` instruction.
+
+If a matched tagged catch block is found, control is transferred to the body of
```
KarlSchimpf: Replaced this sentence with "If a matched tagged catch block is found, control is transferred to the body of the catch block, and the data fields of the exception are pushed back onto the stack."

proposals/Exceptions.md
KarlSchimpf: I agree. The type signature only defines the data fields of the exception.

proposals/Exceptions.md
KarlSchimpf: Clearly, we want exceptions to be imported, so that one module can throw it to another?

Separate entries in the exception section can have the same data fields, and still define different exceptions. For example, how could you tell if one is a C++ exception, and the other is not?

rossberg: Sorry if my comment wasn't clear, I was referring to the `tag` field specifically, which I can't make sense of. What would it be used for? A tag is the very thing that gets imported, so why a reference to another tag definition here? (Consider, for example, that you of course ought to be able to import exceptions without defining any locally -- that can't possibly work if the import somehow needs to refer to a local definition.)

KarlSchimpf: Without the tag, how does the catch instruction know to catch an imported exception? We need some way to refer to the imported exception in the catch!

KarlSchimpf: I think I understand now. Brad told me about function index spaces. If we use a similar concept for an "Exception index space", where imports appear first, we can remove the tag from the import.

KarlSchimpf: Removed the exception tag. Also defined an "exception index space" where imported exceptions must appear first. This makes imported exceptions behave similarly to function import.

KarlSchimpf (commented):
I've updated the proposal based on feedback. Please review.

proposals/Exceptions.md
```diff
@@ -84,7 +84,7 @@ else
 end
 ```
 
-A try block may contain one or more catch blocks, and all but the last catch
+A try block also contain one or more catch blocks, and all but the last catch
```
rossberg: Typo: contains

proposals/Exceptions.md
```diff
@@ -222,7 +222,19 @@ declares exception types using exception type signatures.
...
 |-------|------|-------------|
 | count | `varuint32` | count of the number of exceptions to follow |
-| sig | `except_type*` | The type signature of the data fields for the exception |
+| sig | `except_type*` | The type signature of the data fields for each exception |
+
+#### Exception index space
+
+The _exception index space_ indexes all imported and internally-defined
+exceptions, assigning monotonically-increasing indices based on the order
+defined in the exception section. Thus, the index space starts at zero with
```
rossberg: Nit: "in the import and exception section"

rossberg (commented):
Excellent, thanks! LGTM, modulo typo/nit.

Issue #25, opened 2017-08-14
Exception section position and identifier


eholk:
The proposal does not list a section id for the exception section. Based on [this change to V8](https://chromium-review.googlesource.com/c/591910/7/src/wasm/module-decoder.h#37), I assume this should be 13?

It looks like the exception section must come before the code section, so sections no longer come strictly in order (modulo custom sections). How precisely should we specify the location of the exception section?

titzer:
IMO the exception section should have a well-specified order, even though it might not match the ID. Current decoders were designed with this forward-looking assumption, since they simply do something like:

if (section_code == kFirstCode) {
  readFirstSection();
  section_code = next_code();
}
if (section_code == kSecondCode) {
  readSecondSection();
  section_code = next_code();
} 
...

The actual numerical values of the section codes were expected to get out of order.

aheejin:
Closed by #62 and #75.

PR #27, opened 2017-08-15
Clarify the ID and location of the exception section

eholk:
The exception is assigned the number 13 because this is the next available identifier. The [spec](https://webassembly.github.io/spec/binary/modules.html#sections) currently shows numbers 0 through 11 are assigned. Section 12 appears to be reserved for the name section.

The position of the exception section is driven by the requirement that items be defined before they are used. Having the exception section after the import section allows imported exceptions to be assigned the lower number tags. Similarly, defining the exceptions before the export section allows exceptions to be exported. Beyond these constraints, no additional requirements are imposed.

Issue #25

### DIFF
--- a/proposals/Exceptions.md
+++ b/proposals/Exceptions.md
@@ -307,13 +307,29 @@ or defined:
 ### Exception section
 
 The `exception` section is the named section 'exception' with id 13. The
-exception section must occur before the code section.  The exception section
-declares exception types using exception type signatures.
+exception section may occur anywhere after the import section and before the
+export section. There may be at most one exception section.  The exception
+section declares exception types using exception type signatures.
 
 | Field | Type | Description |
 |-------|------|-------------|
 | count | `varuint32` | count of the number of exceptions to follow |
 | sig | `except_type*` | The type signature of the data fields for each exception |
+
+#### Rationale
+
+The exception is assigned the number 13 because this is the next available
+identifier.
+The [spec](https://webassembly.github.io/spec/binary/modules.html#sections)
+currently shows numbers 0 through 11 are assigned. Section 12 appears to be
+reserved for the name section.
+
+The position of the exception section is driven by the requirement that items be
+defined before they are used. Having the exception section after the import
+section allows imported exceptions to be assigned the lower number
+tags. Similarly, defining the exceptions before the export section allows
+exceptions to be exported. Beyond these constraints, no additional requirements
+are imposed.
 
 ### Import section
 
--- a/proposals/Exceptions.md
+++ b/proposals/Exceptions.md
@@ -306,7 +306,8 @@ or defined:
 
 ### Exception section
 
-The `exception` section is the named section 'exception'. The exception section
+The `exception` section is the named section 'exception' with id 13. The
+exception section must occur before the code section.  The exception section
 declares exception types using exception type signatures.
 
 | Field | Type | Description |


KarlSchimpf (approved):

jfbastien:
What is the rationale for this ordering? It should be documented.

eholk:
What's the rationale for the other section positions? It sounds like the general principle is that things must be defined before they are used. The rest of the sections are required to come in order. Was that the only solution to the define-before-use constraint?

In this case, for the exception section needs to come before the code section. @KarlSchimpf says that imported exceptions get the lowest numbers indices, so that means exceptions would also have to come after the import section. I'm not sure why imported exceptions should have the lowest numbers though.

KarlSchimpf:
Putting imported exception ID's first follows the same conventions of other id namespaces. That is, the imported elements should come first. For example, the (module) function id namespace places imported function indices before functions appearing within the module.

jfbastien:
The commit message should contain such justification, and the design document for this proposal should as well. It doesn't seem like you strongly believe this to be the case though, am I right? I advise this to be solid before we discuss EH at an in-person meeting, I think such a discussion would take too long if it's not backed by a solid reasoning.

proposals/Exceptions.md
```diff
@@ -307,14 +307,30 @@ or defined:
...
 | count | `varuint32` | count of the number of exceptions to follow |
 | sig | `except_type*` | The type signature of the data fields for each exception |
 
+#### Rationale
+
+The exception is assigned the number 13 because this is the next available
+identifier.
+The [spec](https://webassembly.github.io/spec/binary/modules.html#sections)
+currently shows numbers 0 through 11 are assigned. Section 12 appears to be
+reserved for the name section.
```
binji, line 325: Where is that mentioned?

eholk: Where is what? The name section? I was going based on this line of code: https://cs.chromium.org/chromium/src/v8/src/wasm/module-decoder.h?type=cs&q=NameSection+f:v8&l=36

binji: Yeah, I meant where is the name sections specified as section 12. I'd say that v8 is wrong here -- the spec says nothing about the name section being numbered.

eholk:
I updated the description and added the rationale to the doc.

I don't have strong opinions about this, I just want to make sure the document is clear about what is expected. The define-before-use requirement and following existing conventions regarding numbering are both perfectly acceptable justifications to me.

Right now I haven't given a fixed location, which breaks with the convention for the rest of the sections. If we fix a location, that means we have to pick one and I don't see a clear criteria yet to do this. I'm open to ideas, since I may have missed something.

proposals/Exceptions.md
jfbastien: Name section is a custom section, numbered zero, and using an explicit string as name.

jfbastien:
The argument you're trying to make, I think, is that the Code section uses exception IDs, and therefore needs them to type check? And a similar argument about sections that need to come before?

proposals/Exceptions.md
```diff
@@ -306,14 +306,31 @@ or defined:
 
 ### Exception section
 
-The `exception` section is the named section 'exception'. The exception section
-declares exception types using exception type signatures.
+The `exception` section is the named section 'exception' with id 13. The
+exception section may occur anywhere after the import section and before the
```
rossberg, line 310: The order of all known sections is fixed. We shouldn't lose that property. For maximum future-proofness I suggest inserting it between import section (2) and function section (3).

aheejin:
Closed by #62 and #75.

PR #28, opened 2017-08-15
Initial support for exceptions in spec interpreter

eholk:
This change provides some initial support for exceptions in the spec interpreter.

We support basic exception declarations, `try` blocks, `catch_all` clauses and `throw`. This includes type checking, encoding, decoding, printing and execution.

There are still some limitations. Exceptions cannot yet take arguments. We do not support `catch` or `rethrow` yet. The allowed locations for the exceptions section is overly strict now, in part because the location is still an open question (#25).

Future PRs will add the missing functionality, but this is sufficient to give initial support and help clarify some of the details of the spec.

### DIFF
--- a/interpreter/spec/ast.ml
+++ b/interpreter/spec/ast.ml
@@ -96,6 +96,10 @@ and instr' =
   | Unary of unop                     (* unary numeric operator *)
   | Binary of binop                   (* binary numeric operator *)
   | Convert of cvtop                  (* conversion *)
+  | Throw of var                      (* throw exception *)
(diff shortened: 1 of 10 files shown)


eholk:
@rossberg-chromium - Could you take a look at this? Thanks!

eholk:
One clarification: this change requires `try` blocks to have a `catch_all` clause. The reason is that `catch` clauses are not supported yet. Once `catch` is added, `try` blocks will be required to have at least one `catch` clause or a `catch_all`.

aheejin:
This is for the old spec. Superceded by #151 and #160.

Issue #30, opened 2017-10-14
Problem of rethrow instruction


aheejin:
## Problem

The current `rethrow` instruction has some problems that make its use very
limited. For example, suppose there are two try/catch blocks and in both catch
they jump to some common code, which is followed by a `rethrow` instruction.
```
block $label0
  try
    ...
  catch i
    br $label$0
  end
  ...
  try
    ...
  catch i
    br $label$0
  end
end

some common code
rethrow
```

This cannot be supported in the current spec because `rethrow` can occur only in
the scope of a `catch` block. But this code pattern is very common especially
when there is some cleanup code (calling destructors) to run before rethrowing
an exception up to a caller. If you compile the code below,
```
MyClass obj;
try {
  foo(); // might throw
} catch (int n) {
  ...
}
...
try {
  foo(); // might throw
} catch (int n) {
  ...
}
```

The generated code will look like this:
```
block $label0
  try
    call $foo
  catch i
    br $label$0
  end
  ...
  try
    ...
  catch i
    br $label$0
  end
end

call MyClass's destructor to destruct obj
rethrow
```
In this case, cleanup action consists of only a single destructor call, but it
can take more code space if there are many objects to destroy, so I think
duplicating this cleanup code into possibly `n` catch blocks is not a viable
idea. This is a classic case in which we need a `rethrow`, but it cannot be
executed at the end of this code because it's not in the scope of a `catch`.

This is one example of code sharing that's very common, but code sharing between
catch blocks can occur in other cases as well. You can use `goto` within catch
clauses to jump to a common block. Or any middle-IR-level compiler optimization
pass can factor out some common code in catch blocks.

While this problem can be worked around using not a `rethrow` but just a normal
`throw`, `throw` is considered as throwing a completely new exception, and the
VM wouldn't be able to carry an attached backtrace with it, which can be useful
when we later support backtrace debugging in the future. And more importantly,
this problem effectively makes `rethrow` unusable at all, because the most
common usecase of it is, as illustrated above, when it occurs after some common
cleanup code which is shared between many catch blocks. It can also occur when
there is shared cleanup code between `catch i` and `catch_all` clauses, which
will be very common case as well, but I'm actually planning on proposing
something else for `catch` clauses... but anyway.

## Idea?

This is a rough idea and not a complete spec yet. And it's not I'm proposing
this as a single concrete alternative and I appreciate comments and suggestions.

I think it is necessary to make it possible to access some kind of handle to an
[...]

PR #33, opened 2018-02-09
Initial draft of the layer 1.

KarlSchimpf:


proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+concepts and structures. These concept and structures are introduced in later
+layers, and either:
+
+1. Improves readability by combining concepts in layer 1 into higher-level
+   constructs, thereby reducing code size.
+
+2. Allow performance improvements in the VM.
+
+3. Introduce additional new functionality not available in layer 1.
+
```
eholk, line 14: We have a list here of what the proposal does not include, but it'd be good to include the requirements for the MVP. To me, these include:
1. Sufficient to implement C++ exception semantics by translating from Clang-generated LLVM code.
2. Allows WebAssembly to throw exceptions that can be caught by JavaScript.
3. Allows WebAssembly to catch exceptions thrown by JavaScript.
4. Can be implemented with low performance overhead.

The performance goal could be made more precise. I think it's roughly "code that does not use exceptions does not run any slower once the exception proposal has been adopted."

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
+# Layer 1 exception handling
```
eholk, line 1: Is this meant to supersede the existing [exceptions proposal](https://github.com/WebAssembly/exception-handling/blob/master/proposals/Exceptions.md)?

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+3. Introduce additional new functionality not available in layer 1.
+   
+## Overview
+
+Exception handling allows code to break control flow when an exception is
+thrown.  The exeception can be any exception known by the WebAssembly module, or
+it may an unknown exception that was thrown by a called imported function.
+
+One of the problems with exception handling is that both WebAssembly and the
+host VM probably have different notions of what exceptions are, but both must be
```
eholk, line 22: nit: I'd probably say "embedder" rather than "host VM" here and elsewhere.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+aware of the other.
+
+It is difficult to define exceptions in WebAssembly because (in general)
+it doesn't have knowledge of the host VM. Further, adding such knowledge to
+WebAssembly would limit the ability for other host VMs to support WebAssembly
+exceptions.
```
eholk, line 28: For the first "host VM" here, it feels like you're talking about the JavaScript VM. Otherwise, it seems like adding exceptions to other host VMs would be no more limited than the host VM.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+
+One issue is that both sides need to know if an exception was thrown by the
+other, because cleanup may need to be performed.
+
+Another problem is that WebAssembly doesn't have direct access to the host VM's
+memory.  As a result, WebAssembly defers the handling of exceptions to the host
+VM.
+
+To access exceptions, WebAssembly provides instructions to check if the
+exception is one that WebAssembly understands. If so, the data of the
```
eholk, line 38: It would probably be helpful to make a distinction between "WebAssembly, the spec" and "WebAssembly, the program that is running embedded in a host VM." For the former, I'd suggest just WebAssembly, and for the latter I'd suggest "user code" or "user WebAssembly code."

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+exceptions.  For host VMs that have garbage collection (such as JavaScript),
+this is not a problem.
+
+However, not all host VMs may have a garbage collector. For this reason,
+WebAssembly exceptions are designed to allow the use of reference counters to
+perform the the garbage collection in the host VM.
+
+To do this, WebAssembly exceptions are immutable once created, to avoid cyclic
+data structures that can't be garbage collected. It also means that exceptions
+can't be stored into linear memory. The rationale for this is twofold:
```
eholk, line 57: The reason that exceptions can't be stored in linear memory is that they need to be abstract, whereas if you could store them in linear memory you could trivially load them under a different type to get access to the raw bits and underlying representation.

Although, I realize if I'd just read a little further I'd see that you addressed this :smile:

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+  stored. This allows spoofing of exception references that may allow a
+  WebAssembly module to access data it should not know in the host VM.
+  
+* The host VM does not know the layout of data in linear memory, so it can't
+  find places where exception references are stored.
+
+Hence, while an exception reference is a new first class type, this proposal
+disallows their usage in linear memory.
+
+A WebAssembly exception is created by the `except` instruction. Once an
```
eholk, line 70: We can bikeshed the names later, but we should consider something like `create_exception` or `exception.create` too.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+
+1. They can be caught by a catch block in an enclosing try block of a function
+   body. The caught exception is pushed onto the stack.
+
+1. Throws not caught within a function body continue up the call stack, popping
+   call frames, until an enclosing try block is found.
+   
+1. If the call stack is exhausted without any enclosing try blocks, the host VM
+   defines how to handle the uncaught exception.
+
```
eholk, line 82: Do we need to mention something here about interleaving host and guest frames?

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+of values associated with the exception.
+
+Within the module, exception types are identified by an index into the
+[exception index space](#exception-index-space). This index is referred to as
+the `exception tag`. The `tagged exception type` is the corresponding
+exception type refered to by the exception tag.
+
+Exception types can be imported and exported by adding the appropriate entries
+to the import and export sections of the module. All imported/exported exception
+types must be named to reconcile exception tags between modules.
```
eholk, line 100: This is the same as it works for imported and exported functions, right?

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+
+1. The `except` instruction which creates a WebAssembly instance of the
+   corresponding tagged exception type, and pushes a reference to it onto the
+   stack.
+
+2. The `if_except` instruction that queries an exception to see if it is an
+   instance of the corresponding tagged exception class, and if true it pushes
+   the corresponding values of the exception onto the stack.
+   
+### The exception refernce data type
```
eholk, line 112: refernce -> reference

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+to the import and export sections of the module. All imported/exported exception
+types must be named to reconcile exception tags between modules.
+
+Exception tags are used by:
+
+1. The `except` instruction which creates a WebAssembly instance of the
+   corresponding tagged exception type, and pushes a reference to it onto the
+   stack.
+
+2. The `if_except` instruction that queries an exception to see if it is an
```
eholk, line 108: I like the name `if_except`. What about `exception.match` if we decide to go with `exception.*` prefixes? Although we should probably save this discussion for a future bikeshedding period.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+   instance of the corresponding tagged exception class, and if true it pushes
+   the corresponding values of the exception onto the stack.
+   
+### The exception refernce data type
+
+Data types are extended to have a new `except_ref` type. The representation of
+an exception type is left to the host VM, but its size must be fixed. The actual
```
eholk, line 115: I think whether the size is fixed can be left up to the host VM. One could imagine a representation where they have a length field followed by that number of bytes. In practice, these are always going to be word-sized pointers though.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+   constructs, thereby reducing code size.
+
+2. Allow performance improvements in the VM.
+
+3. Introduce additional new functionality not available in layer 1.
+   
+## Overview
+
+Exception handling allows code to break control flow when an exception is
+thrown.  The exeception can be any exception known by the WebAssembly module, or
```
aardappel, line 18: typo: exeception

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+
+A try block ends with a `catch block` that is defined by the list of
+instructions after the `catch` instruction.
+
+Try blocks, like control-flow blocks, have a _block type_. The block type of a
+try block defines the values yielded by the evaluation the try block when either
+no exception is thrown, or the exception is successfully caught by the catch
+block.
+
+In the initial implementation, try blocks may only yield 0 or 1 values.
```
eholk, line 142: It sounds like we may get multi-value blocks pretty soon, so we don't we just do whatever the rest of Wasm does?

proposals/Layer-1.md
aardappel, line 112: typo: refernce

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+
+### Debugging
+
+Earlier discussion implied that when an exception is thrown, the runtime will
+pop the operand stack across function calls until a corresponding, enclosing try
+block is found. The implementation may actually not do this. Rather, it may
+first search up the call stack to see if there is an enclosing try. If none are
+found, it could terminate the thread at the point of the throw. This would
+allow better debugging capability, since the corresponding call stack is still
+there to query.
```
eholk, line 231: Alternatively, the embedder can capture a stack trace at the point the exception object is created, which I think is roughly what JavaScript will do.

One question is whether stack traces should be captured at the point of creating the exception, or at the point of throwing the exception. I think this is irrelevant to the core spec, since stack traces are provided purely by the embedder and we provide no way to access them in WebAssembly. It's worth discussing in the context of the JS embedding though.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+
+### Exception index space
+
+The _exception index space_ indexes all imported and internally-defined
+exceptions, assigning monotonically-increasing indices based on the order
+defined in the import and exception sections. Thus, the index space starts at
+zero with imported exceptions followed by internally-defined exceptions in
+the [exception section](#exception-section).
+
+## Changes to the binary model
```
eholk, line 271: binery model -> binary encoding

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+   defines how to handle the uncaught exception.
+
+### Exceptions
+
+An `exception` is an internal construct in WebAssembly that is maintained by the
+host. WebAssembly exceptions (as opposed to host exceptions) are defined by a
+new `exception section` of a WebAssembly module. The exception section is a list
+of exception types, from which exceptions can be created.
+
+Each exception type has a `type signature`. The type signature defines the list
```
aardappel, line 90: Playing advocate of the devil here: since in the general case a C++ exception type cannot be represented as a self-contained wasm signature (e.g. if it contains a string) I'd expect the front-end to often or always opt for a signature that is just a single linear memory index. Similarly, a host exception will likely often be a single anyref, once we get that. This in turn does not explain to me why we need the generality of a type signature, other than, "why not, may be nice to have".

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+#### except_ref
+
+An exception reference pointing to an instance of an exception. The size
+is fixed, but unknown in WebAssembly (the host defines the size in bytes).
+
+### Language Types
+
+| Opcode | Type constructor |
+|--------|------------------|
+| -0x41  |  `except_ref`    |
```
eholk, line 289: I'd make this `0x7B`, since `except_ref` feels more like a [value type](https://webassembly.github.io/spec/core/binary/types.html#value-types) to me.

proposals/Layer-1.md
aardappel, line 100: What if my signature is simple, e.g. a single i32, does it mean that if I try to catch exceptions of this type, I will also catch exceptions from other modules that are unrelated? Do exceptions need some form of identifier to make them unique?

eholk (approved):
Thanks for writing this up! I left quite a few comments, but I'm going to go ahead and merge this as is. We can address the comments in later PRs. Some will need more in-depth discussion, which is probably best done via issues.

proposals/Layer-1.md
eholk: I opened #34 to track this.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+
+If the selected catch block does not throw an exception, it must yield the
+value(s) expected by the corresponding catching try block. This includes popping
+the caught exception.
+
+Note that a caught exception can be rethrown using the `throw` instruction.
+
+### Exception data extraction
+
+The `if_except block` defines a conditional query of the exception on top of the
```
aardappel, line 190: This adds another type of control-flow block specifically for this purpose, which has an implementation cost. What alternatives have been considered, e.g. an instruction that loads the type or Nth element from an exception object, such that a the existing if block can be used to inspect it?

proposals/Layer-1.md
eholk: Tracking in #35

proposals/Layer-1.md
eholk: Tracking in #35

proposals/Layer-1.md
eholk: Added #36 to discuss this relationship.

proposals/Layer-1.md
eholk: Discuss in #37

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+
+To access exceptions, WebAssembly provides instructions to check if the
+exception is one that WebAssembly understands. If so, the data of the
+WebAssembly exceptions's data is extracted and copied onto the stack, allowing
+succeeding instructions to process the data.
+
+Lastly, exception lifetimes must be maintained by the host VM, so that it can
+collect and reuse the memory used by exceptions. This implies that the host must
+know where exceptions are stored, so that it can determine when an exception can
+be garbage collected.
```
aardappel, line 45: If we had more strict structure around throw/rethrow, in theory at least exceptions could have static life-times. Requiring them to be GC/refcounted seems necessary specifically to accomodate hosts that want to manage their own exceptions in a more dynamic fashion.

proposals/Layer-1.md
KarlSchimpf: I added this constraint because local variables/global variables may be allocated in a linear memory in the embedder, and hence would require that.

proposals/Layer-1.md
KarlSchimpf: I don't disagree with this. I only wanted to state that it is up to the host on how to best handle this, and, depending on whether it is compiled for debugging or release, do slightly different implementations.

proposals/Layer-1.md
KarlSchimpf: yes.

proposals/Layer-1.md
KarlSchimpf: Because an except_ref can hold any type of exception, it should have a common fixed size.

proposals/Layer-1.md
KarlSchimpf: Its more of the latter. This topic was brought up in many earlier CG meetings and the general consensus was that exceptions should not just be C++ specific, but more general so that other languages could be translated down to WebAssembly.

proposals/Layer-1.md
eholk: I think most reasonable implementations will have `except_ref` be a fixed size, but this is ultimately a question of representation. We don't expose any details of the representation to Wasm programs, so I think we should not make any requirements on what representations are allowed for the embedder provided they can otherwise implement the Wasm semantics.

proposals/Layer-1.md
eholk: Sounds good.

proposals/Layer-1.md
eholk: Each exception type has an opaque tag associated with it, so if my module declared `$exception1` at index 1 as type `i32` and `$exception2` at index 2 as type `i32`, these are different types. So, an `if_except 1` block would not match anything of type `$exception2` and vice-versa.

Technically, I think these can alias, if they are both imported exceptions. The host may supply the same underlying exception to both imports, the same way that one could import the same function under two different names.

proposals/Layer-1.md
KarlSchimpf: I agree that there is an implementation cost to the current proposal. On the other hand, there
may be implementation cost of loading elements individually - I.e. does the element loader need to verify the access is correct on each element lookup.

In the current vacuum of realistic examples, it is hard to say which is better, so one was chosen, based on previous discussions in the original proposal (in Exceptions.md).

eholk: I think the main advantage of the approach we've presented here is the guarantees provided by the type system. Let's say we had a `get_type_index` instruction and a `get_exception_field` instruction, we'd end up with code like:

```
if get_type_index == 0:
  get_exception_field 0, 0
else:
  ...
```

The problem is that at `get_exception_field`, we don't statically know the exception has the correct type (here it's pretty obvious because it shows up immediately after a type check, but it may not always be that easy to see).  This means we have to insert a dynamic type check on top of the one we just generated.

Instead, the combined match and unpack that we have with `if_except` eliminates the need to insert any dynamic check beyond what the programmer has written already.

proposals/Layer-1.md
KarlSchimpf: Remember, not all exception lifetimes are controlled by WebAssembly code. Exceptions thrown by the host must follow the lifetime expectation of the host. Hence, using static lifetimes limit this use case.

eholk: The biggest problem here is that we are trying to generate Wasm code from LLVM. LLVM exception code has been lowered to the point that it now longer has the usual exception structure (for example, landing pads may be combined, there's cleanup code to handle, etc.). With more structured Wasm instructions, we can probably implement C++'s semantics, but the tricky part is to be able to rethrow a foreign (i.e. JavaScript) exception once C++ is done with its handling code.

I and others have advocated a couple of approaches that give us statically known exception lifetimes, but unfortunately these just haven't been workable for generating code from LLVM inputs. I'd love to be proven wrong, but this seems like a surprisingly hard problem.

proposals/Layer-1.md
aheejin: If we are going to do the two-phase unwinding, this may have to be incorporated into wasm spec, because the toolchain (more precisely, the library like libcxxabi) has to support this behavior.

For example, for x86 EH, in the first phase, libcxxabi does not actually unwind the stack but just walks up the stack to see if there is a matching handler. If a matching handler is found, in the second phase it actually starts unwinding the stack until it reaches the handler found. If no matching handler is found, libcxxabi does not unwind the stack but just crashes at the point of throwing, leaving the whole stack trace intact.

I'm not suggesting we do this now, and for Layer 1 I think we don't need to do this anyway. I'm just suggesting if we do this later, this can't be just left to the discretion of the host VM but explicitly needs to be specified in the spec.

proposals/Layer-1.md
aheejin: This is a hard problem that eventually made us introduce the first-class exception object. tl;dr: CFG does not have enough original C++ scope information. While it may not be impossible we preserve that information from clang and carry that all the way to the backend, we kind of reached a conclusion that it is non-trivial and has the possibility of hindering optimization. #30 and #31 discussed this problem.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+This section describes change in the
+[Modules document](https://github.com/WebAssembly/design/blob/master/Modules.md).
+
+### Exception index space
+
+The _exception index space_ indexes all imported and internally-defined
+exceptions, assigning monotonically-increasing indices based on the order
+defined in the import and exception sections. Thus, the index space starts at
+zero with imported exceptions followed by internally-defined exceptions in
+the [exception section](#exception-section).
```
aheejin, line 269: 1. Are 'exception index' and 'exception tag' the same thing?
2. If the indices of imported exceptions start from 0, how does a wasm compiler know which tag to use to denote, let's say, C++ exception? Is it the linker's job to reconcile all the indices and fix all indices within a wasm program?

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+| --------- | ---- | ----------- |
+| [Function](#function-names) | `1` | Assigns names to functions |
+| [Local](#local-names) | `2` | Assigns names to locals in functions |
+| [Exception](#exception-names) | `3` | Assigns names to exception types |
+
+###### Exception names
+
+The exception names subsection is a `name_map` which assigns names to a subset
+of the _exception_ indices from the exception section. (Used for both imports
+and module-defined).
```
aheejin, line 374: I can be mistaken, but can there be any debug info attached to not an actual exception object but an exception tag?

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+
+### Control flow operators
+
+The control flow operators are extended to define try blocks, catch blocks,
+throws, and rethrows as follows:
+
+| Name | Opcode | Immediates | Description |
+| ---- | ---- | ---- | ---- |
+| `try` | `0x06` | sig : `block_type` | begins a block which can handle thrown exceptions |
+| `catch` | `0x07` | | begins the catch block of the try block |
```
aheejin, line 384: So this is supposed to be `catch_all`? If we are gonna add static the `catch tag` instruction later, shouldn't this be `catch_all` to make them different?

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+Try blocks, like control-flow blocks, have a _block type_. The block type of a
+try block defines the values yielded by the evaluation the try block when either
+no exception is thrown, or the exception is successfully caught by the catch
+block.
+
+In the initial implementation, try blocks may only yield 0 or 1 values.
+
+### Exception creation
+
+A `except` instruction has a single immediate argument, an exception tag.  The
```
rossberg, line 146: Separating out exception value creation from throwing is likely to make throw less efficient, because the compiler does not statically know what it is throwing. In the previous design, the compiler statically knew where to read off the handler to jump to (each exn could have a thread-local handler stack attached to it); with the separation it probably requires a hash table lookup.

The separation will also cause problems with future extensions. In particular, when you consider resumption, the type of `throw` will depend on the return type of the exception. That would only work if the `exn_ref` type you propose tracked that result type.

proposals/Layer-1.md
rossberg: Well, anyfunc will become a value type as well eventually. I think it makes sense to put all the reference types close together.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+#### Other Types
+
+##### except_type
+
+An exception is described by its exception type signature, which corresponds to
+the data fields of the exception.
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `count` | `varuint32` | The number of types in the signature |
```
rossberg, line 305: To allow for future extension (esp resumption), we should change this in two ways:

1. Have the `type` be defined by a function type (denoted by an index into the type section).

2. Leave space for an additional reserved "attribute" byte that we require to be zero in the MVP (a hook to allow as adding more information in the future).

(I apologise that I steered us away from 1 before, which is more similar to what @KarlSchimpf has suggested long ago. I didn't consider resumption at the time.)

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+* `2` indicating a `Memory` [import](Modules.md#imports) or [definition](Modules.md#linear-memory-section)
+* `3` indicating a `Global` [import](Modules.md#imports) or [definition](Modules.md#global-section)
+* `4` indicating an `Exception` [import](#import-section) or [definition](#exception-sectio)
+
+### Module structure
+
+#### High-level structure
+
+A new `exception` section is introduced and is named `exception`. If included,
+it must appear between the `Export` and `Start` sections of the module.
```
rossberg, line 325: Why after export? Since exceptions themselves can be exported it would be natural that it appears before the export section.

proposals/Layer-1.md
rossberg: Agreed with @aheejin. OTOH, we may not even need to distinguish the two. Instead, catch has a list of exception ids that defines what it catches. In the degenerate case that it is empty it is a catch-all. In the MVP, we could then require it to be empty (although I don't necessarily think we should).

proposals/Layer-1.md
eholk: > 1. Are 'exception index' and 'exception tag' the same thing?

No, but they're similar. The exception tag is an opaque identifier for the type of the exception. The exception index points to an entry in the exception table, which has its own tag.

> 2. If the indices of imported exceptions start from 0, how does a wasm compiler know which tag to use to denote, let's say, C++ exception? Is it the linker's job to reconcile all the indices and fix all indices within a wasm program?

By wasm compiler, do you mean something like the C++ to Wasm compiler (as opposed to the Wasm to Native Code VM)? If so, then the linker seems like a reasonable place to reconcile all the exception indices. Ultimately, the notion of "the C++ exception type" does not really exist in the Wasm spec, but is a convention used by the tools.

proposals/Layer-1.md
eholk: You could attach a name to an exception tag, which could count as very basic debug info, but in general I'd say no. Runtime info like a stack trace can only be attached to an exception object.

proposals/Layer-1.md
eholk: So to make sure I understand the idea, you mean have something like:
```
try:
  ...
catch 0, 1:
  ...
catch 2:
  ...
catch ε:
  ...
```

I kind of like that idea (although we'd probably want to require disjoint exception indices for each catch clause, and aliasing might be a problem).

Or did you mean have a single catch clause overall?

The reason for allowing only catch all for MVP is that our main consumer, C++, doesn't really have any use for non-catch_all. I think it makes sense to not spec and implement a feature until we have users for it. Are there other languages that are interested in being among the first exception users?

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
+# Layer 1 exception handling
+
+Layer 1 of exception handling is the MVP (minimal viable proposal) for
```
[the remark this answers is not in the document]
eholk: I'll follow up with a PR to clarify this. Sort of like we did in the host bindings discussion, we realized we can separate out some of the aspects of exception handling into what's absolutely necessary to support C++, and things that are useful in that other languages might use them or they might allow VMs to do a more efficient implementation.

Part of the goal in separating these out (and documenting them, I'll follow up with a document listing the extensions today), is to make sure our MVP is future-proof wrt such extensions.

In the interest of keeping the spec small and growing conservatively, my default would be to get the exception MVP into the spec and then follow up with the extensions as needed. I'm open to arguments for including the extensions as well though.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+  WebAssembly module to access data it should not know in the host VM.
+  
+* The host VM does not know the layout of data in linear memory, so it can't
+  find places where exception references are stored.
+
+Hence, while an exception reference is a new first class type, this proposal
+disallows their usage in linear memory.
+
+A WebAssembly exception is created by the `except` instruction. Once an
+exception is created, you can throw it with the `throw` instruction. Thrown
```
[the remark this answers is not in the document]
eholk: We went back and forth on this and couldn't see a clear reason to pick one over the other. I find your resumption argument compelling though.

proposals/Layer-1.md
rossberg: Yes, I meant still just a single exception clause, but annotated with an (optional) list of exception ids. That should give us the best of both worlds: simple instruction structure and efficient implementation.

Why does C++ not have use for specific catch? I would assume that any catch that is not a `catch(...)` in C++ could be using this. Or does LLVM make that impossible? Either way, other languages would probably only catch specific exceptions most of the time.

proposals/Layer-1.md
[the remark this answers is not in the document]
rossberg: Fully agreed. It was mainly the term "layer" and some of the explanation that made it sound like other things would be higher-level abstractions being added *on top*, which would be a quite different scenario.

proposals/Layer-1.md
eholk: I think we'd still want to be able to associate different code with different catch clauses. For example, say my language didn't follow the one exception type per language convention, but instead had Wasm exceptions for things like `FileNotFoundException`, `NetworkTimeoutException`, etc. I'd want to be able to handle these differently. Although there is a workaround of just nesting try blocks, or dynamically inspecting the exception inside the catch clause.

The issue with C++ and specific catch was discussed some in #31. @aheejin can probably explain the issues better but I think it comes down to a combination of LLVM's exception handling being very unstructured, and needing `catch(...)` to apply to both C++ exception and JavaScript exceptions. Basically, Clang/LLVM generates a big chunk of code that checks if an exception is known to C++ or not and does the appropriate thing. It sounds like to use specific catch we'd basically have to duplicate the exception handling code in the catch_all branch, and even finding which code to duplicate is non-trivial.

rossberg: Differentiating clauses only really makes sense if you also unpack the payload data, because the code will probably want to use that (and if you have to do that separately anyway then you can as well distinguish code paths there) -- which would bring us back to the earlier design without a separate match instruction. I'm fine with that as well, but the separation has the appeal that it simplifies instruction structure and is a better fit for the problems caused by LLVM.

For C++, I assume there isn't much win in compiling `catch(E e)` more efficiently because the handler stack is littered with implicit catch-all anyway, because destructors? Because if it wasn't then I wouldn't see why you couldn't use `catch $C++exn` to implement `catch(E e)` more efficiently -- that isn't concerned with foreign exceptions.

eholk: > Differentiating clauses only really makes sense if you also unpack the payload data, because the code will probably want to use that (and if you have to do that separately anyway then you can as well distinguish code paths there) -- which would bring us back to the earlier design without a separate match instruction.

Ah, right. I had been assuming that the specific catch form would also unpack the exception data. It seems like this would let the embedder implement exception handling without materializing any kind of exception object if it is not used, which seems like a likely performance win.

> For C++, I assume there isn't much win in compiling `catch(E e)` more efficiently because the handler stack is littered with implicit catch-all anyway, because destructors? Because if it wasn't then I wouldn't see why you couldn't use `catch $C++exn` to implement `catch(E e)` more efficiently -- that isn't concerned with foreign exceptions.

Yes, pretty much everything gets surrounded by a catch-all to run destructors. The problem with `catch(E e)` is that it also catches any subclass of `E`, meaning libcxxabi has to look at the RTTI to find out whether it can handle an exception.

rossberg: I agree that a combined catch/match makes certain optimisations easier, but I'm not sure how big a deal it is to do a data flow analysis instead. OTOH, the separation allows for more flexible control flow, which I thought was the motivation?

Subtyping on C++ exns seems like an orthogonal issue -- they'd still all have the same signle exn tag, don't they? But I can see how the destructor business makes this mostly a moot point in C++.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+
+```
+try block_type
+  instruction*
+catch
+  instruction*
+end
+```
+
+A try block ends with a `catch block` that is defined by the list of
```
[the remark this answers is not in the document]
KarlSchimpf: Actually, I expected layer 2 to provide a clean static model like that of the original proposal. In that context, no garbage collection is needed because the structure defines the lifetime of the exceptions, and hence it can clean up when each catch is exitted.

proposals/Layer-1.md
KarlSchimpf: Good point. They should be moved before exports, and maybe before imports.

rossberg: Why before imports? ;) That would screw up the order of the index space.

KarlSchimpf: My bad. I didn't realize that a section id number was defined for an exception section.

proposals/Layer-1.md
eholk: The goal is to get both. Keep the dynamic exception matching for cases where you need to more flexible control flow, but allow the more efficient static version when it's sufficient. I'm not sure how valuable this is though, and the complexity is significantly higher to support both in the spec. As you point out, a dataflow analysis would work, and I expect this would be a local analysis making it cheap to do (you wouldn't have to examine the whole program before you start compiling).

I see what you mean about C++ exceptions. If there are no destructors, then there's no reason to handle JavaScript exceptions at all unless there's a `catch(...)`. I'm pretty sure there are basically always destructors though, probably even more so if you're using exceptions.

proposals/Layer-1.md
aheejin: > The exception tag is an opaque identifier for the type of the exception. The exception index points to an entry in the exception table, which has its own tag.

Then an `except` instruction take an exception index as an immediate, right? In the Binary encoding section, it says `tag`, so it was confusing.

proposals/Layer-1.md
aheejin: Then what can the capability of attaching a name to an exception index be used for?

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+   corresponding tagged exception type, and pushes a reference to it onto the
+   stack.
+
+2. The `if_except` instruction that queries an exception to see if it is an
+   instance of the corresponding tagged exception class, and if true it pushes
+   the corresponding values of the exception onto the stack.
+   
+### The exception refernce data type
+
+Data types are extended to have a new `except_ref` type. The representation of
```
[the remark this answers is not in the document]
aheejin: In the current doc, `match` extracts wasm value(s) that were thrown, according to its type signature, from `exn_ref`. Then do you think `match` has to do the two-step extraction, `exn_package` -> `exn_ref` -> wasm values?

proposals/Layer-1.md
aheejin: I think the idea of `catch` instruction that takes variable number of tags (or indices), as @rossberg suggested, haven't been discussed yet. Is it possible for an instruction to take variable number of immediate values?

proposals/Layer-1.md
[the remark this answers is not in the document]
aheejin: @rossberg I'm not sure if I understand what you mean by the efficiency and the maximum cost of a catch-all. I think the reason we separated those two is because C++ is going to only use `catch_all`, so we need a dynamic type checking within the `catch_all`.

@KarlSchimpf So, you mean, we can use `match` and assign an exception object to a local within `catch_all` but not `catch tag` instructions that are going to be introduced in layer 2? I didn't know that... I just thought `catch tag` eliminates the need for dynamic matching in case we can separate code blocks for each catch clause, which is not feasible for C++. To me the idea that an exception object is prevented from being assigned to a local within a certain code block but not others seems a little weird. Then can we assign exception objects to locals outside of catch blocks? (not within `catch_all` or `catch tag`)

proposals/Layer-1.md
eholk: It'd be similar to attaching a name to a function. Without names, Wasm call stacks basically look like "wasm_function_0, wasm_function_1, wasm_function_1, ...." In the same way, names on exceptions would let devtools say something like "Unhandled exception: foo" rather than "Unhandled Exception: my_module.wasm:0" or worse, "Unhandled exception: <wasm exception 0x4ea2df>".

proposals/Layer-1.md
eholk: I can't think of any instructions that currently do this, but there's nothing fundamentally impossible about it. Rather than encoding it as `catch exception_index`, we'd just encode it as `catch vec(exception_index)`, similar to how [function types](https://webassembly.github.io/spec/core/binary/types.html#function-types) are encoded.

proposals/Layer-1.md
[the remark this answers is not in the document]
rossberg: Yes, `match` (I would probably name it `handle`) would take an exn_package and extract the wasm values if the tag matches. I would actually propose getting rid of exn_ref entirely by merging back except and throw.

proposals/Layer-1.md
aheejin: What you describe is we attach debug info to exception objects. But if we assign names to exception indices, that doesn't mean individual exception objects can have separate names, no?

proposals/Layer-1.md
rossberg: See `br_table`, which has  vec(labelidx) as immediate.

proposals/Layer-1.md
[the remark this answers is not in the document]
eholk: I'd been envisioning this as that from Wasm's perspective, there is only `exn_ref`. This points to an exception of a certain type with associated values specified by the exception's type. Then, embedders are welcome to wrap this in additional data, making a full `exn_package` that has embedder-specific data such as stack traces. Since `exn_ref` is opaque to Wasm, I don't think it particularly matters what else the embedder does with it, right?

That said, I could see value in separating these things out, in case we wanted to use tagged values as first class data separate from exception handling (which someone will probably try to do anyway).

proposals/Layer-1.md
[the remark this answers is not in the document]
rossberg: @aheejin, yes, but C++ is an extremely degenerate case in that regard. No other language works that way. Usually you want to be able to jump to the responsible handler directly, because that is cheaper than executing various intermediate handlers only to discover that they are not relevant for the given exception.

proposals/Layer-1.md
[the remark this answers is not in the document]
eholk: @rossberg - oops, we raced. I think we're in agreement in spirit, just using different terms. In my thinking, you do `handle` on the thing that is caught (an `exn_package`), and it gives you the values associated with the exception if it matches).

proposals/Layer-1.md
eholk: Correct, this would just be a name for the exception type, not the individual object.

proposals/Layer-1.md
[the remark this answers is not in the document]
rossberg: @eholk, a plain exn value makes no sense as an argument for e.g. rethrow, so you'd still want the distinction. They are really different things.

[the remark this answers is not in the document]
aheejin: I'm getting confused. What is then the first-class wasm type? `exn_ref` or `exn_package`? Would the composition of `exn_package` object be the same for all embedders?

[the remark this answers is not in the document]
rossberg: aheejin, if we did separate exception construction and throwing like in the current state of the proposal then both would be first-class. How either is represented is likely different for different embedders.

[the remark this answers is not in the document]
eholk: @rossberg, if we combine `throw` and `except`, then we don't need to have separate first class objects for the `exn_ref` and `exn_package`, right?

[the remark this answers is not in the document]
rossberg: @eholk, yes, there would only be the latter.

proposals/Layer-1.md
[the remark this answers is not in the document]
KarlSchimpf: What I was claim wasn't on the `catch_all`, but the other catch blocks that can be added to a try block. The earlier catches (if specified) have tags and allow the embedder to optimize the matches (just like the original proposal). A separate match is needed only if you a) embed `match` in the `catch_all`, or b) you store it locally and processes it later.

Adding these catch clauses to the try block I assumed would be after MVP.

proposals/Layer-1.md
KarlSchimpf: I assumed (like return-call) that a throw that also creates could be added as new operator after MVP.

proposals/Layer-1.md
```diff
@@ -0,0 +1,390 @@
...
+A `except` instruction has a single immediate argument, an exception tag.  The
+corresponding tagged exception type is used to define the data fields of the
+created exception. The values for the data fields must be on top of the operand
+stack, and must correspond to the exception's type signature.  These values are
+popped off the stack and an instance of the exception is then created. A
+reference to the created exception is then pushed onto the stack.
+
+### Throws
+
+The `throw` throws the exception on top of the stack. The exception is popped
```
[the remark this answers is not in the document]
KarlSchimpf: I assumed that the stack trace was added when created. Hence, throw and rethrow are the same.

proposals/Layer-1.md
KarlSchimpf: I agree with both comments.

proposals/Layer-1.md
rossberg: Since the combined operation is the simpler and more extensible choice (e.g. avoids the need for an exn_ref type) it seems more natural to go the other way round. Is there an immediate use case for the separated instructions?

proposals/Layer-1.md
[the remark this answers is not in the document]
rossberg: Yes, but as I argued in [another comment](https://github.com/WebAssembly/exception-handling/pull/33#discussion_r167611819), that bakes in assumptions about the exception mechanism that do not generally hold. It also is plain incompatible with extensions like resumption, where (1) the continuation would have to be captured at the throw point, not the creation point, and (2) rethrow would not capture a continuation, just pass it on.

PR #41, opened 2018-02-12
Clarify the relationship between the original proposal and Layer 1

eholk:
The original proposal has been significantly reworked based on experience trying to implement C++ exceptions. We are now pursuing a layered approach. This change clarifies that the old proposal is obsolete and for reference only, while Layer 1 represents the current MVP.


rossberg:
Thanks, that helps, but I think the term "layer" still is rather confusing. How about "step"?

eholk:
What about "Level," similar to how the various CSS specs are Level 1, Level 2, etc.?

rossberg:
Sounds good as well.

PR #42, opened 2018-02-12
Describe some Level 1+N features

eholk:
This provides some rough detail about possible future extensions, and more detail about the leveling approach.

Feel free to add new features! These are still meant to be in flux, but the goal is to try to capture what we might want to do in the future so we can stay future proof.

### DIFF
--- a/proposals/Level-1+N.md
+++ b/proposals/Level-1+N.md
@@ -138,3 +138,67 @@ catch e: StdPrint(msg):
   resume e()
 end
 ```
+
+### Example: Coroutines
+
+Consider the example below taken from https://en.wikipedia.org/wiki/Coroutine:
+
+```
+var q := new queue
+
+coroutine produce
+    loop
+        while q is not full
+            create some new items
+            add the items to q
+        yield to consume
+
+coroutine consume
+    loop
+        while q is not empty
+            remove some items from q
+            use the items
+        yield to produce
+```
+
+Using similar high level pseudo code, we can convert this as follows to use
+resumable exceptions:
+
+```
+Exception yield: () -> ()
+
+function produce():
+  loop:
+    while q is not full:
+      create some new items
+      add the items to q
+    throw yield()
+
+function consume():
+  loop:
+    while q is not empty:
+      remove some items from q
+      use the items
+    throw yield()
+
+function run():
+  // Start the producer
+  c1 = try:
+    produce()
+  catch e: yield():
+    e
+  end
+
+  // Start the consumer
+  c2 = try:
+    consume()
+  catch e: yeild():
+    e
+  end
+
+  loop:
+    // Run the producer
+    c1 = resume c1();
+    // Run the consumer
+    c2 = resume c2();
+```
--- a/proposals/Level-1+N.md
+++ b/proposals/Level-1+N.md
@@ -52,7 +52,7 @@ try resulttype
 end
 ```
 
-Note that the `catch exteption_index` and `catch` forms of the catch clause should be given different opcodes.
+Note that the `catch exception_index` and `catch` forms of the catch clause should be given different opcodes.
 
 ### Alternate Form: Catch Lists
 
@@ -69,7 +69,7 @@ end
 
 One downside of this form is that it does not directly support running different
 code for different exception types. This can still be accomplished, however, by
-dyanmically inspecting the exception inside the catch block.
+dynamically inspecting the exception inside the catch block.
 
 # Level 3
 
--- a/proposals/Level-1+N.md
+++ b/proposals/Level-1+N.md
@@ -186,14 +186,14 @@ function run():
   try:
     produce()
   catch e: yield():
-    c1 = e
+    e
   end
 
   // Start the consumer
   try:
     consume()
   catch e: yield():
-    c2 = e
+    e
   end
 
   loop:
@@ -248,30 +248,23 @@ https://en.wikipedia.org/wiki/Delimited_continuation:
 
 This can be translated into a resumable exceptions program. We'll start by
 rewriting the program so that it's closer to the high level syntax we've been
-using. Note that this program uses WebAssembly's stack machine nature.
+using.
 
 ```
 function main():
-  2
-  reset:
-    1
-    shift k:
-      k(5)
-    +
-  *
+  2 * (reset => 1 + (shift k => k(5)))
 ```
 
-Now we can convert the program to use resumable exceptions:
+Now we can convert the program to use resumable exceptions. For convenience, we
+use JavaScript-style lambdas (e.g. `(x, y) => x + y`) and assume a later pass
+will compile this away.
 
 ```
-
+Exception shift_k: (fn (i32) -> i32) -> i32
 
 function main():
-  2
-  reset:
-    1
-    shift k:
-      k(5)
-    +
-  *
+  2 * (try:
+         1 + (throw shift_k((k) => k(5)))
+       catch e: shift_k(f):
+         f((v) => resume e(v)))
 ```
--- a/proposals/Level-1+N.md
+++ b/proposals/Level-1+N.md
@@ -25,6 +25,9 @@ included with the initial exception proposal. The goal is to make sure each set
 of features can be discussed in terms of its own merits independent of the other
 features.
 
+Please note that everything in here should be considered fluid and subject to
+change.
+
 # Level 2
 
 ## Specific Catch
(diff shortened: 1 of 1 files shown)


KarlSchimpf (approved):
lgtm

proposals/Level-1+N.md
```diff
@@ -0,0 +1,103 @@
...
+code for different exception types. This can still be accomplished, however, by
+dyanmically inspecting the exception inside the catch block.
+
+# Level 3
+
+## Resumable Exceptions
+
+This proposal allows exception handlers to return a value to the site where the
+exception was thrown. Resumable exceptions enable or are equivalent to a number
+of advanced control flow features such as generators, coroutines, delimited
```
aardappel, line 80: That's a bold claim if you don't know the specifics. For example, the example below shows that the code that can resume (in the catch) is closely tied to the resumable code (the try). In the general case of coroutines these two can be in completely unrelated pieces of code, and there can even be multiple pieces of code that can resume the same coroutines. It be nice to at least hint at how these can be supported.

proposals/Level-1+N.md
```diff
@@ -0,0 +1,103 @@
...
+try resulttype
+  instruction*
+catch exception_index*
+  instrunction*
+end
+```
+
+One downside of this form is that it does not directly support running different
+code for different exception types. This can still be accomplished, however, by
+dyanmically inspecting the exception inside the catch block.
```
aheejin: Why not?

eholk: The problem is that each of the exception indices may have a different type. Say we had:

```
try:
   ...
catch 0, 1:
  ...
end
```

and the exception at index 0 had type `[i32]` while the exception at index 1 had type `[f64 f64]`. We wouldn't know what values to push onto the stack since the same code is run for each exception. In other words, this doesn't work for the combined catch-and-unpack operation.

proposals/Level-1+N.md
rossberg: @aardappel, the extension of exception handling with resumption is equivalent to what is known in recent research as "effect handlers". Turns out that they are powerful enough to express almost any side effect you can think of. Essentially, they provide a structured form of delimited continuations. If you're brave enough and don't mind exploiting a little bit of UB, they can even be implemented [on top of C](https://www.microsoft.com/en-us/research/publication/implementing-algebraic-effects-c/).

To express coroutines, you would define a Yield effect (aka resumable exception), and the handler would be the scheduler loop that when thrown to stores the current exception/continuation in a queue and instead resumes the next one. Underneath, this basically means switching stacks.

eholk: @aardappel, I just pushed another change that shows how to take one of the effect handling examples on eff-lang.org and translates it into a high-level pseudocode that matches the new resumable exceptions instructions. I think you raise a good point, so I will try and include examples for the rest of the features I listed before landing the PR.

proposals/Level-1+N.md
```diff
@@ -0,0 +1,204 @@
...
+  c1 = try:
+    produce()
+  catch e: yield():
+    e
+  end
+
+  // Start the consumer
+  c2 = try:
+    consume()
+  catch e: yeild():
```
rossberg: yield

proposals/Level-1+N.md
```diff
@@ -0,0 +1,204 @@
...
+      remove some items from q
+      use the items
+    throw yield()
+
+function run():
+  // Start the producer
+  c1 = try:
+    produce()
+  catch e: yield():
+    e
```
rossberg: Nit: May be clearer if the assignment to c1 occurs inside here (same below).

rossberg (approved):
Nice!

proposals/Level-1+N.md
eholk: done

proposals/Level-1+N.md
eholk: Good idea. Done.

proposals/Level-1+N.md
aardappel: @rossberg Thanks for the "Implementing Algebraic Effects in C" link, the C code in there clears up a lot of confusion. Having to resume stack copies in the same location as they were created sound problematic, but I am sure wasm implementations can improve on this and omit these "fragments", since they have more knowledge of the stack layout.

@eholk Thanks for the co-routine example, that clears up my confusion on whether a resumable exception can escape its catch-block.

eholk: @aardappel, no problem! It was a good exercise for me to figure out how to represent all of these control structures at once. I'm still trying to get `shift`/`reset` delimited continuations working, but once I do I'll upload the change and be ready to merge this PR.

One thing I realized about having escaping resumable exceptions is that we'll probably need to parameterize `except_ref` by its resumption type.

rossberg: Actually, `except_ref` would need to track the result type of the `try` that created it, not the type of the resumption -- that is only known once you have extracted a tag. More concretely, typing rules would be roughly like this:
```
(throw $e) : [t1*] -> [t2*]
if $e : [t1*] -> [t2*]

(try[t3*] A catch $e* B) : [] -> [t3*]
if A : [] -> [t3*]
   B : [(caught t3*)] -> [t3*]

(rethrow) : [t4* (caught t3*)] -> [t5*]

(handle[t6*] $e A else B) : [(caught t3*)] -> [t6*]
if $e : [t1*] -> [t2*]
   A : [t1* (resumption t2* t3*)] -> [t6*]
   B : [] -> [t6*]

(resume) : [t2* (resumption t2* t3*)] -> [t3*]
```
(Renaming except_ref to caught here.)

In practice, we would probably want to annotate exception types and some of the instructions with a flag saying whether they are resumable. Then we could also distinguish a plain type `(caught)` from the variant `(caught resumable t*)`, so that it doesn't complicate typing in the current exception proposal.

eholk: I was thinking `(resume)` would be typed kind of like `br`, so it would be `[t2* (resumption t2*)] -> unreachable`. Basically, once you resume an exception, the values you passed to it would return out of the `throw` instruction and then control flow would pick up there as if the exception were never thrown, never returning back to the `resume` instruction. How do you end up returning from a `resume`?

rossberg: That’s the magic of delimited continuations. ;) The resume returns when the try that caught that exception terminates — you don’t return to the context of the try (which might already have been exited before you even resume) but that of the resume. That’s what makes it delimited, because you only capture the continuation up to where you entered the try. If you actually were to return to the previous context of the try then you would essentially have captured the *full* continuation.

What happens may be a bit more intuitive if you think about it in terms of how stacks are switched:

1. when you enter a try, a new stack is created and execution switches to that stack, remembering the previous stack
2. when the stack is completed (i.e., the try terminates regularly), then this stack is no longer needed and execution restores the previous stack
3. when the try body throws, then execution also switches back to the previous stack, while the try's own stack gets captured in the caught exception package (as the representation of the resumption)
4. when you resume, execution again switches back to the captured stack, but from whatever the current stack is; i.e., the current stack is what is remembered now as the previous one that you return to in case of (2.)

(In many cases the creation of an additional stack can be optimised away, in particular for a try where the caught exception can be shown to not escape the catch.)

Note that your own example of coroutines very much assumes that `resume` returns. ;)  Though the assignment to c1/c2 should be removed now that you moved the assignments into the handler and the tries do not actually have a result. But the previous version demonstrated exactly how the result type of the resume is the same as the result type of the associated try.

proposals/Level-1+N.md
```diff
@@ -0,0 +1,238 @@
...
+try resulttype
+  instruction*
+(catch exception_index
+  instruction* )*
+(catch
+  instruction* )?
+end
+```
+
+Note that the `catch exteption_index` and `catch` forms of the catch clause should be given different opcodes.
```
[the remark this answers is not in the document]
eholk: Done.

proposals/Level-1+N.md
```diff
@@ -0,0 +1,238 @@
...
+try resulttype
+  instruction*
+catch exception_index*
+  instrunction*
+end
+```
+
+One downside of this form is that it does not directly support running different
+code for different exception types. This can still be accomplished, however, by
+dyanmically inspecting the exception inside the catch block.
```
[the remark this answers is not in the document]
eholk: Done.

eholk:
I finally worked out how to finish off the delimited continuation example, so I'm ready to land this assuming it looks okay to everyone else.

PR #43, opened 2018-02-14
Combine throw and create

KarlSchimpf:
Combine the throw and exception create instructions into a single 'throw' instruction. Add back from the original proposal the 'rethrow' instruction that can rethrow a caught exception.

Also cleans up cases some confusion of exception tags and exception indices. That is, the index is the static value used in the module, while the tag is the corresponding exception identifier for that index.

### DIFF
--- a/proposals/Level-1.md
+++ b/proposals/Level-1.md
@@ -86,34 +86,41 @@ instruction. Thrown exceptions are handled as follows:
 An `exception` is an internal construct in WebAssembly that is maintained by the
 host. WebAssembly exceptions (as opposed to host exceptions) are defined by a
 new `exception section` of a WebAssembly module. The exception section is a list
-of exception types, from which exceptions can be created.
+of exception definitions, from which exceptions can be created.
 
-Each exception type has a `type signature`. The type signature defines the list
-of values associated with the exception.
+Each exception definition describe the structure of corresponding exception
+values that can be generated from it. Exception definitions may also appear in
+the import section of a module.
 
-Within the module, exception types are identified by an index into the
+Each exception definition has a `type signature`. The type signature defines the
+list of values associated with corresponding thrown exception values.
+
+Within the module, exception definitions are identified by an index into the
 [exception index space](#exception-index-space). This static index refers to the
-corresponding runtime index that uniquely identifies the exception type, and is
-called the `exception tag`. The `tagged exception type` is the corresponding
-exception type refered to by the exception tag.
+corresponding runtime tag that uniquely identifies exception values created
+using the exception definition, and is called the `exception tag`.
 
-Exception types can be imported and exported by adding the appropriate entries
-to the import and export sections of the module. All imported/exported exception
-types must be named to reconcile exception tags between modules.
+Each exception definition within a module (i.e. in the exception and import
+section) are associated with unique exception tags.
+
+Exception tags can also be exported by adding the appropriate descriptors in the
+export section of the module. All imported/exported indices must be named to
+reconcile the corresponding exception tag referenced by exception indices
+between modules.
 
 Exception indices are used by:
 
-1. The `throw` instruction which creates a WebAssembly instance of the
-   corresponding tagged exception type, and then throws it.
+1. The `throw` instruction which creates a WebAssembly exception value
+   with the corresponding exception tag, and then throws it.
 
-2. The `if_except` instruction that queries an exception to see if it is an
-   instance of the corresponding tagged exception class, and if true it pushes
-   the corresponding values of the exception onto the stack.
+2. The `if_except` instruction that queries an exception value to see if the
+   exception tag corresponding to the module's exception index. If true it
+   pushes the corresponding values of the exception onto the stack.
    
 ### The exception reference data type
 
 Data types are extended to have a new `except_ref` type. The representation of
-an exception type is left to the host VM.
+an exception value is left to the host VM.
 
 ### Try and catch blocks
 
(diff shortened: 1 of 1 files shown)


proposals/Level-1.md
```diff
@@ -93,19 +92,19 @@ Each exception type has a `type signature`. The type signature defines the list
 of values associated with the exception.
 
 Within the module, exception types are identified by an index into the
-[exception index space](#exception-index-space). This index is referred to as
-the `exception tag`. The `tagged exception type` is the corresponding
+[exception index space](#exception-index-space). This static index refers to the
+corresponding runtime index that uniquely identifies the exception type, and is
```
rossberg: Suggestion: Maybe it's even clearer if you avoid calling the runtime thing an "index" in the first place. How about changing wording to use "tag" right away?

proposals/Level-1.md
```diff
@@ -93,19 +92,19 @@ Each exception type has a `type signature`. The type signature defines the list
...
+called the `exception tag`. The `tagged exception type` is the corresponding
```
rossberg: Terminology nitpick: Technically what's defined in the exception section is not a type. What an exception definition defines is a fresh tag, which is a runtime *value* that *has* the ascribed exception type (and multiple tags can have the same type yet be different tags). This may be nitpicking, but the difference is that it clarifies that (1) multiple tags can have the same type, and (2) dispatch is on values not types (types have no runtime representation in Wasm).

proposals/Level-1.md
```diff
@@ -93,19 +92,19 @@ Each exception type has a `type signature`. The type signature defines the list
...
 to the import and export sections of the module. All imported/exported exception
 types must be named to reconcile exception tags between modules.
 
-Exception tags are used by:
+Exception indices are used by:
 
-1. The `except` instruction which creates a WebAssembly instance of the
-   corresponding tagged exception type, and pushes a reference to it onto the
-   stack.
+1. The `throw` instruction which creates a WebAssembly instance of the
```
rossberg: Suggestion: I would say it creates an exception value with the corresponding tag.

proposals/Level-1.md
```diff
@@ -114,8 +113,7 @@ Exception tags are used by:
 ### The exception reference data type
 
 Data types are extended to have a new `except_ref` type. The representation of
-an exception type is left to the host VM, but its size must be fixed. The actual
-number of bytes it takes to represent any `except_ref` is left to the host VM.
+an exception type is left to the host VM.
```
rossberg: "exception type" -> "exception value"

proposals/Level-1.md
```diff
@@ -143,19 +141,13 @@ block.
...
-stack, and must correspond to the exception's type signature.  These values are
-popped off the stack and an instance of the exception is then created. A
-reference to the created exception is then pushed onto the stack.
-
-### Throws
-
-The `throw` throws the exception on top of the stack. The exception is popped
-off the top of the stack before throwing.
+The `throw` instruction takes an exception index as an immediate argument.  That
+index is used to identify the tagged exception type to throw.  The values to
```
rossberg: Suggestion: "tagged exception type" -> "exception tag"
"values to throw" -> "argument values"

proposals/Level-1.md
```diff
@@ -143,19 +141,13 @@ block.
...
+throw (corresponding to the signature) are popped from the top of the stack, and
+an exception is created. This includes attaching a stack track to the exception.
```
rossberg: Nit: The stack trace isn't part of the core semantics and its presence may depend on the embedder.

rossberg (commented):
Cool, thanks! I have a couple of suggestions but they are all nitpicks on terminology.

proposals/Level-1.md
KarlSchimpf: Sound good. Changing.

proposals/Level-1.md
KarlSchimpf: Ok. Reworking the code to match.

proposals/Level-1.md
KarlSchimpf: Done

proposals/Level-1.md
KarlSchimpf: Ok. reworking this text.

proposals/Level-1.md
KarlSchimpf: Removed.

proposals/Level-1.md
KarlSchimpf: Rewritten to match changes.

proposals/Level-1.md
```diff
@@ -177,23 +188,24 @@ If the selected catch block does not throw an exception, it must yield the
 value(s) expected by the corresponding catching try block. This includes popping
 the caught exception.
 
-Note that a caught exception can be rethrown using the `throw` instruction.
+Note that a caught exception value can be rethrown using the `throw`
```
rossberg: Typo: `rethrow` instruction?

proposals/Level-1.md
```diff
@@ -299,14 +320,17 @@ encoded above.
...
-the data fields of the exception.
+Each exception definition (defining an exception tag) has a type signature,
+which corresponds to the data fields of the exception.
 
 | Field | Type | Description |
 |-------|------|-------------|
 | `count` | `varuint32` | The number of types in the signature |
 | `type` | `value_type*` | The type of each element in the signature |
 
+Note: An `except_type` is not actually a type, just an exception definition that
```
rossberg: Well, the exception type in an exn def is indeed the type of the defined tag, so I believe this note isn't necessary (and probably misleading).

rossberg (approved):
LGTM modulo comments!

proposals/Level-1.md
KarlSchimpf: Done

proposals/Level-1.md
KarlSchimpf: Removed.

Issue #44, opened 2018-02-15
Should `if_except` pop the exception off the stack?


eholk:
For the behavior of `if_except`, the [current proposal](https://github.com/WebAssembly/exception-handling/blob/master/proposals/Level-1.md#exception-data-extraction) states:

>The conditional query of an exception succeeds when the exception on the top of
the stack is an instance of the corresponding tagged exception type (defined by
`except_index`).
>
> If the query succeeds, the data values (associated with the type signature of
the exception class) are extracted and pushed onto the stack, and control
transfers to the instructions in the then block.
>
>If the query fails, it either enters the else block, or transfer control to the
end of the if_except block if there is no else block.

It does not explicitly state what happens to the exception object on the stack. I would interpret this to mean that the exception object remains on the stack in both the then and the else branches. Is this the behavior we want?

There are a couple of other options that I think could be reasonable choices and are worth discussing.

## Option 0: Status Quo

This is the option that matches my read of the current proposal. The exception stays on the stack in both the then and else branches of `if_except`. I'm calling it Option 0 to make it easy to refer to in discussions.

Having the exception available in the then branch is possibly convenient for rethrowing the exception. However, it will be buried by the exception arguments and the producer will have to pop these off before getting access to the exception, meaning it may not actually be all that useful.

Having the exception available in the else branch makes it more compact to chain multiple tests, such as:
```
if_except 0:
 ... then 0 ...
else:
  if_except 1:
    ... then 1 ...
  end
end
```

## Option 1: `if_except` Always Pops the Exception

In this case, the producer would have to explicitly save the exception to a local if it wants to refer to it multiple times. For example:
```
tee_local 0
if_except 0:
 ... then 0 ...
else:
  get_local 0
  if_except 1:
    ... then 1 ...
  end
end
```

This seems more consistent with the rest of the Wasm instructions, which generally remove any arguments used from the stack. 

The downside is that every chained `if_except` will need a `get_local`, although chained `if_except`s are likely to be rare, I expect.

## Option 2: Pop in Then Branch, Leave in Else Branch

This feels natural to me. The `if_except` instruction attempts to unpack the exception, and if it exceeds then it consumes the exception and leaves the contents of the exception on the stack. In the else case, things are left unchanged since it failed to unpack the exception.

Care will be needed to make sure the types balance on the two branches, and the one armed `if_except` case seems weird under this strategy. This option does allow chaining like Option 0 does.

## Option 3: Multiple Matches

[...]

PR #48, opened 2018-02-26
Generalize exceptions to events

KarlSchimpf:
In order to make the exception handling proposal able to handle future events, the concept of an exception is extended to be an event. This PR introduces this concept.


proposals/Level-1.md
```diff
@@ -51,13 +51,12 @@ exceptions.  For embedders that have garbage collection (such as JavaScript),
 this is not a problem.
 
 However, not all embedders may have a garbage collector. For this reason,
-WebAssembly exceptions are designed to allow the use of reference counters (to
-exceptions) to perform the the garbage collection in the embedder.
+WebAssembly exceptions are designed to allow the use of reference counters to
+perform the the garbage collection in the embedder.
```
eholk, line 55: nit: 'the the' -> 'the'

I might rephrase this as "WebAssembly exceptions are designed to allow other storage management methods, such as reference counting, to perform the garbage collection in the embedder." I don't know what other methods their might be, but the idea is tracing GC is not required, using reference counting is an existence proof of this fact, but other methods are allowed too. We don't make any requirements on how the embedder manages storage, we simply say that it's possible to be done.

proposals/Level-1.md
```diff
@@ -51,13 +51,12 @@ exceptions.  For embedders that have garbage collection (such as JavaScript),
...
 
-To do this, WebAssembly exceptions are immutable once created. This avoids
-cyclic data structures that can't be garbage collected. It also means that
-exceptions can't be stored into WebAssembly's linear memory. The rationale for
-this is twofold:
+To do this, WebAssembly exceptions are immutable once created, to avoid cyclic
+data structures that can't be garbage collected. It also means that exceptions
```
eholk, line 58: nit: for spec text, I think spelling out contractions is better. i.e. "can't" -> "cannot"

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
 1. Throws not caught within a function body continue up the call stack, popping
    call frames, until an enclosing try block is found.
    
-1. If the call stack is exhausted without any enclosing try blocks, embedder
+1. If the call stack is exhausted without any enclosing try blocks, the embedder
    defines how to handle the uncaught exception.
    
 ### Event handling
```
eholk, line 83: We talked about this offline some, but I'll mention it here in case other people have suggestions: I'm not sure "event" is the right term here, since it evokes the idea of events in the `addEventHandler` sense. I suggested something like "condition," "effect," or maybe "signal" but I'm not convinced these are better terms.

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
...
+general in its specification such that more general kinds of event handling can
+be added to WebAssembly in the future without having to redo exception handling.
 
-Events are generated by event yielding instructions. They yeild to an
-appropriate event handler. The event handler is then run. That event handler
-can either resume control at the point of the event handler, or optionally
-resume by sending values back to the yielding instruction allow the
-originating code to resume.
+An event handler allows one to process an event generated by a block of
+code. Events yield the current execution and look for a corresponding event
```
eholk, line 90: What about "suspend" instead of "yield?"

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
...
+handler. If found, the corresponding event handler is run.  The event handler
+can optionally resume by sending values back to the yielded instruction,
+allowing the originating code to resume.
```
eholk, line 93: Wait, we're not specing resumable exceptions yet, are we?

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
...
 
 Exceptions are a special case of an event in that they never resume. Similarly,
-a throw instruction is the yielding instruction of an exception, and the
-corresponding catching block is the event handler of the exception.
+a throw instruction is the yielding event of an exception. The catch block
+associated with a try block defines how to handle the throw.
```
eholk, line 97: Is this implying that other kinds of events will have different instructions that go with them?

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
...
+section of a WebAssembly module. The event section is a list of event types
+that are defined by the module.
 
-The event section defines a list of event types that may be handled by the
-module.  Each `event type` has (at a minimum) an `event specifier` and a `type
-signature`. The event specifier defines the kind of event while the type
-signature defines the list of values associated with the event.  Depending on
+Each `event type` has a `event specifier` and a `type signature`. Depending on
 the value of the event specifier, additional values may be specified in the
-event type.
```
eholk, line 109: Did you mean to delete this line?

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
...
-event. For exceptions, that can't resume, the result type must always be type
-'void'.
-
-Note: Resumable events may not use the `throw` instruction. Rather, they may use
-a new (yet to be defined) `yeild` instruction that passes value(s) to its
-enclosing contexts. Such an instruction would allow the calling context to
-return values back, and resume with the instruction following the yeild
-instruction, based on the values returned.
+event type corresponds to an exception. Future extensions my add additional
+event specifiers.
```
eholk, line 109: Is it better to be more vague about event specifiers and just say that event entries are required to have a 0 byte, which is reserved for future expansion?

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
...
-corresponding runtime tag is called an `event tag`.
-
-Event types are numbered numerically, starting with the import section, and then
-followed by event types in the event section.  Each event type within a module
-are associated with unique event tags.
+The type signature of an event type is an index to the corresponding function
+signature defined in the `type` section. The parameter of the function signature
+defines the list of values associated with the event. The return value(s) of the
+signature defines what is passed back to yeilding instruction so that execution
+may resume.
```
eholk, line 115: I think for now it'd be better to say that events cannot have a return value. I don't think we want resumable exceptions in the Level 1 proposal, even though we are trying to leave room for them.

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
...
 
-Event types can also be exported by adding appropriate descriptors in the export
-section of the module. Each imported/exported event index must be named to
-reconcile the corresponding (runtime) event tag.
+Within a module, event types are identified by an `event index` to the [event
+index space](#event-index-space). This (module specific) index identifies the
```
eholk, line 118: nit: "This module-specific index"

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
...
 
-1. The `throw` instruction which creates a WebAssembly exception with the
-   corresponding exception tag, and then throws it.
+1. The `throw` instruction which creates a WebAssembly exception
+   with the corresponding exception tag, and then throws it.
 
-2. The `if_except` instruction that queries an exception to see if the exception
-   tag corresponding to the module's exception index. If true it pushes the
-   corresponding values of the exception onto the stack.
+2. The `if_except` instruction queries a exception to see if the corresponding
```
eholk, line 135: nit: "an exception"

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
...
+   exception tag of the exception matches the wanted exception index.  If true
```
eholk, line 136: "matches the _tag of_ the wanted exception index."

Just to be explicit that we are comparing tags to tags, and not tags to indices.

proposals/Level-1.md
```diff
@@ -203,80 +176,93 @@ corresponding exception.
...
+The `if_except` block begins with an `if_except` instruction, and
+has two instruction blocks, defined by the `then` and `else` instructions
+like that of an `if` block.
+
+The `if_except` instruction queries of the exception on top of the stack, and
+then pops the exception. If the exception has the same exception tag as defined
+by the `if_except` instruction, the `then` block is entered. Otherwise the
+`else` block is entered. If the if_except block doesn't have an else block,
+control passes to the instruction following that block.
+The exception value is not popped when queried.
```
eholk, line 233: In line 229 it says it pops the exception, but in line 233 it says the exception is not popped. In #44 I thought we decided to always pop the exception.

proposals/Level-1.md
```diff
@@ -203,80 +176,93 @@ corresponding exception.
...
+
+If the query fails, it either enters the else block, or transfer control to the
+end of the if_except block if there is no else block.
 
 ### Stack traces
 
-When an exception is thrown, conceptually a runtime stack is associated with the
-exception. However, WebAssembly does not have (direct) access to the stack
-trace. Hence, it is left to the embedder to maintain and use at it sees fit.
+Earlier discussion implied that when an exception is thrown, the runtime will
```
eholk, line 261: You can probably drop the reference to earlier discussion. I feel like this document should capture the current state of the proposal rather than the history.

proposals/Level-1.md
```diff
@@ -300,9 +287,9 @@ instructions are *structured* control flow instructions, and can be
 labeled. This allows branch instructions to exit try and `if_except` blocks.
 
 The `except_index` of the `throw` and `if_except` instructions defines the
-exception type (and hence, exception tag) to create/extract form. See [event
-index space](#event-index-space) for further clarification of exception
-tags.
+exception value (and hence, exception tag) to create/extract form. See
```
eholk, line 290: nit: form -> from

proposals/Level-1.md
```diff
@@ -313,36 +300,38 @@ This section describes change in the
...
-import and event sections. Thus, the index space starts at zero with imported
-events followed by internally-defined events in the [event
+import and event sections. Thus, the index space starts at zero with
+imported events, followed by internally-defined events in the [event
 section](#event-section).
 
-The event index space defines the (module) static version of runtine event
-tags. These (runtime) event tags are guaranteed to be unique over all loaded
+The event index space defines the (module) static version of runtine
+event tags. For event indicies that are not imported/exported, the
```
eholk, line 308: nit: indicies -> indices

proposals/Level-1.md
```diff
@@ -313,36 +300,38 @@ This section describes change in the
...
+unique name imported/exported, and are aliased to the corresponding event index
+defined in each module.
 
 ## Changes to the binary model
 
-This section describes changes in the [binary encoding design
-document](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md).
+This section describes changes in
+the
+[binary encoding design document](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md).
```
eholk, line 320: We can probably just reference the spec now that we have one, instead of the design doc.

proposals/Level-1.md
```diff
@@ -313,36 +300,38 @@ This section describes change in the
...
-WebAssembly.
+An exception reference points to an exception. The size is defined by
+the embedder.
 
 ### Language Types
 
 | Opcode | Type constructor |
 |--------|------------------|
-| -0x0x5 |  `except_ref`    |
+| -0x05  |  `except_ref`    |
```
eholk, line 334: I forget, did we decide to keep the type here, or move it up to 0x7whatever with `i32` and friends?

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
...
-signature` of the exception is the corresponding type signature of the event
-type. That is, the parameter list of the type signature defines the list of
-values associated with the exception, and the result type must be type `void`.
-
-Exceptions are identified by the by the corresponding event index, and is called
-an `exception index`. The corresponding (runtime) event tag is called the
-`exception tag`.  Like all event types, each exception type within a module
-(i.e. in the event and import sections) are associated with unique exception
-tags.
+An `exception` is an internal construct in WebAssembly that is maintained by the
```
rossberg, line 123: Same comment here about "maintained by embedder". Perhaps avoid duplicating this text.

proposals/Level-1.md
rossberg: "matches the tag denoted by the exception index"

proposals/Level-1.md
```diff
@@ -203,80 +176,93 @@ corresponding exception.
...
+end of the if_except block if there is no else block.
 
 ### Stack traces
 
-When an exception is thrown, conceptually a runtime stack is associated with the
-exception. However, WebAssembly does not have (direct) access to the stack
-trace. Hence, it is left to the embedder to maintain and use at it sees fit.
+Earlier discussion implied that when an exception is thrown, the runtime will
+pop the operand stack across function calls until a corresponding, enclosing try
+block is found. It also associates a stack trace that can be used to report
```
rossberg, line 263: Nit: "It also associates" -> "It may also associate", since this is not really a Wasm requirement because it's not observable in Wasm.

proposals/Level-1.md
```diff
@@ -313,36 +300,38 @@ This section describes change in the
...
-tags. These (runtime) event tags are guaranteed to be unique over all loaded
+The event index space defines the (module) static version of runtine
+event tags. For event indicies that are not imported/exported, the
+corresponding event tag is guaranteed to be unique over all loaded
 modules.
 
-For event indices imported/exported, unique event tags are created for
-each unique name imported/exported, and are aliased to the corresponding
-event index defined in each module.
+For event indices imported/exported, unique event tags are created for each
```
rossberg, line 312: How about simply saying "Events that are imported or exported alias the respective events defined elsewhere, denoting the same tag."

proposals/Level-1.md
```diff
@@ -313,36 +300,38 @@ This section describes change in the
...
 
 
 ### Data Types
 
 #### except_ref
 
-An exception reference points to an exception. The size is fixed, but unknown in
-WebAssembly.
+An exception reference points to an exception. The size is defined by
+the embedder.
```
rossberg, line 328: I would drop this -- why require that there is a fixed size? That's not observable. In particular, I would avoid formulations that suggest that the semantics depends the an embedder.

proposals/Level-1.md
rossberg: It is a reference type, so it should be in the same cluster as `anyfunc`. Probably want to leave 0x41 for anyref, though.

proposals/Level-1.md
```diff
@@ -389,55 +374,58 @@ or defined:
...
-| entries | `event_type* | list of defined event types.              |
 
 ##### Import section
 
-The import section is extended to include exception definitions by extending an
+The import section is extended to include event definitions by extending an
 `import_entry` as follows:
 
-If the `kind` is `event`:
+If the `kind` is `Event`:
```
rossberg, line 396: Is this defined anywhere?

proposals/Level-1.md
rossberg: Btw, I'm a bit confused about the diff here. It seems like an earlier version of that section has already landed?

proposals/Level-1.md
rossberg: Sorry, I meant 0x11 when I said 0x41.

proposals/Level-1.md
```diff
@@ -78,94 +77,69 @@ instruction. Thrown exceptions are handled as follows:
...
-corresponding catching block is the event handler of the exception.
+a throw instruction is the yielding event of an exception. The catch block
+associated with a try block defines how to handle the throw.
 
-An `event` is an internal construct in WebAssembly that is maintained by the
-embedder. WebAssembly events are defined by a new `event section` of a
-WebAssembly module.
+An `event` (such as an exception) is an internal construct in WebAssembly that
+is maintained by embedder. WebAssembly events are defined by a new `event`
+section of a WebAssembly module. The event section is a list of event types
```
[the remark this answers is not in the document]
rossberg: Here is how I would put it: A tag is a value to distinguish different events, and it has a type that defines what values are attached to it when creating an event value with it. The event section is a list of definitions of event tags that are generated at runtime/inittime. An event index is a numeric name to refer to an (imported or defined) event tag within a module.

In practice, you can probably often say "event" as a short for "event tag" where it is clear.

proposals/Level-1.md
rossberg: Well, the specification does not enforce specific implementation strategies. A hypothetical engine could heap-allocate everything, eschew the stack entirely for CPS, and still be conforming.

proposals/Level-1.md
rossberg: My latest discussion with Luke seems to suggest that we gonna have a few more reference types in the ref proposal, so perhaps go for 0x18. :)

proposals/Level-1.md
rossberg: Ah sorry, didn't show up in the diff.

Issue #19, opened 2017-05-17
Should producers/consumers assume throwing is "rare" and, if so, can the spec note this?


lukewagner:
This is certainly the basic assumption for C++ but not for many other languages.  The underlying question is whether engines can assume they can use the classic "zero-cost" EH strategy without tanking on performance of some language that throws all the time.  AFAICS, if one has to assume throwing/catching can be a hot path, it'd be preferable to use either an extra 'throwing' return value (with branches on every callsite) or a setjmp/longjmp-like strategy, which produce a general slowdown in the non-throwing case.

Given that:
 1. the reason to add EH to wasm is to allow compilers/runtimes to achieve better performance than they could otherwise in wasm-without-EH,
 2. the zero-cost strategy is something only the engine can do (at least in the short- to medium-term),
 3. with [multi-return](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#multiple-return), a compiler/toolchain could implement a pretty-fast non-zero-cost EH strategy in wasm-without-EH,

I think we should non-normatively state this assumption in the spec so engines can predictably provide the performance of the zero-cost strategy.

KarlSchimpf:
I agree that zero-cost exceptions are the long-term goal. I didn't want to add it to the document because I thought it was a distraction from the primary concern of getting the concept in WebAsembly.

lukewagner:
Partly, this issue is just a question for the entire wasm cg because it'd be good to see if everyone is on the same page here.

Assuming we are: for the spec/proposal, we probably don't have to go into any implementation details (like "zero-cost").  Rather, we could just say, in a non-normative note in the spec and an informative para in the proposal, that exceptions are assumed to be used for exceptional control flow and thus engines may choose impl strategies that bias towards the best non-exceptional performance at the cost of exceptional performance.

jfbastien:
It would be good to quantify this: how close are we perf-wise and size-wise to:

* Exceptions **off** on native platforms
* Exceptions **on** but never thrown
** Exceptions **on** and throws

eholk:
Related discussions have cropped up at https://github.com/WebAssembly/exception-handling/pull/49#issuecomment-368889883 and https://github.com/WebAssembly/exception-handling/pull/42#discussion_r167716629.

[...]

PR #49, opened 2018-02-26
Generalize exceptions to events

KarlSchimpf:
Change the exception handling proposal to use a more general notion of "events", so that it will be easier to generalize the proposal when yeild/resume events are added.


proposals/Level-1.md
```diff
@@ -88,82 +88,60 @@ general in its specification such that more general kinds of event handling can
...
+2. The `if_except` instruction queries a exception to see if the corresponding
+   exception tag of the exception matches the wanted exception index.  If true
+   it pushes the corresponding values of the exception onto the stack.
    
 ### The exception reference data type
 
-Data types are extended to have a new `except_ref` type. The representation of
-an exception value is left to the host VM.
+Data types are extended to have a new `except_ref` type, that refers to an
+exception. The representation of an exception is left to the embedder.
```
rossberg: "is left to the implementation"

proposals/Level-1.md
```diff
@@ -202,27 +180,30 @@ associated with the exception index. These values are popped of the stack and
 are used (along with the corresponding exception tag) to create the
 corresponding exception value. That exception value is then thrown.
 
-When an exception value is thrown, the host VM searches for nearest enclosing
-try block body that execution is in. That try block is called the _catching_ try
-block.
+When an exception value is thrown, the embedder searches for the nearest
```
rossberg: "the embedder searches" - it's the Wasm VM. How about: "When an exception is thrown, the active `try` block entered last is called the _catching_ try block."

proposals/Level-1.md
```diff
@@ -202,27 +180,30 @@ associated with the exception index. These values are popped of the stack and
...
 
 A throw inside the body of a catch block is never caught by the corresponding
 try block of the catch block, since instructions in the body of the catch block
 are not in the body of the try block.
 
-Once a catching try block is found for the thrown exception value, the operand
-stack is popped back to the size the operand stack had when the try block was
-entered, and then the values of the caught exception value is pushed onto the
-stack.
+As expected, it is left to the embedder to define how WebAssembly exceptions are
```
rossberg: This seems to repeat what is already said on L193.

proposals/Level-1.md
```diff
@@ -231,25 +212,27 @@ If the selected catch block does not throw an exception, it must yield the
...
 
-The `if_except block` defines a conditional query of the exception value on top
-of the stack. The exception value is not popped when queried. The if_except
-block has two subblocks, the `then` and `else` subblocks, like that of an `if`
-block. The then block is a sequence of instructions following the `if_except`
-instruction. The else block is optional, and if it appears, it begins with the
-`else` instruction.  The scope of the if_except block is from the `if_except`
-instruction to the corresponding `end` instruction.
+The `if_except` block begins with an `if_except` instruction, and
+has two instruction blocks, defined by the `then` and `else` instructions
```
rossberg: There isn't really a `then` instruction.

proposals/Level-1.md
```diff
@@ -231,25 +212,27 @@ If the selected catch block does not throw an exception, it must yield the
...
+like that of an `if` block.
+
+The `if_except` instruction queries of the exception on top of the stack, and
```
rossberg: Stray "of"

proposals/Level-1.md
```diff
@@ -317,42 +301,41 @@ exception tags.
...
 modules.
 
-For exception indices imported/exported, unique exception tags are created for
-each unique name imported/exported, and are aliased to the corresponding
-exception index defined in each module.
+For event indices imported/exported, unique event tags are created for each
+unique name imported/exported, and are aliased to the corresponding event index
+defined in each module.
 
 ## Changes to the binary model
```
rossberg: model -> format

proposals/Level-1.md
```diff
@@ -317,42 +301,41 @@ exception tags.
...
 [binary encoding design document](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md).
 
-
 ### Data Types
 
 #### except_ref
 
-An exception reference points to an exception value. The size
-is fixed, but unknown in WebAssembly (the host defines the size in bytes).
+An exception reference points to an exception. The size is defined by
```
rossberg: See earlier comment. I would drop any reference to size, given that it is not observable in any way.

sunfishcode:
In some places ([example](https://github.com/WebAssembly/design/blob/master/HighLevelGoals.md)), this feature is called "zero-cost exceptions", referring to implementation strategies in which non-throwing code is optimized, at the expense of making throwing slow. This is useful for languages like Rust and C++.

Is this proposal still intending to enable zero cost exceptions for WebAssembly?

I'm aware that this is a performance consideration, rather than a semantic consideration.

eholk:
@sunfishcode - While we want to support many languages, I think the primary goal with this proposal is still to make sure C++ is well-supported. I agree that code that does not use throw or try/catch should not take a performance penalty just because the VM supports those exceptions.

Do you have concerns that this proposal will be hard to implement without imposing a pervasive cost on all code?

sunfishcode:
@eholk The term "zero cost" comes from the fact that in typical "zero cost" implementations, entering a `try` block does not involve executing any code. The main point is that even code that does use `try`/`catch` frequently should not take a (significant) performance penalty, in a "zero cost" system.

Unifying exception handling with yield/resume seems counter to the assumptions of a "zero cost" model, because the whole point of a generator, for example, is to yield, often many times. It'd help me to understand this whole proposal if such considerations were documented.

proposals/Level-1.md
KarlSchimpf: Done

proposals/Level-1.md
KarlSchimpf: Deleted

proposals/Level-1.md
aheejin: Can't we add one, to be symmetric with `if` and `else` instruction pair?

proposals/Level-1.md
```diff
@@ -380,45 +348,46 @@ or defined:
...
+The export section is extended to reference event types by
 extending an `export_entry` as follows:
 
-If the `kind` is `Exception`, then the `index` is into the corresponding
-exception index in the [exception index space](#exception-index-space).
+If the `kind` is `Event`:
 
+| Field | Type | Description |
+|-------|------|-------------|
+| `index` | `varuitn32` | the index into the corresponding event index space |
```
aheejin: varuint32

aheejin (commented):
Mostly typo fixes

aheejin:
Can we add the syntactic sugar `then` to `if_except`/`else`/`end` as well, to be symmetric with the `if`/`else` instruction pair?

aardappel:
@sunfishcode : I voiced some similar concerns here: https://github.com/WebAssembly/exception-handling/pull/42

The paper @rossberg linked https://www.microsoft.com/en-us/research/publication/implementing-algebraic-effects-c/ is very interesting because it has some actual timings you can expect from such resumable exceptions, which initially look pretty bad (compared to an implementation specialized purely for, say, generators), but it also shows a transformation that in a common case can avoid a stack-swap all-together, which would result in a very fast generator.

I agree that it seems that exceptions and generators would have opposite requirements of what path should be fast. I am not sure if the transform mentioned in the paper can be applied commonly enough to bridge that gap.

proposals/Level-1.md
```diff
@@ -20,50 +20,49 @@ Exception handling allows code to break control flow when an exception is
...
+WebAssembly exception is extracted and copied onto the stack, allowing
 succeeding instructions to process the data.
 
-Lastly, exception lifetimes must be maintained by the host VM, so that it can
-collect and reuse the memory used by exceptions. This implies that the host must
-know where exceptions are stored, so that it can determine when an exception can
-be garbage collected.
+Lastly, exception lifetimes may be maintained by the embedder, so that it can
+collect and reuse the memory used by exceptions. This implies that an embedder
+needs know where exceptions are stored, so that it can determine when an
```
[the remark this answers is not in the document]
KarlSchimpf: Done

proposals/Level-1.md
```diff
@@ -20,50 +20,49 @@ Exception handling allows code to break control flow when an exception is
...
+exception can be garbage collected.
 
-This also implies that the host VM must provide a garbage collector for
-exceptions.  For host VMs that have garbage collection (such as JavaScript),
+This also implies that that embedders must provide a garbage collector for
+exceptions.  For ebedders that have garbage collection (such as JavaScript),
```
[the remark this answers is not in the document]
KarlSchimpf: Done

proposals/Level-1.md
```diff
@@ -78,92 +77,63 @@ instruction. Thrown exceptions are handled as follows:
...
 
-An event handler allows one to process an event generated by a block of
-code. Events yield the current execution and look for an enclosing try block to
-handle the event. If found, the corresponding event handler is run.  The event
-handler can optionally resume by sending values back to the yielded instruction,
-allowing the originating code to resume.
+In general, an event handler allows one to process an event generated by a block
+of code. Events suspend the current execution and look for a corresponding event
+handler. If found, the corresponding event handler is run.  Some event handlers
+my send values back to the suspended instruction, allowing the originating code
```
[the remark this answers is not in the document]
KarlSchimpf: Done

proposals/Level-1.md
```diff
@@ -78,92 +77,63 @@ instruction. Thrown exceptions are handled as follows:
...
+WebAssembly events (i.e. exceptions) are defined by a new `event` section of a
+WebAssembly module. The event section is a list of declared events associated
+with the module.
 
-Each `event type` has a `event specifier` and a `type signature`. Depending on
-the value of the event specifier, additional values may be specified in the
-event type.
+Each event has an `attribute` and a `type`. Currently, the attribute can only
+specify that the event is an exception. In the future, additional attribute
+values may be added when other events are added to WebAssemble.
```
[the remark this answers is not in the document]
KarlSchimpf: Done

proposals/Level-1.md
```diff
@@ -265,25 +229,30 @@ else
 end
 ```
 
-The conditional query of an exception succeeds when the exception value on the
-top of the stack has the corresponding exception tag (defined by `except_index`).
+In the first form, the instructions between the `if_except` and 'end' define the
+`then block`. In the second form, the instructions between the `if_except` and
+`else` define the `then block`, while the instructions between the `else` and
+the `end define the `else block`.
```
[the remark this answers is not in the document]
KarlSchimpf: Done

proposals/Level-1.md
```diff
@@ -265,25 +229,30 @@ else
...
+
+The conditional query of an exception checks the exception tag of exception on
+top of the stack. It succeeds only if the exception index of the instruction
+matches the corresponding exception tag. Once the query completes, the exception
+is poppod off the stack.
```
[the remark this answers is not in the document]
KarlSchimpf: Done

proposals/Level-1.md
```diff
@@ -265,25 +229,30 @@ else
...
-block is found. The implementation may actually not do this. Rather, it may
-first search up the call stack to see if there is an enclosing try. If none are
-found, it could terminate the thread at the point of the throw. This would
-allow better debugging capability, since the corresponding call stack is still
-there to query.
+When an exception is thrown, the runtime will pop the operand stack across
+function calls until a corresponding, enclosing try block is found. It may also
+associate a stack trace that can be used to report uncaught exceptions. However,
+the details of this is left to the embedder, and WebAssembly does not have
+(direct) access to the stack trace.
```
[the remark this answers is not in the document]
KarlSchimpf: 1 - The stack trace is intended to be a call stack trace, and my read still states this.

2. We may have two phase (for debugging), but we never stated that it would be mandatory. Dropped because that is an implementation detail and not a requirement.

proposals/Level-1.md
KarlSchimpf: Done

KarlSchimpf (commented):
Fixed misspellings.

proposals/Level-1.md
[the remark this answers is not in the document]
eholk: For what it's worth, stack traces are also an implementation detail since there is no way to access them from Wasm.

eholk:
@sunfishcode @aardappel - @lukewagner voices similar concerns about the cost of throwing and catching exceptions in #19. Why don't we consolidate this discussion there and update the proposal as needed based on the results of the discussion?

proposals/Level-1.md
```diff
@@ -1,4 +1,4 @@
-# Level 1 exception handling
+ Level 1 exception handling
```
eholk: nit: I'd leave the `#` so this is a heading.

proposals/Level-1.md
```diff
@@ -265,25 +229,30 @@ else
 end
 ```
 
-The conditional query of an exception succeeds when the exception value on the
-top of the stack has the corresponding exception tag (defined by `except_index`).
+In the first form, the instructions between the `if_except` and 'end' define the
+`then block`. In the second form, the instructions between the `if_except` and
+`else` define the `then block`, while the instructions between the `else` and
+the `end` define the `else block`.
```
eholk, line 235: @sunfishcode pointed out that this optional else block complicates decoding. We should consider making this not optional (#54). This is an issue for another PR though.

proposals/Level-1.md
```diff
@@ -265,25 +229,30 @@ else
...
+### Stack traces
 
-Earlier discussion implied that when an exception is thrown, the runtime will
-pop the operand stack across function calls until a corresponding, enclosing try
-block is found. The implementation may actually not do this. Rather, it may
-first search up the call stack to see if there is an enclosing try. If none are
-found, it could terminate the thread at the point of the throw. This would
-allow better debugging capability, since the corresponding call stack is still
-there to query.
+When an exception is thrown, the runtime will pop the operand stack across
```
eholk: It's probably okay to be a little imprecise and just say "pop the stack across function calls" rather than specifying it's the operand stack.

proposals/Level-1.md
```diff
@@ -380,45 +348,46 @@ or defined:
...
 
 #### High-level structure
 
-A new `exception` section is introduced and is named `exception`. If included,
-it must appear after immediately after the global section.
+A new `event` section is introduced and is named `event`. If included, it must
+appear after immediately after the global section.
 
-##### Exception section
+##### event section
```
eholk: nit: capitalize "Event."

eholk (approved):
lgtm modulo nits

proposals/Level-1.md
aheejin: What I meant by "why only the operand stack traces and not call stack traces" [here](https://github.com/WebAssembly/exception-handling/pull/49/files#r171537148) is this too.

proposals/Level-1.md
KarlSchimpf: Done.

proposals/Level-1.md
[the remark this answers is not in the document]
KarlSchimpf: Removed the end of the last sentence (on direct access), since it is optional of the implementation.

proposals/Level-1.md
KarlSchimpf: Removed "operand".

proposals/Level-1.md
KarlSchimpf: Done

PR #60, opened 2018-10-17
Merge two proposals in a single doc

aheejin:
So far there have been two proposals, we have not decided on which to
use for the instruction part. We have agreed to use the second proposal
for the module and binary model (new event section and its format).

This patch tries to update the spec to the status quo: this describes
both proposals for the instruction part, and present the single
agreed-on proposal for the module/binary part. I moved old files to old/
directory, just for ease of browsing.

I plan to add some more stuff to the proposal, such as some
modifications to `rethrow` and the restriction on the order of the event
section, but it will come after this patch.

- Little bit of backstory:
The initial motivation for the second proposal, which proposes a
first-class exception ref type, was to ease some of toolchain
implementation problems. The problems do not exist anymore for some
other reason, but some people want to keep the second proposal on table
because possible benefits it may bring, for example, better
expressiveness. So we update this proposal to reflect the status quo and
not to confuse people. We will eventually unify this to a single
proposal.


aheejin:
@mstarzinger I don't seem to be able to add you as a reviewer. Maybe because your ID does not belong to [WebAssembly team in github](https://github.com/WebAssembly/)?

aheejin:
Closing in favor of #61. Turns out I can't make a stacked PR diffs if I use branches in my own repo to compare.

PR #61, opened 2018-10-17
Merge two proposals in a single doc

aheejin:
So far there have been two proposals, we have not decided on which to
use for the instruction part. We have agreed to use the second proposal
for the module and binary model (new event section and its format).

This patch tries to update the spec to the status quo: this describes
both proposals for the instruction part, and present the single
agreed-on proposal for the module/binary part.

I plan to add some more stuff to the proposal, such as some
modifications to `rethrow` and the restriction on the order of the event
section, but it will come after this patch.

- Little bit of backstory:
The initial motivation for the second proposal, which proposes a
first-class exception ref type, was to ease some of toolchain
implementation problems. The problems do not exist anymore for some
other reason, but some people want to keep the second proposal on table
because possible benefits it may bring, for example, better
expressiveness. So we update this proposal to reflect the status quo and
not to confuse people. We will eventually unify this to a single
proposal.


aheejin:
@mstarzinger I don't seem to be able to add you as a reviewer. Maybe because your ID does not belong to WebAssembly team in github?

proposals/Exceptions.md
```diff
@@ -55,9 +43,9 @@ by a new _exception_ section of a WebAssembly module. The exception section is a
 list of exceptions. Each exception has a _type signature_. The type signature
 defines the list of values associated with an exception.
 
-Within the module, exceptions are identified by an index into
-the [exception index space](#exception-index-space).  This index is referred to
-as the _exception tag_.
+Within the module, exceptions are identified by an index into the [exception
+index space](#exception-index-space). This index is referred to as the
```
rossberg: Old text, but worth fixing while you're here: the index is not the tag, the index references the tag.

proposals/Exceptions.md
```diff
@@ -262,130 +250,471 @@ Like the `block`, `loop`, and `if` instructions, the `try` instruction is a
 *structured* instruction, and is implicitly labeled. This allows branch
 instructions to exit try blocks.
 
-The `except_index` of the `catch` instruction is the exception tag for the caught
-exception. Similarly, the `except_index` of the `throw` instruction is the tag for
-the constructed exception.  See [exception index space](#exception-index-space)
-for further clarification of exception tags.
+The `except_index` of the `catch` instruction is the exception tag for the
```
rossberg: Nit: "is the exception tag" -> "denotes the exception tag"

proposals/Exceptions.md
```diff
@@ -262,130 +250,471 @@ Like the `block`, `loop`, and `if` instructions, the `try` instruction is a
...
+caught exception. Similarly, the `except_index` of the `throw` instruction is
```
rossberg: Same here

proposals/Exceptions.md
```diff
@@ -262,130 +250,471 @@ Like the `block`, `loop`, and `if` instructions, the `try` instruction is a
...
+This also implies that that embedders must provide a garbage collector for
+exceptions. For embedders that have garbage collection (such as JavaScript),
+this is not a problem.
+
+However, not all embedders may have a garbage collector. For this reason,
+WebAssembly exceptions are designed to allow other storage management methods,
+such as reference counting, to perform the garbage collection in the embedder.
+
+To do this, WebAssembly exceptions are immutable once created, to avoid cyclic
+data structures that cannot be garbage collected. It also means that exceptions
```
rossberg: Nit: "that cannot be garbage collected" -> "that cannot easily be reference-counted"

proposals/Exceptions.md
```diff
@@ -262,130 +250,471 @@ Like the `block`, `loop`, and `if` instructions, the `try` instruction is a
...
+can't be stored into linear memory. The rationale for this is twofold:
+
+* For security. Loads and stores do not guarantee that the data read was of the
+  same type as stored. This allows spoofing of exception references that may
+  allow a WebAssembly module to access data it should not know in the host VM.
+
+* The embedder does not know the layout of data in linear memory, so it can't
+  find places where exception references are stored.
+
+Hence, while an exception reference is a new first class type, this proposal
```
rossberg, line 350: Perhaps include a link to the reference-types proposal here and clarify the intention that exceptions are added as an additional reference type.

rossberg (commented):
It seems like this change consists mostly of copy & pasting the previous proposal docs into one. It would be helpful to rework that and remove all the redundancy between the descriptions of the two variants, and thereby make the difference clearer.

My understanding is that the 2nd variant only differs as follows: (1) a reference type of first-class exception packages is added, (2) the monolithic `try` instruction is split into separate `try` and `handle` instructions, and (3) `rethrow` takes an exn operand instead of an index immediate. Everything else is (or could be) pretty much the same. Is that correct? It would help to have this spelled out more explicitly.

aheejin:
The first proposal and the second one offer different instructions w/ different functionalities.

The instructions are
1st: `try` / `catch` / `catch_all` / `rethrow`
2nd: `try` / `catch` / `if_except` / `else` / `rethrow`
And 2nd's `catch` is different from that of 1st's; it is more like `catch_all` of 1st proposal. And the semantics of `rethrow` is also different too. And most importantly, the 2nd proposal includes the first-class exception ref type. So you are right that the instruction parts are mostly copy and paste from the old proposals, I thought there was very little that could be combined. I extracted the common binary / module changes part at the bottom, which is not relevant to the instruction changes.

proposals/Exceptions.md
```diff
@@ -262,130 +249,475 @@ Like the `block`, `loop`, and `if` instructions, the `try` instruction is a
...
+| Name | Opcode | Immediates | Description |
+| ---- | ---- | ---- | ---- |
+| `unreachable` | `0x00` | | trap immediately |
+| `nop` | `0x01` | | no operation |
+| `block` | `0x02` | sig : `block_type` | begin a sequence of expressions, yielding 0 or 1 values |
+| `loop` | `0x03` |  sig : `block_type` | begin a block which can also form control flow loops |
+| `if` | `0x04` | sig : `block_type` | begin if expression |
+| `else` | `0x05` | | begin else expression of if or try  |
+| `try` | `0x06` | sig : `block_type` | begins a block which can handle thrown exceptions |
+| `catch` | `0x07` | tag : `varuint32` | begins a block when the exception `tag` is thrown |
```
mstarzinger: nit: "thrown" -> "caught"

proposals/Exceptions.md
```diff
@@ -262,130 +249,475 @@ Like the `block`, `loop`, and `if` instructions, the `try` instruction is a
...
+This section describes change in the [instruction syntax
+document](https://github.com/WebAssembly/spec/blob/master/document/core/instructions.rst).
+
+### New instructions
+
+The following rules are added to *instructions*:
+
+```
+  try resulttype instruction* catch instruction* end |
+  except except_index |
```
mstarzinger: The "except except_index" line confused me on my read-through the first time around. Is this still accurate or can we drop this line?

proposals/Exceptions.md
```diff
@@ -262,130 +249,475 @@ Like the `block`, `loop`, and `if` instructions, the `try` instruction is a
...
+[definition](Modules.md#global-section)
+* `4` indicating an `Event` [import](#import-section) or
+[definition](#event-section)
+
+### Module structure
+
+#### High-level structure
+
+A new `event` section is introduced and is named `event`. If included, it must
+appear after immediately after the global section.
```
mstarzinger: nit: One redundant "after" (the first one).

proposals/Exceptions.md
```diff
@@ -262,130 +249,475 @@ Like the `block`, `loop`, and `if` instructions, the `try` instruction is a
...
 
-### Exception section
+##### Event section
 
-The `exception` section is the named section 'exception'. The exception section
-declares exception types using exception type signatures.
+The `section` section is the named section 'event'. The event section declares a
```
mstarzinger: nit: "`section` section" -> "`event` section"

mstarzinger (approved):
Looks good from my side. Only some small nits.

proposals/Exceptions.md
aheejin: Done

proposals/Exceptions.md
aheejin: I think we can drop this line. Thanks for the catch.

proposals/Exceptions.md
aheejin: Done

proposals/Exceptions.md
aheejin: Done

PR #64, opened 2018-10-24
Explain try blocks can be targeted from branches

aheejin:


proposals/Exceptions.md
```diff
@@ -109,7 +109,8 @@ Try blocks, like control-flow blocks, have a _block type_. The block type of a
 try block defines the values yielded by the evaluation the try block when either
 no exception is thrown, or the exception is successfully caught by one of its
 catch blocks, and the instructions within the catch block can recover from the
-throw.
+throw. Because `try` and `end` instructions define a control-flow block, they
```
mstarzinger, line 112: I think the same change should applied to the corresponding paragraph for the second proposal in line 443 below.

mstarzinger (approved):
Looks good with one comment to address.

proposals/Exceptions.md
aheejin: Done

Issue #58, opened 2018-06-01
Simplify exception dispatch to branches


rossberg:
The current `if_except` for handling exceptions is a structured instruction and as such quite complex. For example, embodying blocks, it has to deal with block signatures, labels, etc.

In the context of working on the GC proposal, where a similar construct is needed for downcasts, I realised that we probably don't want to introduce a zoo of block constructs. And it turns out that with the multi-value proposal we can simplify all these by replacing them with simple branches.

For the exception proposal, this would mean that instead of the structured `if_except` construct, we simply have a new branch of the form
```
br_on_exn <labelidx> <exnidx>
```
which checks whether the exception package on the top of the stack matches the exception denoted by `<exnidx>`, and if so, branches to `<labelidx>`. The trick, however, is that the block signature of the branch target has to be compatible with the exception's tag type -- because it receives the exception arguments as branch operands. If the exception does not match, it remains on the stack.

For example:
```
(exception $e i32 i64)
...
block $l (result i32 i64)
  ...
  try
    (i32.const 1)
    (i64.const 2)
    (throw $e)
  catch
    (br_on_exn $l $e)  ;; branch to $l with $e's arguments
    (rethrow)  ;; ignore other exceptions
  end
  ...
end
;; (i32.const 1) (i64.const 2) are now on the stack
```

This can now be used to construct handler switches in the same way `br_table` is used to construct regular switch:
```
block $end
  block $l1
    ...
      block $lN
        (br_on_exn $l1 $e1)
        ...
        (br_on_exn $lN $eN)
        (rethrow)
      end $lN
      ;; handler for $eN here
      (br $end)
    ...
  end $l1
  ;; handler for $e1
end $end
```

I think this is a simpler primitive and makes better reuse of existing instructions. (The `try` construct remains unchanged.) WDYT?

binji:
Interesting. I also was a bit concerned about making everything a block, so it's nice that this avoids that. The structure it requires seems a bit more awkward, but I suppose it's only as awkward as the `br_table` structure typically is.

eholk:
My first impression is that I like this proposal. It's nice that we can just add an instruction rather than a new block type.

I'm curious how this would affect the tools. I got the sense that `if_except` was kind of awkward for the tools to use. Maybe this `br_on_exn` would actually be easier to use? Or maybe it's just awkward in a different way.

aheejin:
Closed by #71.

PR #71, opened 2018-12-01
Simplify exception dispatch to branches

aheejin:
This proposes to use the idea suggested by #58 on using `br_on_exn` instead of `if_except`.

### DIFF
--- a/proposals/Exceptions.md
+++ b/proposals/Exceptions.md
@@ -128,10 +128,10 @@ Exception indices are used by:
 1. The `throw` instruction which creates a WebAssembly exception with the
    corresponding exception tag, and then throws it.
 
-2. The `br_on_exn` instruction queries an exception to see if the corresponding
-   exception tag denoted by the exception index. If true it branches to the
-   given label and pushes the corresponding values of the exception onto the
-   stack.
+2. The `br_on_exn` instruction queries an exception to see if it matches the
+   corresponding exception tag denoted by the exception index. If true it
+   branches to the given label and pushes the corresponding argument values of
+   the exception onto the stack.
 
 ### The exception reference data type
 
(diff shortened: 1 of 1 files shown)


proposals/Exceptions.md
```diff
@@ -215,39 +216,57 @@ exception on top of the stack is popped and then thrown.
...
+does not match, the `except_ref` value remains on the stack. For example, when
+an `except_ref` contains an exception of type (i32 i64), the target block
+signature should be (i32 i64) as well, as in the following example:
 
-if_except block_type except_index
-  Instruction*
-else
-  Instruction*
+```
+block $l (result i32 i64)
```
aheejin, line 240: Even though #58 suggested the multi-value proposal is a prerequisite for this and this example also uses a multi-value return block, the multi-value proposal is not necessarily a prerequisite for this exception handling proposal; we can do with a single value yielding blocks until the multi value proposal arrives.

proposals/Exceptions.md
```diff
@@ -128,9 +128,10 @@ Exception indices are used by:
 1. The `throw` instruction which creates a WebAssembly exception with the
    corresponding exception tag, and then throws it.
 
-2. The `if_except` instruction queries an exception to see if the corresponding
-   exception tag denoted by the exception index. If true it pushes the
-   corresponding values of the exception onto the stack.
+2. The `br_on_exn` instruction queries an exception to see if the corresponding
```
rossberg: "to see if the" -> "to see if it matches the"

proposals/Exceptions.md
```diff
@@ -128,9 +128,10 @@ Exception indices are used by:
...
+   exception tag denoted by the exception index. If true it branches to the
+   given label and pushes the corresponding values of the exception onto the
```
rossberg: "value" -> "argument values" for clarity?

proposals/Exceptions.md
```diff
@@ -215,39 +216,57 @@ exception on top of the stack is popped and then thrown.
...
-  Instruction*
-end
+br_on_exn label except_index
+```
+
+The `br_on_exn` instruction checks the exception tag of an `except_ref` on top
+of the stack if it matches the given exception index. If it does, it branches
+out to the label referenced by the instruction (In the binary form, the label
+will be converted to a relative depth immediate, like other branch
+instructions), and while doing that, pops exception values from the `except_ref`
```
rossberg: "pops exception values" -> "pops the `except_ref` value from the stack and instead pushes the exception's argument values"

rossberg (approved):
Looks good modulo some wording tweaks.

proposals/Exceptions.md
aheejin: Done

proposals/Exceptions.md
aheejin: Done

proposals/Exceptions.md
aheejin: Done

mstarzinger:
Looks good. Thanks!

PR #86, opened 2019-07-16
[js-api] Add specification.

Ms2ger:


Ms2ger:
This is based on what I understand the Chromium behaviour to be; it's not clear to me if that is the current consensus.

Probably this will need some text changes after the changes to `func_invoke` land in the core spec.

document/js-api/index.bs
```diff
@@ -977,6 +980,7 @@ This slot holds a [=function address=] relative to the [=surrounding agent=]'s [
     1. Let (|store|, |ret|) be the result of [=func_invoke=](|store|, |funcaddr|, |argsSeq|).
     1. Set the [=surrounding agent=]'s [=associated store=] to |store|.
     1. If |ret| is [=error=], throw an exception. This exception should be a WebAssembly {{RuntimeError}} exception, unless otherwise indicated by <a href="#errors">the WebAssembly error mapping</a>.
+    1. If |ret| is an exception, throw a {{RuntimeError}} exception.
```
mstarzinger: Currently the implementation just rethrows the existing exception (i.e. the "anyref" representing the exception package in WebAssembly), which happens to be an instance of {{RuntimeError}}. I totally see need to make these distinguishable from a {{RuntimeError}} thrown for the "error" case on line above though.

Ms2ger: How does that work? The core spec can't create RuntimeErrors, AIUI. Also, I get RuntimeErrors even when I'm throwing an anyref value I passed in from JS (https://github.com/web-platform-tests/wpt/pull/18294).

Ms2ger:
Any more thoughts on this, anyone?

aheejin:
Could you split this PR into adding the existing base JS API and the modified part? It is hard to see the difference related to exceptions.

Ms2ger:
The exceptions-related change is in the [last commit](https://github.com/WebAssembly/exception-handling/pull/86/commits/88f4029356536fc7cd76e6bac452d5d53893d268).

aheejin:
Thank you! But I still think splitting this is worth it, because it's possible we end up adding more commits to this, in which case it is hard to see only the related changes later. What do you think? @mstarzinger

mstarzinger:
+1 on splitting out the unmodified import of of the js-api documents. Those should be uncontroversial and we could land them right away. All further delta are then easier to see and review I think.

document/js-api/index.bs
mstarzinger: I see your point of having to keep out the {{RuntimeError}} out of the core spec. Agreed.

However there might be situations where the exception being propagated to the embedding (i.e. JavaScript) is not a {{RuntimeError}}. I am thinking about the case where the function just uses rethrow on an existing exnref. In this case the exception package being propagated to the JavaScript would be the original exception, not necessarily a {{RuntimeError}}. At least that was my current understanding, but I am open to specifying different semantics of course. Here is the sketched web-platform test I have in mind (just for illustration, might contain typos):

```
promise_test(async () => {
  const kWasmAnyRef = 0x6f;
  const kSig_v_r = makeSig([kWasmAnyRef], []);
  const kSig_r_v = makeSig([], [kWasmAnyRef]);
  const builder = new WasmModuleBuilder();
  const fnIndex = builder.addImport("module", "fn", kSig_v_v);
  const except = builder.addException(kSig_v_r);
  builder.addFunction("catch_and_rethrow", kSig_r_v)
    .addBody([
      kExprTry, kWasmStmt,
        kExprCallFunction, fnIndex,
      kExprCatch,
        kExprRethrow,
      kExprEnd,
      kExprRefNull,
    ])
    .exportFunc();

  const buffer = builder.toBuffer();

  class MyError extends Error {};
  const error = new MyError();
  const fn = () => { throw error };
  const {instance} = await WebAssembly.instantiate(buffer, {
    module: { fn }
  });
  assert_throws(new MyError(), () => instance.exports.catch_and_rethrow());
}, "Imported JS function throws, Wasm catches and rethrows");
```

Ms2ger: Okay, this makes sense to me. What doesn't make sense is how `kExprRefNull` gets turned into a RuntimeError in a case like this:

```js
promise_test(async () => {
  const kWasmAnyRef = 0x6f;
  const kSig_v_r = makeSig([kWasmAnyRef], []);
  const builder = new WasmModuleBuilder();
  const except = builder.addException(kSig_v_r);
  builder.addFunction("throw_null", kSig_v_v)
    .addBody([
      kExprRefNull,
      kExprThrow, except,
    ])
    .exportFunc();
  const buffer = builder.toBuffer();
  const {instance} = await WebAssembly.instantiate(buffer, {});
  assert_throws(new WebAssembly.RuntimeError(), () => instance.exports.throw_null());
}, "Wasm function throws null");
```

What is special about getting an anyref value into Wasm by throwing it in JS and catching it in Wasm, as opposed to creating it in Wasm or passing it as an argument?

Ms2ger:
I've rebased and updated the PR to what I think the behaviour is; it's not very formal, but that'll need to wait until the core spec exists.

document/js-api/index.bs
mstarzinger: I don't think there is anything special about getting an `anyref` value passed into Wasm this way. The difference is between `kExprThrow` and `kExprRethrow` I feel. The model I am currently imagining is the following:

- `kExprThrow`: This will always create a new exception package, wrapping values according to the exception signature, as well as some internal notion of stack trace (or means of reconstructing it) into a package. This package has type `exnref`. Should such a package hit JavaScript (e.g. via stack unwinding) then it should be represented as an instance of RuntimeError (or some other, yet to be determined error constructor).

- `kExprRethrow`: This will take an existing exception package of type `exnref` and use it as is. Such an exception package might be one that was created by Wasm (e.g. by `kExprThrow` as described above). It could however be any other JavaScript value that was passed into Wasm as well. The `exnref` type is currently inhabited by the same set of values as the `anyref` type. Should an exception package raised via `kExprRethrow` hit JavaScript, then it is _not_ guaranteed to be an instance of RuntimeError, it can be any value (the same as the originally passed value).

Note that all of the above is just describing the model that V8 follows at the moment. If the JavaScript API ends up spec'ing this differently, then V8 will of course adapt according to the spec.

For example I could very well imagine a model where `exnref` type is only inhabited by instances of RuntimeError. This would required an arbitrary JavaScript value caught by Wasm to be implicitly wrapped into an exception package. Should Wasm rethrow this package via `kExprRethrow`, then it would be represented as an instance of RuntimeError (not the same as the originally thrown value) to JavaScript. Would such a model make more sense?

aheejin:
I'm very sorry for long radio silence 😢

I think there are some issues to be resolved (regarding traps and such) before we finalize the JS API. Does this PR describe the current behavior of V8? If so, it is likely subject to change, how about waiting on landing this? I opened #89 to discuss some of the issues.

aheejin:
Is this up-to-date with the current spec? (It doesn't look like it) Are you planning to update this?

document/js-api/index.bs
```diff
@@ -1126,6 +1163,143 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+
+<h4 id="js-exceptions">JavaScript exceptions</h4>
+
+The <dfn>JavaScript exception tag</dfn> is a [=tag address=] reserved by this
+specification to distinguish exceptions originating from JavaScript.
+
+For any [=associated store=] |store|, the result of
+[=tag_parameters=](|store|, [=JavaScript exception tag=]) must be « [=externref=] ».
+
+Issue: Should it be possible for `br_on_exn` to extract the payload from an exception with this tag?
```
ioannad:
```suggestion
```

or 

```suggestion
 Issue: Should it be possible for `catch <JS-exception-tag>` to extract the payload from an exception with this tag?
```

Either way, I don't think this issue does not be resolved for this PR to be merged.

document/js-api/index.bs
```diff
@@ -391,6 +400,12 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
             1. Let |tableaddr| be |v|.\[[Table]].
             1. Let |externtable| be the [=external value=] [=external value|table=] |tableaddr|.
             1. [=list/Append=] |externtable| to |imports|.
+        1. If |externtype| is of the form [=externtype/tag=] |attribute| <var ignore>functype</var>,
+            1. Assert: |attribute| is [=tagtype/attribute/exception=].
+            1. If |v| does not [=implement=] {{Tag}}, throw a {{LinkError}} exception.
+            1. Let |tag| be |v|.\[[Address]].
+            1. Let |extern| be the [=external value=] [=external value/tag=] |tag|.
+            1. [=list/Append=] |extern| to |imports|.
```
ioannad: Shouldn't this be as follows, similarly to the other imports?

```suggestion
        1. If |externtype| is of the form [=externtype/tag=] |attribute| <var ignore>functype</var>,
            1. Assert: |attribute| is [=tagtype/attribute/exception=].
            1. If |v| does not [=implement=] {{Tag}}, throw a {{LinkError}} exception.
            1. Let |tag| be |v|.\[[Tag]].
            1. Let |externtag| be the [=external value=] [=external value/tag=] |tagaddr|.
            1. [=list/Append=] |externtag| to |imports|.
```

ioannad (approved):
Apart from the two possible changes suggested, it LGTM.

document/js-api/index.bs
```diff
@@ -1057,7 +1087,14 @@ Note: Exported Functions do not have a \[[Construct]] method and thus it is not
         1. [=Clean up after running a callback=] with |stored settings|.
         1. [=Clean up after running script=] with |relevant settings|.
         1. Assert: |result|.\[[Type]] is <emu-const>throw</emu-const> or <emu-const>normal</emu-const>.
-        1. If |result|.\[[Type]] is <emu-const>throw</emu-const>, then trigger a WebAssembly trap, and propagate |result|.\[[Value]] to the enclosing JavaScript.
+        1. If |result|.\[[Type]] is <emu-const>throw</emu-const>, then:
+            1. If |v| [=implements=] {{RuntimeException}},
+                1. Let |type| be |v|.\[[Type]].
+                1. Let |payload| be |v|.\[[Payload]].
```
aheejin, line 1093: - Where is this `[[Payload]]` defined? I can see `[[Value]]`, `[[Type]]`, and `[[Target]]` in [completion record specification type](https://tc39.es/ecma262/#sec-completion-record-specification-type) doc in ECMAScript spec but not `[[Payload]]`. But I also see this note:
https://github.com/WebAssembly/exception-handling/blob/ca2b7155837c136ce8b3c95c7e49afdd60974016/document/js-api/index.bs#L1043
So what should we change to reflect this change?
-  Should we add that to the ECMAScript spec ourselves?
- As we discussed in #189 we need to preserve the identity of exceptions by attaching an `externref`. Should we extend `func_invoke`'s return value from
```
(|store|, <var ignore>val</var>* | [=error=] | (exception |exntag| |payload|)
```
to 
```
(|store|, <var ignore>val</var>* | [=error=] | (exception |exntag| |payload| |stack|)
```
?

document/js-api/index.bs
```diff
@@ -172,6 +176,8 @@ urlPrefix: https://webassembly.github.io/reference-types/core/; spec: WebAssembl
         text: table
         text: mem
         text: global
+        for: externtype
+            text: tag
```
aheejin, line 180: Why is this `externaltype/tag` special and not just
```
text:tag
```
? And because they are differently defined they are used in different ways too:

`externtype/table`:
https://github.com/WebAssembly/exception-handling/blob/ca2b7155837c136ce8b3c95c7e49afdd60974016/document/js-api/index.bs#L402
`externtype/tag`:
https://github.com/WebAssembly/exception-handling/blob/ca2b7155837c136ce8b3c95c7e49afdd60974016/document/js-api/index.bs#L407

I think either way is fine, but do you think it'd be better to make them consistent? The same for `external value/tag` too.

document/js-api/index.bs
```diff
@@ -391,6 +400,12 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
             1. Let |tableaddr| be |v|.\[[Table]].
             1. Let |externtable| be the [=external value=] [=external value|table=] |tableaddr|.
             1. [=list/Append=] |externtable| to |imports|.
+        1. If |externtype| is of the form [=externtype/tag=] |attribute| <var ignore>functype</var>,
+            1. Assert: |attribute| is [=tagtype/attribute/exception=].
+            1. If |v| does not [=implement=] {{Tag}}, throw a {{LinkError}} exception.
+            1. Let |tagaddr| be |v|.\[[Address]].
```
aheejin, line 406: Which address is this? For tables and globals, it says `|v|.\[[Table]]` and `|v|.[[Global]]`. Is there a set of values that `v` contains?

document/js-api/index.bs
```diff
@@ -1005,8 +1028,15 @@ This slot holds a [=function address=] relative to the [=surrounding agent=]'s [
         1. [=list/Append=] [=ToWebAssemblyValue=](|arg|, |t|) to |args|.
         1. Set |i| to |i| + 1.
     1. Let (|store|, |ret|) be the result of [=func_invoke=](|store|, |funcaddr|, |args|).
+    1. Note: The expectation is that [=func_invoke=] will be updated to return (|store|, <var ignore>val</var>* | [=error=] | (exception |exntag| |payload|)).
     1. Set the [=surrounding agent=]'s [=associated store=] to |store|.
     1. If |ret| is [=error=], throw an exception. This exception should be a WebAssembly {{RuntimeError}} exception, unless otherwise indicated by <a href="#errors">the WebAssembly error mapping</a>.
+    1. If |ret| is exception |exntag| |payload|, then
+        1. If |exntag| is the [=JavaScript exception tag=], then
+            1. Let « [=ref.extern=] |externaddr| » be |payload|.
+            1. Throw the result of [=retrieving an extern value=] from |externaddr|.
```
aheejin, line 1037: Do we need these steps? Aren't these two steps cancelling each other?

document/js-api/index.bs
```diff
@@ -184,9 +190,12 @@ urlPrefix: https://webassembly.github.io/reference-types/core/; spec: WebAssembl
     text: memaddrs; for: moduleinst; url: exec/runtime.html#syntax-moduleinst
     text: signed_64; url: exec/numerics.html#aux-signed
     text: sequence; url: syntax/conventions.html#grammar-notation
+    text: exception; for: tagtype/attribute; url: syntax/types.html#syntax-tagtype
 urlPrefix: https://heycam.github.io/webidl/; spec: WebIDL
     type: dfn
         text: create a namespace object; url: create-a-namespace-object
+urlPrefix: https://webassembly.github.io/js-types/js-api/; spec: WebAssembly JS API (JS Type Reflection)
+    type: abstract-op; text: FromValueType; url: abstract-opdef-fromvaluetype
```
aheejin, line 198: Where is this `FromValueType`? I can't find it in https://webassembly.github.io/js-types/js-api/#abstract-opdef-fromvaluetype. Also `ToValueType` is defined within this file; then why is only this method from the type reflection proposal?

document/js-api/index.bs
```diff
@@ -1126,6 +1163,143 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+  boolean is(Tag exceptionTag);
+};
+</pre>
+
+A {{RuntimeException}} value represents an exception.
+
+<div algorithm>
+
+To <dfn>create a RuntimeException object</dfn> from a [=tag address=] |tagAddress| and [=list=]
+of WebAssembly values |payload|, perform the following steps:
```
aheejin, line 1228: How are "create an Exception object" algorithm and "`new Exception(exceptionTag, payload)` constructor steps" algorithm different?

https://github.com/WebAssembly/exception-handling/blob/ca2b7155837c136ce8b3c95c7e49afdd60974016/document/js-api/index.bs#L1272-L1283

https://github.com/WebAssembly/exception-handling/blob/ca2b7155837c136ce8b3c95c7e49afdd60974016/document/js-api/index.bs#L1289-L1301

Why do they exist separately?

document/js-api/index.bs
```diff
@@ -1126,6 +1163,143 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+
+1. Let |store| be the [=surrounding agent=]'s [=associated store=].
+1. Let |types| be [=tag_parameters=](|store|, |tagAddress|).
+1. Assert: |types|'s [=list/size=] is |payload|'s [=list/size=].
+1. [=list/iterate|For each=] |value| and |resultType| of |payload| and |types|, paired linearly,
+    1. Assert: |value|'s type matches |resultType|.
```
aheejin, line 1234: Here we assert for invariants, but in the constructor algorithm below we throw. What's the difference?

document/js-api/index.bs
```diff
@@ -1126,6 +1163,143 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+1. Let |exception| be a [=new=] {{RuntimeException}}.
+1. Set |exception|.\[[Type]] to |tagAddress|.
+1. Set |exception|.\[[Payload]] to |payload|.
```
aheejin, line 1237: In the constructor algorithm below, we run `ToWebAssemblyValue` on each payload value, but here we do not. What's the difference?

document/js-api/index.bs
```diff
@@ -1126,6 +1163,143 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+
+<h4 id="js-exceptions">JavaScript exceptions</h4>
+
+The <dfn>JavaScript exception tag</dfn> is a [=tag address=] reserved by this
+specification to distinguish exceptions originating from JavaScript.
+
+For any [=associated store=] |store|, the result of
+[=tag_parameters=](|store|, [=JavaScript exception tag=]) must be « [=externref=] ».
+
+ Issue: Should it be possible for `catch <JS-exception-tag>` to extract the payload from an exception with this tag?
```
aheejin, line 1290: Why should it be not possible? Wasm's `catch` wouldn't be able to examine the internals of the `externref` payload, but that's true for all `externref`s, no?

document/js-api/index.bs
```diff
@@ -1126,6 +1163,143 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+
+<div algorithm>
+
+To <dfn for=WebAssembly>throw</dfn> with a [=tag address=] |type| and matching [=list=] of WebAssembly values |payload|, perform the following steps:
+
+1. Unwind the stack until reaching the *catching try block* given |type|.
+1. Invoke the catch block with |payload|.
+
+Note: This algorithm is expected to be moved into the core specification.
```
aheejin, line 1299: Is this specific to JS exceptions?

aheejin (commented):
Sorry that I wasn't able to review this at the time of merging! I'm not very familiar with bikeshed and WebIDL syntax, so I'm trying to understand the current EH JS API as written. Here are some questions:

document/js-api/index.bs
Ms2ger: It's basically namespacing - I don't recall if I had an actual conflict with "tag" or if I just thought it made more sense. I'd be happy to change the others as well, but I'd rather do that in the main spec repo directly.

document/js-api/index.bs
Ms2ger: Yeah, the publishing is broken - https://github.com/WebAssembly/js-types/issues/23. js-types only needs one direction, but it seems risky to just copy/paste. I can turn the dependency in the other direction if that turns out to be helpful.

document/js-api/index.bs
Ms2ger: This is the field set in "create a Tag object" algorithm below. The supported fields depend on the interface that the value implements, which is checked in the step above.

document/js-api/index.bs
Ms2ger: I don't fully understand your question. The goal of the "JavaScript exception tag" check is to ensure an exception thrown from js → wasm → js maintains its identity.

document/js-api/index.bs
Ms2ger: "create an Exception object" is when you already have the wasm values; "new Exception" is the js-side constructor called from webidl. I don't see a super obvious way to share more code between them, unfortunately.

document/js-api/index.bs
Ms2ger: I don't feel strongly either way, but to support it, we need to expose the tag to core wasm somehow, right? I'm a bit fuzzy on how that would work.

document/js-api/index.bs
Ms2ger: No, this is supposed to be basically equivalent to the throw instruction in core wasm. The core spec wasn't finished when I wrote this, so I put this here so I had an api to use in the rest of the spec.

Ms2ger (commented):
Thanks for the reply, I hope this clarifies. Let me know if anything is still confusing.

document/js-api/index.bs
aheejin: Oh I think I was confused. I thought you apply `ref.extern` to the returned payload; I guess this means if the payload is the form `« [=ref.extern=] |externaddr| »`, then we retrieve an extern value from `externaddr`. Right?

document/js-api/index.bs
aheejin: Ping

document/js-api/index.bs
aheejin: I guess the only way wasm can catch JS exceptions is to import the JS exception tag. Can we define the tag in the JS side and make wasm import it?

document/js-api/index.bs
Ms2ger: Yeah, exactly. If you have thoughts on how to make this destructuring assignment pattern more readable, I'd love to hear them.

document/js-api/index.bs
Ms2ger: I don't have time today to page #189 back in, but I'll try to work on it next week.

aheejin:
@Ms2ger Thanks for the explanation!

document/js-api/index.bs
aheejin: Ping in case this was missed :) Do you think this would work?

PR #91, opened 2019-12-23
Merge 'document/' directory of 'WebAssembly/reference-types/master', excluding 'document/js-api/'.

ioannad:
To work on the `exception-handling` formal spec, this repository should be up to date with the required `reference-types` spec. 
I only merged the `document/` directory (on which I plan to work) excluding `document/js-api/` because of the open PR #86.
Another reason for not touching other directories except `document\` is the open PR #28 which adds changes to `interpreter/` and `test/`.


aheejin:
Thank you! Sorry for the delayed reply.

I think we can also add JS-API; I'm not sure if #86 will be updated any time soon, and we also have another plans for JS API because of #89. All JS API changes can be rebased onto this, so I don't think it will be a problem.

And also I think we can update the interpreter too. #28 is an old PR that implements [the first version of the spec](https://github.com/WebAssembly/exception-handling/blob/master/proposals/old/Exceptions.md), which had different sets of instructions. So I think we shouldn't wait on that PR either. But I also think the interpreter would better be updated by a separate PR.

aheejin (approved):
Thank you! Could you delete 'excluding ...' part from the title?

rossberg:
May I suggest that we only merge proposal dependencies wholesale? That is, if one proposal depends on another, merge/rebase the entire repo. Plus, document this dependency. That is the only way to stay up-to-date cleanly with upstream changes in the other repo and avoid versioning and merging chaos later (in my recent experience, that is already a hard enough problem without starting to fragment dependencies).

For example, a while ago I started rebasing the reftype proposal on the bulk proposal by merging it in.

PR #99, opened 2020-02-10
Dependencies in proposal

ioannad:
This PR adds the dependency to multi-value to the exception handling proposal text and to the README. 

I wrote an explanation of this dependency on the proposal text, but it's easier to see this once the verification and execution steps of `br_on_exn` and of `try` blocks are written out, as done [here](https://github.com/WebAssembly/exception-handling/issues/87#issuecomment-517216137) by @rossberg :

Validation:

```
ft = t1* -> t2*
C, label t2* |- e1* : t1* -> t2*
C, label t2* |- e2* : exnref -> t2*
-----------------------------------
C |- try ft e1* catch e2* end : ft

C_label(l) = C_exn(x) = t*
-------------------------------------
C |- br_on_exn l x : exnref -> exnref
```

Execution: 

```
v^n (try ft e1* catch e2* end)  -->  catch_m{e2*} (label_m{} v^n e1* end) end)
  (iff ft = t1^n -> t2^m)
S; F; catch_m{e*} T[v^n (throw a)] end  -->  S; F; label_m{} (exn a v^n) e* end
  (iff S_exn(a) = {typ t^n})

F; (exn a v*) (br_on_exn l x)  -->  F; v* (br l)
  (iff F_exn(x) = a)
```

Concerning the functionality of `try`-`catch` blocks, note especially the passing of `v^n` values into a `label_m{}`.
Concerning the functionality of `br_on_exn`, note especially the execution step resulting in a `br` instruction.

cc: @aheejin @Ms2ger @dhil

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,15 @@ holds a
 [proposal](https://github.com/WebAssembly/exception-handling/blob/master/proposals/Exceptions.md) for
 adding exception handling to WebAssembly.
 
+The exception handling proposal depends on the [reference-types](https://github.com/WebAssembly/reference-types) proposal
+and on the [multi-value](https://github.com/WebAssembly/multi-value) proposal.
+
 The repository is a clone
-of [WebAssembly/spec](https://github.com/WebAssembly/spec), and is rebased on the spec of its dependent proposal [WebAssembly/reference-types](https://github.com/WebAssembly/reference-types).
+of [WebAssembly/spec](https://github.com/WebAssembly/spec), first rebased on the spec of its dependency [reference-types](https://github.com/WebAssembly/reference-types), and then merged with the other dependency [multi-value](https://github.com/WebAssembly/multi-value). 
 
-The remainder of the document is contents of the [README.md](https://github.com/WebAssembly/reference-types/blob/master/README.md)
-document of that repository.
+The remainder of the document has contents of the two README files of the dependencies: [reference-types/README.md](https://github.com/WebAssembly/reference-types/blob/master/README.md) and [multi-value/README.md](https://github.com/WebAssembly/multi-value/blob/master/README.md).
 
-# Reference Types Proposal for WebAssembly
+## Reference Types Proposal for WebAssembly
 
 [![Build Status](https://travis-ci.org/WebAssembly/reference-types.svg?branch=master)](https://travis-ci.org/WebAssembly/reference-types)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -24,9 +26,20 @@ It is meant for discussion, prototype specification and implementation of a prop
...
 * See the [modified spec](https://webassembly.github.io/reference-types/core/) for details.
 
+## Multi-value Proposal for WebAssembly
+
+[![Build Status](https://travis-ci.org/WebAssembly/multi-value.svg?branch=master)](https://travis-ci.org/WebAssembly/multi-value)
+
+This repository is a clone of [github.com/WebAssembly/spec/](https://github.com/WebAssembly/spec/).
+It is meant for discussion, prototype specification and implementation of a proposal to add support for returning multiple values to WebAssembly.
+
+* See the [overview](proposals/multi-value/Overview.md) for a summary of the proposal.
```
aheejin: This link does not point to the correct document. I think we need an absolute address. The same for already committed reference type overview link above.

README.md
```diff
@@ -24,9 +26,20 @@ It is meant for discussion, prototype specification and implementation of a prop
 
 * See the [modified spec](https://webassembly.github.io/reference-types/core/) for details.
```
aheejin: Should we delete `/core/` here, to be consistent with the multi-value spec below?

README.md
```diff
@@ -7,13 +7,15 @@ holds a
...
 The repository is a clone
-of [WebAssembly/spec](https://github.com/WebAssembly/spec), and is rebased on the spec of its dependent proposal [WebAssembly/reference-types](https://github.com/WebAssembly/reference-types).
+of [WebAssembly/spec](https://github.com/WebAssembly/spec), first rebased on the spec of its dependency [reference-types](https://github.com/WebAssembly/reference-types), and then merged with the other dependency [multi-value](https://github.com/WebAssembly/multi-value). 
 
-The remainder of the document is contents of the [README.md](https://github.com/WebAssembly/reference-types/blob/master/README.md)
-document of that repository.
+The remainder of the document has contents of the two README files of the dependencies: [reference-types/README.md](https://github.com/WebAssembly/reference-types/blob/master/README.md) and [multi-value/README.md](https://github.com/WebAssembly/multi-value/blob/master/README.md).
 
-# Reference Types Proposal for WebAssembly
+## Reference Types Proposal for WebAssembly
```
aheejin: Why two ##s? The same for `spec` later.

proposals/Exceptions.md
```diff
@@ -10,10 +10,15 @@ on the second proposal, which uses first-class exception types, mainly based on
...
-should be represented as a subtype of `anyref`.
+This proposal requires the following proposals as prerequisites.
+
+- The [reference types proposal](https://github.com/WebAssembly/reference-types/blob/master/proposals/reference-types/Overview.md),
+  since the [`exnref`](#the-exception-reference-data-type) type should be represented as a subtype of `anyref`.
+
+- The [multi value proposal](https://github.com/WebAssembly/multi-value/blob/master/proposals/multi-value/Overview.md),
+  since otherwise the [`br_on_exn?`](#Exception data extraction) instruction would only work with exceptions which consume one value.
+  Moreover, by using [multi value](https://github.com/WebAssembly/multi-value/blob/master/proposals/multi-value/Overview.md),
+  the [`try` blocks](#Try and catch blocks) may use values already in the stack.
```
aheejin: - Let's use 'multi-value' for consistency, because it's the official name of that proposal.
- What does the question mark after `br_on_exn` mean?
- The links do not work. They should be `#try-and-catch-blocks` and `#exception-data-extraction` respectively. You can check if they work by 'view file' menu on the upper right.
- In 'exceptions which consume one value', it's not very clear to me what 'consume' means. How about 'exceptions that contains one value' or something?
- ' Moreover, by using multi-value, the try blocks may use values already in the stack': 'try' also can push multiple values onto the stack. How about mentioning both?

aheejin (commented):
Thank you! Sorry for the delayed reply.

rossberg (approved):
LGTM modulo @aheejin's comments.

ioannad:
Sorry for the delayed action! I think I addressed all your comments, @aheejin. Thank you and @rossberg for the review.

PR #103, opened 2020-03-23
Fix inconsistency in section order

backes:
In #90 it was decided to move the event section between memory section and global section. This change is reflected in the next paragraph, but not in the introduction.


aheejin (approved):
Thanks for catching this!

PR #107, opened 2020-04-14
Fix 'instruction syntax document' link

danbev:
Currently, the link to the instruction syntax document results in a `404`.
This commit updates the link to point to a instructions.rst that exists
and hopefully is the correct one to link to.


aheejin (approved):
Thank you!

PR #96, opened 2020-01-08
Merging the entire reference-types repo

ioannad:
This is a followup to PR #91 , addressing @rossberg 's comment https://github.com/WebAssembly/exception-handling/pull/91#issuecomment-571585205

The main commit's message:
"Import reference-types repo

Source: [WebAssembly/reference-types@d4bc208](https://github.com/WebAssembly/reference-types/tree/d4bc208c0f62ac19529e9e734b10c17c960549e5)"

I did not take care to keep the history, as @aheejin suggested, to not add hundreds of commits to the EH repo. 

I also updated the `README.md` to include the `README.md` from WebAssembly/reference-types.

I didn't add `deploy_key.enc`.

When this is merged, I will make another fork to start adding changes to the formal spec, as discussed with @aheejin. I'll start with `/document/core` changes, which are on their way.

PR #112, opened 2020-05-28
Merging the entire reference types repository

ioannad:
- Import reference types repo: https://github.com/WebAssembly/reference-types/tree/5bc46bbd387b51f52ee0088e76ff401ece0c8e4a

- Update README

As last time, I just copy pasted the entire reference type repository,


aheejin:
Thanks. By the way, I know I once asked you to squash all commits in #96 because I was not sure adding hundreds of commits from the upstream to this repo as is was a good idea, but I noticed other downstream repos are merging upstream commits as they are. Is there a rule on how to do this? If merging commits as they are is the rule, should we revert #96? @rossberg

Ms2ger:
https://github.com/WebAssembly/proposals/blob/master/howto.md#syncing-with-upstream-changes suggests merging

rossberg:
If you squash upstream commits, then I think you might run into a lot more merge conflicts when you re-merge again later.

aheejin:
Hmm, I see. Thanks for the answers. 
@ioannad I'm sorry that I misguided you on squashing commits in #96. I think it'd be better we revert #96, and merge all upstream change that happened so far afresh. I can take care of it too, unless you want to do it yourself. Sorry again for the trouble 😢

ioannad:
@aheejin no problem at all, I am sorry for not noticing https://github.com/WebAssembly/proposals/blob/master/howto.md#syncing-with-upstream-changes earlier! 😬 I'm on it.

I just noticed that #91 was also squashed, so what I'll do is merge reference-types/master on top of e7196b7 (just before #91) and then cherry pick all changes (except #91 and #96). Then I'll force push to my branch to update this PR.

ioannad:
@aheejin: the changes I force-pushed are just as I described in my [last comment](https://github.com/WebAssembly/exception-handling/pull/112#issuecomment-640091081). Reverting the two commits I mention above, naturally creates lots of conflicts. Is the PR ok like this or shall I retry, merging on top of the current master branch instead?

PR #115, opened 2020-06-12
Revert "Merging the entire reference-types repo (#96)"

aheejin:
This reverts commit 10d6c6c738f370431efbb65cc9c0d1de8bc75723.

This reverts #96, which squashed all upstream commits into one, which
was not the recommended way in
https://github.com/WebAssembly/proposals/blob/master/howto.md#syncing-with-upstream-changes.

This leaves README.md untouched because #99 made significant changes on
top of the merged version already.

cc @ioannad


rossberg (approved):
Rubberstamped

PR #117, opened 2020-06-12
Revert "Merge 'document/' directory of 'WebAssembly/reference-types/master' (#91)"

aheejin:
This reverts commit 0d8fab65ed4bf5d90e265f8e54991bbd7d8ccbef.

This reverts #91, which squashed all upstream commits into one, which
was not the recommended way in
https://github.com/WebAssembly/proposals/blob/master/howto.md#syncing-with-upstream-changes.

cc @ioannad

PR #88, opened 2019-08-15
Import the JS API specification.

Ms2ger:

PR #120, opened 2020-06-15
Merge reference types repo

aheejin:
This merges reference types repo. The only thing manually fixed was
resolving conflict in README.md.

cc @ioannad


aheejin:
When merging, which button should I use, between "Create a merge commit" and "Rebase and merge"? (Certainly not "Squash and merge" this time :) ) "Create a merge commit" is currently disabled for this repo, but I can enable it if necessary.

binji:
@aheejin You'll want "Create a merge commit". I typically merge manually from the command-line so I don't have to enable the option (since we normally don't want it).

aheejin:
@binji Thanks!

By the way, this proposal also depends on multi-value. How should I maintain the merges between the two upstream repos? Just do merges from both repos and resolve conflicts?

ioannad:
@aheejin I think the multi-value proposal is now merged in the main spec, so only the reference types spec needs to be merged here. What was the problem with #112 ?

aheejin:
@ioannad I didn't mean there was a problem with #112, but you said there were lots of conflicts (because we didn't revert all squashed commits, I presume?), I thought it would be cleaner to revert all squashed commits (which I did in #115, #116, #117, and #119) and merge upstream from the fresh state.

ioannad:
Oh I see, thank you for explaining! I had reverted #91 and #96 in #112 but I missed that #88 was also squashed. I'll remove #112 then, and I'll rebase my version of the current EH core spec, when this PR here is merged.

rossberg (approved):
Rubberstamp

aheejin:
@Ms2ger Uh, sorry, you merged this using the command line..? This repo does not enable that option (unless you do that using the command line), so I'm wondering....

Ms2ger:
Yeah, it would be a little silly to do all this work to create an actual merge commit and then squash it in the UI :)

PR #121, opened 2020-06-18
Current core formal spec for WebAssembly Exception Handling

ioannad:
The rendered spec of this PR can be seen here: https://ioannad.github.io/exception-handling

Included in this PR:

- Detailed core formal spec additions aiming to fully implement:
  + the "informal formal" core spec laid out by Andreas @rossberg here: https://github.com/WebAssembly/exception-handling/issues/87#issuecomment-517216137 and
  + the core spec outlined in the [proposal overview](https://github.com/WebAssembly/exception-handling/blob/master/proposals/Exceptions.md), while removing the mention of "events" everywhere but in the binary format ,
  + prose for the above in: syntax, validation, execution, binary format, text format, and the appendix, as well as an example for throw contexts.
- Travis-ci build status icon to `README.md`. The contents of this PR get built and deployed on my fork's github page successfully, using [the howto instructions](https://github.com/WebAssembly/proposals/blob/master/howto.md#setting-up-travis-and-github-page).

Not included in this PR:

- Neither the new values representing unresolved throw contexts, nor the additional rules for the execution contexts, which we discussed in #87, are added here. I plan to add these separately.
- No JS-api/WebIDL changes.

Side comments:

- In [document/core/text/types.rst](https://ioannad.github.io/exception-handling/core/text/types.html#text-refedtype): I'm not sure whether `refedtype ::= ... | event ⇒ exnref` should have `event` there or something else (perhaps `exception`?) However, since currently the only events are exceptions, this is probably not really an issue.
- The "informal formal spec" defines the administrative instruction which represents an exception package as `exnref`, i.e.,  the same name as the type. I called this administrative instruction `ref.exn` instead for two reasons;  to match the style of `ref.null` and `ref.extern`, and to avoid ambiguity in the discussions.
- I removed multi-value mentions from `README.md` because multi-value is now part of the main spec. For the same reason I didn't add "multi-value" to the list of included proposals at the [top of the core spec landing page](https://ioannad.github.io/exception-handling/core/).


document/core/exec/runtime.rst
```diff
@@ -533,19 +577,27 @@ In order to express the reduction of :ref:`traps <trap>`, :ref:`calls <syntax-ca
      \REFFUNCADDR~\funcaddr \\ &&|&
      \REFEXTERNADDR~\externaddr \\ &&|&
      \INVOKE~\funcaddr \\ &&|&
+     \EXNREFADDR~\eventaddr~val^\ast \\ &&|&
+     \THROWADDR~\eventaddr \\ &&|&
+     \CATCHN_n~\{\instr_2^\ast\}~\instr_1^\ast~\END \\ &&|&
      \LABEL_n\{\instr^\ast\}~\instr^\ast~\END \\ &&|&
      \FRAME_n\{\frame\}~\instr^\ast~\END \\
```
```suggestion
     \LABEL_n\{\instr^\ast\}~\instr^\ast~\END \\ &&|&
     \CATCHN_n\{\instr^\ast\}~\instr^\ast~\END \\ &&|&
     \FRAME_n\{\frame\}~\instr^\ast~\END \\
```

document/core/exec/runtime.rst
```diff
@@ -608,6 +660,66 @@ This definition allows to index active labels surrounding a :ref:`branch <syntax
...
+     \LABEL_n\{\instr^\ast\}~\XT~\END \\ &&|&
+     \CATCHN_n\{\instr^\ast\}~\XT~\END \\ &&|&
+     \FRAME_n\{F\}~\XT~\END \\
+   \end{array}
+
+.. math::
+   \begin{array}{rcl}
+   S;~F;~\CATCHN_m~\{\instr^\ast\}~\val^m~\END &\stepto& S;~F;~\val^m \\
+   S;~F;~\CATCHN_m~\{\instr^\ast\}~\XT[\val^n~[\_]~(\THROWADDR~a)]~\END &\stepto&
+      S;~F;~\LABEL_m~\{\}~(\EXNREFADDR~a~\val^n)~{\instr}^\ast~\END \\
```
```suggestion
   S;~F;~\CATCHN_m\{\instr^\ast\}~\val^m~\END &\stepto& S;~F;~\val^m \\
   S;~F;~\CATCHN_m\{\instr^\ast\}~\XT[\val^n~[\_]~(\THROWADDR~a)]~\END &\stepto&
      S;~F;~\LABEL_m\{\}~(\EXNREFADDR~a~\val^n)~{\instr}^\ast~\END \\
```

document/core/exec/runtime.rst
```diff
@@ -608,6 +660,66 @@ This definition allows to index active labels surrounding a :ref:`branch <syntax
...
+   \production{(throw contexts)} & \XT &::=&
+     \val^\ast~[\_]~\instr^\ast \\ &&|&
+     \LABEL_n\{\instr^\ast\}~\XT~\END \\ &&|&
+     \CATCHN_n\{\instr^\ast\}~\XT~\END \\ &&|&
+     \FRAME_n\{F\}~\XT~\END \\
+   \end{array}
+
+.. math::
+   \begin{array}{rcl}
+   S;~F;~\CATCHN_m~\{\instr^\ast\}~\val^m~\END &\stepto& S;~F;~\val^m \\
```
rossberg: These reduction rules seem out of place here. I think they all belong into instructions.rst, probably into an auxiliary section on exception handlers there, analogous to the ones for blocks.

document/core/exec/runtime.rst
```diff
@@ -608,6 +660,66 @@ This definition allows to index active labels surrounding a :ref:`branch <syntax
...
+   S;~F;~\CATCHN_m~\{\instr^\ast\}~\XT[\val^n~[\_]~(\THROWADDR~a)]~\END &\stepto&
+      S;~F;~\LABEL_m~\{\}~(\EXNREFADDR~a~\val^n)~{\instr}^\ast~\END \\
+   && \hspace{-12ex} (\iff S.\SEVENTS[a]=\{\EVATTRIBUTE~\EXCEPTION, \EVTYPE~[t^n]\to[]\}) \\
+   \end{array}
+
+Throw contexts allow to potentially capture the throw of an exception event by surrounding catch blocks.
```
```suggestion
Throw contexts allow matching the program context around a throw instruction up to the nearest enclosing :math:`\CATCHN` handler, thereby selecting the exception handler responsible for an exception.
```

document/core/exec/runtime.rst
```diff
@@ -533,19 +577,27 @@ In order to express the reduction of :ref:`traps <trap>`, :ref:`calls <syntax-ca
...
 
 The |INVOKE| instruction represents the imminent invocation of a :ref:`function instance <syntax-funcinst>`, identified by its :ref:`address <syntax-funcaddr>`.
 It unifies the handling of different forms of calls.
 
-The |LABEL| and |FRAME| instructions model :ref:`labels <syntax-label>` and :ref:`frames <syntax-frame>` :ref:`"on the stack" <exec-notation>`.
+The |THROWADDR| instruction represents the imminent throw of an :ref:`exception reference value <syntax-exnrefaddr>` based on an :ref:`event instance <syntax-eventinst>`, identified by its :ref:`address <syntax-eventaddr>`.
+The values it will consume to create the exception depend on the event's :ref:`event type <syntax-eventtype>`.
+It unifies the throwing of different forms of events.
+
+The |LABEL|, |FRAME|, and |CATCHN| instructions model :ref:`labels <syntax-label>`, :ref:`frames <syntax-frame>`, and :ref:`try-catch blocks <syntax-try>` respectively, :ref:`"on the stack" <exec-notation>`.
```
```suggestion
The |LABEL|, |FRAME|, and |CATCHN| instructions model :ref:`labels <syntax-label>`, :ref:`frames <syntax-frame>`, and active :ref:`exception handlers <syntax-try>`, respectively, :ref:`"on the stack" <exec-notation>`.
```

document/core/exec/runtime.rst
```diff
@@ -608,6 +660,66 @@ This definition allows to index active labels surrounding a :ref:`branch <syntax
...
+
+.. note::
+   For example, catching a simple :ref:`throw <exec-throw>` in a :ref:`try-catch block <exec-try>` would be as follows.
+
+   Let :math:`x` be the :ref:`event index <syntax-eventidx>` of an event in :math:`F.\AMODULE`, such that
+
+   .. math::
+      F.\AMODULE.\MIEVENTS[x] = \{ \EVIATTRIBUTE~\EXCEPTION, \EVITYPE~[t^n]\to[]\} = S.\SEVENTS[a]
+
+   Let :math:`\expand_F(\X{bt}) = [t^n] \to [\EXNREF]`, and consider arbitrary :ref:`values <syntax-val>` :math:`(val : t)^n`.
```
rossberg: This seems tangential to the example, maybe remove?

document/core/syntax/modules.rst
```diff
@@ -112,16 +120,17 @@ Conventions
...
+.. index:: ! type definition, type index, function type, event type
    pair: abstract syntax; type definition
 .. _syntax-typedef:
 
 Types
 ~~~~~
 
-The |MTYPES| component of a module defines a vector of :ref:`function types <syntax-functype>`.
+The |MTYPES| component of a module defines a vector of :ref:`function types <syntax-functype>`,
+which may also be part of an :ref:`event type <syntax-eventtype>`.
```
rossberg:
```suggestion
The |MTYPES| component of a module defines a vector of :ref:`function types <syntax-functype>`.
```
Function types have a number of uses, e.g., they can also describe block types or the types of instructions. So I wouldn't bother mentioning event types specifically at this point.

document/core/syntax/modules.rst
```diff
@@ -112,16 +120,17 @@ Conventions
...
 
-All function types used in a module must be defined in this component.
+All function types and all event function types used in a module must be defined in this component.
```
rossberg:
```suggestion
All function types used in a module must be defined in this component.
```

document/core/syntax/types.rst
```diff
@@ -54,13 +54,15 @@ Reference Types
 .. math::
    \begin{array}{llll}
    \production{reference type} & \reftype &::=&
-     \FUNCREF ~|~ \EXTERNREF \\
+     \FUNCREF ~|~ \EXTERNREF ~|~ \EXNREF \\
```
rossberg:
```suggestion
     \FUNCREF ~|~ \EXNREF ~|~ \EXTERNREF \\
```

document/core/appendix/algorithm.rst
```diff
@@ -25,13 +25,13 @@ Types are representable as an enumeration.
 
 .. code-block:: pseudo
 
-   type val_type = I32 | I64 | F32 | F64 | Funcref | Externref
+   type val_type = I32 | I64 | F32 | F64 | Funcref | Externref | Exnref
```
```suggestion
   type val_type = I32 | I64 | F32 | F64 | Funcref | Exnref | Externref
```

document/core/appendix/algorithm.rst
```diff
@@ -25,13 +25,13 @@ Types are representable as an enumeration.
...
 
    func is_num(t : val_type) : bool =
      return t = I32 || t = I64 || t = F32 || t = F64
 
    func is_ref(t : val_type) : bool =
-     return t = Funcref || t = Externref
+     return t = Funcref || t = Externref || t = Exnref
```
```suggestion
     return t = Funcref || t = Exnref || t = Externref
```

document/core/syntax/types.rst
```diff
@@ -54,13 +54,15 @@ Reference Types
...
    \end{array}
 
 The type |FUNCREF| denotes the infinite union of all references to :ref:`functions <syntax-func>`, regardless of their :ref:`function types <syntax-functype>`.
 
 The type |EXTERNREF| denotes the infinite union of all references to objects owned by the :ref:`embedder <embedder>` and that can be passed into WebAssembly under this type.
 
+The type |EXNREF| denotes the infinite union of all references to exception :ref:`events <syntax-event>`, regardless of their :ref:`event types <syntax-eventtype>`.
```
rossberg:
```suggestion
The type |EXNREF| denotes the a caught exception :ref:`event <syntax-event>`.
```
(there is only one such type, so no union.)
Nit: also, move before externref paragraph.

document/core/text/instructions.rst
```diff
@@ -106,6 +115,9 @@ All other control instruction are represented verbatim.
    \production{plain instruction} & \Tplaininstr_I &::=&
      \text{unreachable} &\Rightarrow& \UNREACHABLE \\ &&|&
      \text{nop} &\Rightarrow& \NOP \\ &&|&
+     \text{throw}~~x{:}\Teventidx_I &\Rightarrow& \THROW~x \\ &&|&
+     \text{rethrow} &\Rightarrow& \RETHROW \\ &&|&
+     \text{br\_on\_exn}~~l{:}\Tlabelidx_I~~x{:}\Teventidx_I &\Rightarrow& \BRONEXN~l~x \\ &&|&
```
rossberg: In line with the direction we have taken for other similar cases, the text format syntax for this instruction should probably be
```
br_on_exn $l (exception $x)
```

document/core/text/instructions.rst
```diff
@@ -561,7 +573,10 @@ Such a folded instruction can appear anywhere a regular instruction can.
      \text{(}~\text{if}~~\Tlabel~~\Tblocktype~~\Tfoldedinstr^\ast
        &\hspace{-3ex} \text{(}~\text{then}~~\Tinstr_1^\ast~\text{)}~~\text{(}~\text{else}~~\Tinstr_2^\ast~\text{)}^?~~\text{)}
        \quad\equiv \\ &\qquad
-         \Tfoldedinstr^\ast~~\text{if}~~\Tlabel~~\Tblocktype &\hspace{-1ex} \Tinstr_1^\ast~~\text{else}~~(\Tinstr_2^\ast)^?~\text{end} \\
+         \Tfoldedinstr^\ast~~\text{if}~~\Tlabel~~\Tblocktype &\hspace{-1ex} \Tinstr_1^\ast~~\text{else}~~(\Tinstr_2^\ast)^?~\text{end} \\ &
+     \text{(}~\text{try}~~\Tlabel~~\Tblocktype~~\Tinstr_1^\ast
```
rossberg: I think we recently decided that an additional `do` is to be inserted:
```suggestion
     \text{(}~\text{try}~~\Tlabel~~\Tblocktype~~\text{(}~\text{do}~~\Tinstr_1^\ast~~\text{)}
```

document/core/text/modules.rst
```diff
@@ -380,6 +387,56 @@ Memories can be defined as :ref:`imports <text-import>` or :ref:`exports <text-e
...
+
+.. math::
+   \begin{array}{llcl}
+   \production{event} & \Tevent_I &::=&
+     \text{(}~\text{exception}~~\Tid^?~~x,I'{:}\Ttypeuse_I~\text{)} \\ &&& \qquad
+       \Rightarrow\quad \{ \EVATTRIBUTE~\EXCEPTION, \EVTYPE~x \} \\
+   \end{array}
+
+.. note::
+   The syntax for events just covers the only possible event attribute, that of an exception.
```
rossberg:
```suggestion
```

document/core/text/types.rst
```diff
@@ -34,10 +34,12 @@ Reference Types
    \begin{array}{llcll@{\qquad\qquad}l}
    \production{reference type} & \Treftype &::=&
      \text{funcref} &\Rightarrow& \FUNCREF \\ &&|&
-     \text{externref} &\Rightarrow& \EXTERNREF \\
+     \text{externref} &\Rightarrow& \EXTERNREF \\ &&|&
+     \text{exnref} &\Rightarrow& \EXNREF \\
```
rossberg:
```suggestion
     \text{exnref} &\Rightarrow& \EXNREF \\ &&|&
     \text{externref} &\Rightarrow& \EXTERNREF \\
```

document/core/text/types.rst
```diff
@@ -34,10 +34,12 @@ Reference Types
...
    \production{referenced type} & \Trefedtype &::=&
      \text{func} &\Rightarrow& \FUNCREF \\ &&|&
-     \text{extern} &\Rightarrow& \EXTERNREF \\
+     \text{extern} &\Rightarrow& \EXTERNREF \\ &&|&
+     \text{event} &\Rightarrow& \EXNREF \\
```
rossberg:
```suggestion
     \text{exn} &\Rightarrow& \EXNREF \\ &&|&
     \text{extern} &\Rightarrow& \EXTERNREF \\
```
(should match the prefix in `exnref`).

document/core/text/modules.rst
```diff
@@ -380,6 +387,56 @@ Memories can be defined as :ref:`imports <text-import>` or :ref:`exports <text-e
...
+
+Abbreviations
+.............
+
+Events can be defined as :ref:`imports <text-import>` or :ref:`exports <text-export>` inline:
+
+.. math::
+   \begin{array}{llclll}
+   \production{module field} &
+     \text{(}~\text{exception}~~\Tid^?~~\text{(}~\text{import}~~\Tname_1~~\Tname_2~\text{)}~~\Teventtype~\text{)} \quad\equiv \\ & \qquad
```
rossberg: Something seems wrong here: the `Teventtype` as described would require another `exception` keyword. Instead, there has to be a typeuse here (cf. the analogous abbrevs for functions).

document/core/text/modules.rst
```diff
@@ -380,6 +387,56 @@ Memories can be defined as :ref:`imports <text-import>` or :ref:`exports <text-e
...
+       \text{(}~\text{import}~~\Tname_1~~\Tname_2~~\text{(}~\text{exception}~~\Tid^?~~\Teventtype~\text{)}~\text{)}
```
rossberg: Same here.

document/core/text/types.rst
```diff
@@ -138,6 +140,21 @@ Table Types
...
+   pair: text format; exception attribute
+.. _text-eventtype:
+
+Event Types
+~~~~~~~~~~~
+
+.. math::
+   \begin{array}{llclll}
+   \production{event type} & \Teventtype &::=&
+     \text{(}~\text{exception}~~t^\ast{:}\Tvec(\Tparam)~\text{)} &\Rightarrow& \EXCEPTION~~[t^\ast] \to [] \\
```
rossberg: I'm not sure where this would actually be used in the text format, unless you define it as a synonym for a `typeuse`.

document/core/valid/modules.rst
```diff
@@ -98,6 +98,36 @@ Memories :math:`\mem` are classified by :ref:`memory types <syntax-memtype>`.
...
+.. _valid-event:
+
+Events
+~~~~~~
+
+Events :math:`\event` are classified by the contents of their :ref:`event types <syntax-eventtype>`.
+Currently the only allowed events are exceptions, i.e., events with type of the form :math:`\EXCEPTION~[t^\ast]\to[]`.
+
+:math:`\{ \EVATTRIBUTE~\EXCEPTION, \EVTYPE~x \}`
+.................................................
```
rossberg:
```suggestion
................................................
```

document/core/valid/modules.rst
```diff
@@ -476,6 +521,26 @@ Imports :math:`\import` and import descriptions :math:`\importdesc` are classifi
...
+
+* The :ref:`event type <syntax-eventtype>` :math:`\EXCEPTION~C.\CTYPES[x]` must be a :ref:`valid event type <valid-eventtype>`.
+
+* Then the import description is valid with type :math:`\ETEVENT~\EXCEPTION~C.\CTYPES[x]`.
+
+.. math::
+   \frac{
+     \vdasheventtype \EXCEPTION~C.\CTYPES[x] \ok
+   }{
+     C \vdashimportdesc \IDEVENT~\{ \EVATTRIBUTE~\EXCEPTION, \EVTYPE~x \}: \ETEVENT~\EXCEPTION~C.\CTYPES[x]
```
rossberg:
```suggestion
     C \vdashimportdesc \IDEVENT~\{ \EVATTRIBUTE~\EXCEPTION, \EVTYPE~x \} : \ETEVENT~\EXCEPTION~C.\CTYPES[x]
```

document/core/valid/modules.rst
```diff
@@ -560,6 +630,9 @@ Instead, the context :math:`C` for validation of the module's content is constru
   * For each :math:`\mem_i` in :math:`\module.\MMEMS`,
     the definition :math:`\mem_i` must be :ref:`valid <valid-mem>` with a :ref:`memory type <syntax-memtype>` :math:`\X{mt}_i`.
 
+  * For each :math:`\event_i` in :math:`\module.\MEVENTS`,
+    the definition :math:`\event_i` must be :ref:`valid <valid-event>` with a :ref:`event type <syntax-eventtype>` :math:`\X{ev}_i`.
```
rossberg:
```suggestion
    the definition :math:`\event_i` must be :ref:`valid <valid-event>` with an :ref:`event type <syntax-eventtype>` :math:`\X{ev}_i`.
```

document/core/valid/modules.rst
```diff
@@ -590,6 +663,8 @@ Instead, the context :math:`C` for validation of the module's content is constru
 
 * Let :math:`\X{mt}^\ast` be the concatenation of the internal :ref:`memory types <syntax-memtype>` :math:`\X{mt}_i`, in index order.
 
+* Let :math:`\X{ev}^\ast` be the concatenation of the internal :ref:`event types <syntax-eventtype>` :math:`\X{ev}_i`, in index order.
```
rossberg:
```suggestion
* Let :math:`\X{evt}^\ast` be the concatenation of the internal :ref:`event types <syntax-eventtype>` :math:`\X{evt}_i`, in index order.
```

document/core/valid/modules.rst
```diff
@@ -611,32 +686,37 @@ Instead, the context :math:`C` for validation of the module's content is constru
      \quad
      (C \vdashmem \mem : \X{mt})^\ast
      \quad
-     (C' \vdashglobal \global : \X{gt})^\ast
+     (C \vdashevent \event : \X{ev})^\ast
```
rossberg:
```suggestion
     (C \vdashevent \event : \X{evt})^\ast
```

document/core/valid/modules.rst
```diff
@@ -611,32 +686,37 @@ Instead, the context :math:`C` for validation of the module's content is constru
...
      \X{imt}^\ast = \etmems(\X{it}^\ast)
+     \\
+     \X{iev}^\ast = \etevents(\X{it}^\ast)
      \qquad
      \X{igt}^\ast = \etglobals(\X{it}^\ast)
      \\
      x^\ast = \freefuncidx(\module \with \MFUNCS = \epsilon \with \MSTART = \epsilon)
      \\
-     C = \{ \CTYPES~\type^\ast, \CFUNCS~\X{ift}^\ast\,\X{ft}^\ast, \CTABLES~\X{itt}^\ast\,\X{tt}^\ast, \CMEMS~\X{imt}^\ast\,\X{mt}^\ast, \CGLOBALS~\X{igt}^\ast\,\X{gt}^\ast, \CELEMS~\X{rt}^\ast, \CDATAS~{\ok}^n, \CREFS~x^\ast \}
+     C = \{ \CTYPES~\type^\ast, \CFUNCS~\X{ift}^\ast\,\X{ft}^\ast, \CTABLES~\X{itt}^\ast\,\X{tt}^\ast, \CMEMS~\X{imt}^\ast\,\X{mt}^\ast, \CEVENTS~\X{iev}^\ast\,\X{ev}^\ast,\\
```
rossberg:
```suggestion
     C = \{ \CTYPES~\type^\ast, \CFUNCS~\X{ift}^\ast\,\X{ft}^\ast, \CTABLES~\X{itt}^\ast\,\X{tt}^\ast, \CMEMS~\X{imt}^\ast\,\X{mt}^\ast, \CEVENTS~\X{iev}^\ast\,\X{evt}^\ast,\\
```

document/core/valid/modules.rst
```diff
@@ -647,6 +727,7 @@ Instead, the context :math:`C` for validation of the module's content is constru
          \MFUNCS~\func^\ast,
          \MTABLES~\table^\ast,
          \MMEMS~\mem^\ast,
+	 \MEVENTS~\event^\ast,
```
rossberg:
```suggestion
         \MEVENTS~\event^\ast,
```

rossberg (commented):
Great stuff! Looks pretty good.

Apart from my comments, I wonder at this point if it's really worth having all the extra verbosity of defining "events" as a possible generalisation of exceptions already. It creates a lot of extra hoops, and we don't know when we will have the generalisation. I think it would totally suffice if the spec removed any mention of events and just spoke of exceptions, merely leaving the extension point in the binary format (the attribute 0-byte) for now. That would make the spec considerably simpler and more uniform (there is just one concept). But that's best left for a follow-up.

ioannad:
Thank you for your careful and clarifying review @rossberg. I applied your suggestions and will address the rest of your comments in a separate commit.

document/core/text/instructions.rst
ioannad: I'm not sure which other cases you are referring to. In this list of plain instruction productions, the other instructions are mixed: function and table indices are written as plain indices, only typeuses are surrounded in "(type ...)". But I agree that exception indices are more like type indices, than function or table indices.
Having said that, I think [wast expects br_on_exn](https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wast/src/ast/expr.rs#L1352) to be followed by two indices, a label and an exception index. Shall I make the change you suggest anyway?

document/core/text/types.rst
ioannad: Oh, this was just a remnant of having a more general event text format previously. I'm removing it.

document/core/appendix/algorithm.rst
```diff
@@ -210,6 +210,15 @@ Other instructions are checked in a similar manner.
          error_if(frame.opcode =/= if)
          push_ctrl(else, frame.start_types, frame.end_types)
 
+       case (try t1*->t2*)
+         pop_vals([t1*])
+	 push_ctrl(try, [t1*], [t2*])
```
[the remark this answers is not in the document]
ioannad: Apologies for the tabs! :grimacing: Something was wrong with my emacs init.

document/core/exec/runtime.rst
```diff
@@ -303,6 +315,31 @@ The bytes can be mutated through :ref:`memory instructions <syntax-instr-memory>
...
+   \production{(event instance)} & \eventinst &::=&
+     \{ \EVIATTRIBUTE~\EXCEPTION, \EVITYPE~\functype \} \\
+   \end{array}
+
+Because :ref:`events <syntax-event>` :ref:`have type <valid-event>` |eventtype|, it is an invariant of the semantics that the function type :math:`\functype` has void result type.
+
+.. note::
+   In the current version of WebAssembly, events may only be exceptions. This may change in future versions.
+
+   The event address of an event instance is also called an exception tag.
```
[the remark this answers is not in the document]
ioannad: I'm changing this to emphasis, so `*exception tag*`,  or did you mean something else?

document/core/exec/runtime.rst
ioannad: I added the bt expansion to emphasize that this instruction will return the exception reference as a value, not throw it. Then I guess I got carried away and added all the details.  Since my initial point is also spelled out in the end of the example, I will remove these lines as you suggest.

document/core/exec/instructions.rst
```diff
@@ -1615,6 +1615,117 @@ Control Instructions
...
+
+2. Let :math:`[t_1^n] \to [t_2^m]` be the :ref:`function type <syntax-functype>` :math:`\expand_F(\blocktype)`.
+
+3. Assert: due to :ref:`validation <valid-if>`, there are at least :math:`n` values on the top of the stack.
+
+4. Pop the values :math:`\val^n` from the stack.
+
+5. Let :math:`L` be the label whose arity is :math:`m` and whose continuation is the end of the |TRY| instruction.
+
+4. Execute the administrative instruction :math:`\CATCHN_m` by :ref:`entering <exec-instr-seq-enter>` the block :math:`\instr_1^\ast` with label :math:`L`, and reducing the resulting :ref:`throw context <syntax-ctxt-throw>`.
```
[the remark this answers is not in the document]
ioannad: I was trying to point to the reduction rule for `catch_n`, which now I see is strange, considering I didn't do the same for `throwaddr` in the last rule of `throw x`. Removing this and pointing "Execute" to a new place in exec/instructions containing exception-related reduction rules, as you suggested in another comment.

document/core/appendix/algorithm.rst
```diff
@@ -210,6 +210,15 @@ Other instructions are checked in a similar manner.
...
+
+       case (catch)
+         let frame = pop_ctrl()
+	 error_if(frame.opcode =/= try)
+	 push_ctrl(catch, [exnref], frame.end_types)
```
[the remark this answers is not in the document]
ioannad: Done

document/core/binary/types.rst
```diff
@@ -168,3 +169,28 @@ Global Types
...
+Event Types
+~~~~~~~~~~~
+
+:ref:`Event types <syntax-eventtype>` are encoded as function types with a preceding flag indicating their :ref:`attribute <syntax-attribute>`.
+
+.. math::
+   \begin{array}{llclll}
+   \production{event type} & \Beventtype &::=&
+     a{:}\Battribute~~ft{:}\Bfunctype &\Rightarrow& a~ft \\
+   \production{attribute} & \Battribute &::=&
```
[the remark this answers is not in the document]
ioannad: Done, everywhere.

document/core/exec/instructions.rst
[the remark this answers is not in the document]
ioannad: Now "Execute" is pointing to the new section "Exception Handling" of this file, and the string you objected to is removed.

document/core/exec/runtime.rst
```diff
@@ -37,11 +38,15 @@ or *external references* pointing to an uninterpreted form of :ref:`extern addre
...
-     \REFEXTERNADDR~\externaddr \\
+     \REFEXTERNADDR~\externaddr \\&&|&
+     \EXNREFADDR~\eventaddr~\val^\ast \\
    \production{(value)} & \val &::=&
      \num ~|~ \reff \\
    \end{array}
 
+The administrative value |EXNREFADDR| is an exception reference of type |EXNREF|, representing a thrown exception on the stack, identified by the exception's :ref:`event address <syntax-eventaddr>`, and carrying the exception event's arguments.
+If it is uncaught by a |CATCHN| block, then the embedder defines how to handle it. It is an abstract value, and it is an implementation detail how and where its argument values are stored. It is separate from any linear memory.
+
```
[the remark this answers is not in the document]
ioannad: Done

document/core/exec/runtime.rst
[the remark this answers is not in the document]
ioannad: Changed to `*exception tag*`

document/core/exec/runtime.rst
```diff
@@ -510,11 +551,14 @@ Conventions
    \end{array}
 
 
-.. index:: ! administrative instructions, function, function instance, function address, label, frame, instruction, trap, call, memory, memory instance, table, table instance, element, data, segment
+.. index:: ! administrative instructions, function, function instance, function address, label, frame, instruction, trap, call, memory, memory instance, table, table instance, element, data, segment, event, event instance, event address, exception, reftype
    pair:: abstract syntax; administrative instruction
 .. _syntax-trap:
 .. _syntax-reffuncaddr:
 .. _syntax-invoke:
+.. _syntax-exnrefaddr:
```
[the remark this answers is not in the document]
ioannad: Done, everywhere.

document/core/exec/runtime.rst
ioannad: Done.

document/core/exec/runtime.rst
ioannad: Done.

document/core/text/modules.rst
ioannad: Done

document/core/text/modules.rst
ioannad: Done

document/core/text/types.rst
ioannad: Done

ioannad:
@rossberg I addressed all your comments, except [the text format of br_on_exn](https://github.com/WebAssembly/exception-handling/pull/121#discussion_r443684166). If that text format is `br_on_exn l (exception x)`, wouldn't it be confusing that `x` is a `typeidx` instead of an `eventidx` (soon to be `exnidx`)? 

The built spec is also updated to reflect all the changes: https://ioannad.github.io/exception-handling/core/

document/core/text/instructions.rst
ioannad: Moreover (as I said in my [last comment](https://github.com/WebAssembly/exception-handling/pull/121#issuecomment-647719174)), if that text format is `br_on_exn l (exception x)`, wouldn't it be confusing that `x` is a `typeidx` instead of an `eventidx`?

document/core/exec/instructions.rst
```diff
@@ -1624,13 +1624,13 @@ Control Instructions
...
 
-3. Assert: due to :ref:`validation <valid-if>`, there are at least :math:`n` values on the top of the stack.
+3. Assert: due to :ref:`validation <valid-try>`, there are at least :math:`n` values on the top of the stack.
 
 4. Pop the values :math:`\val^n` from the stack.
 
 5. Let :math:`L` be the label whose arity is :math:`m` and whose continuation is the end of the |TRY| instruction.
 
-4. Execute the administrative instruction :math:`\CATCHN_m` by :ref:`entering <exec-instr-seq-enter>` the block :math:`\instr_1^\ast` with label :math:`L`, and reducing the resulting :ref:`throw context <syntax-ctxt-throw>`.
+4. :ref:`Execute <exec-catchn>` the administrative instruction :math:`\CATCHN_m` by :ref:`entering <exec-instr-seq-enter>` the block :math:`\instr_1^\ast` with label :math:`L`.
```
rossberg: The prose tries to avoid referring to administrative instructions, so I would express this a bit differently. I would phrase it as pushing and popping a handler on/off the stack, similar to how the prose describes function frames. That is, describe handlers as a new form of stack element in the Stack section (runtime.rst), and in the new auxiliary section Exception Handling have subsections for "installing" and exiting a handler, analogous to the ones for invoking functions.

document/core/syntax/modules.rst
```diff
@@ -241,10 +240,10 @@ The |MEVENTS| component of a module defines a vector of *events* with the follow
 
 .. math::
    \begin{array}{llll}
-   \production{event} & \event &::=& \{ \EVATTRIBUTE~\EXCEPTION, \EVTYPE~\typeidx \} \\
+   \production{event} & \event &::=& \{ \EVATTR~\EXCEPTION, \EVTYPE~\typeidx \} \\
    \end{array}
 
-The event's :ref:`attribute <syntax-attribute>` may only specify the attribute |EXCEPTION|, meaning that the event is an exception.
+The :ref:`event attribute <syntax-eventattr>` may only specify the attribute |EXCEPTION|, meaning that the event is an exception.
```
rossberg: As long as it basically carries no information (there is only one possible value), I'd completely remove event attributes from the abstract syntax for the time being. For future extensibility it suffices to have the zero byte in the binary format.

document/core/text/instructions.rst
rossberg: Yeah, I'm fine addressing this separately, since it requires a change to what's currently implemented.

The idea is that in all cases where an instruction has an index that is not of the "primary" sort of the construct, the text format makes the kind explicit. We do that e.g. for `call_indirect (type $x)`, `data (memory $x)`, `elem (table $x)`, etc., and there will probably be more in the future.

> if that text format is `br_on_exn l (exception x)`, wouldn't it be confusing that `x` is a `typeidx` instead of an `eventidx`?

Not sure I follow, `x` is an eventidx/exnidx, isn't it?

rossberg (commented):
Thanks! Can you rename event{idx,addr,instance,...} to exn{idx,addr,instance,...} as well? As far as I am concerned, there currently is no compelling reason to use the word event at all, since all we provide is exceptions. The redundant naming may be unnecessarily confusing.

document/core/syntax/modules.rst
ioannad: Sure, no problem at all to completely remove "event" everywhere (except the binary format byte you mentioned). I just thought that in your [first comment ](https://github.com/WebAssembly/exception-handling/pull/121#pullrequestreview-435019490) you suggested to do this in a different PR, that is the reason why I didn't do that yet. 

The next commit will remove all occurrences of event. 

Just to make sure I understand correctly, I should also rename "event section" to "exception section", so really leave no trace of "event" except in that binary format bit, correct?

document/core/text/instructions.rst
ioannad: > Not sure I follow, x is an eventidx/exnidx, isn't it?

Yes it is an eventidx/exnidx, sorry for not being clear enough. I was also wrong in my assumption that it would be confusing. What I meant was that it would be confusing if an event/exception type were defined with `(exception typeidx)` but that is wrong, it's defined with `(exception (type typeidx))` so please ignore this comment.

I would change this in the next commit as well to what you suggested, and then mention that change at the wast repository. Other implementations would notice the change when tests are added the latest.

document/core/syntax/modules.rst
rossberg: Ah, okay, I wasn't sure what the plan was, since I think there were some changes already. Anyway, I think we're on the same page, so I'm fine either way.

Yes, I'd rename the event section as well (for now).

document/core/text/instructions.rst
ioannad: For some reason this comment doesn't seem to appear at the main discussion, so here's a tag @rossberg, in case you haven't seen this.

rossberg: Thanks, SGTM.

ioannad:
Sorry for the delay, @rossberg. I think the commit I just pushed should implement your latest comments and suggestions. Moreover, I made some changes to `proposal/Exceptions.md` which you didn't request, but which I thought would make sense after the discussion here. In particular, I removed "events" from the abstract syntax and text format, and changed the text format of `br_on_exn`, and additionally of `throw`. The commit message is fully detailed.

As before, I rebuilt the spec of my fork to include these changes: https://ioannad.github.io/exception-handling/core/

document/core/valid/types.rst
```diff
@@ -146,28 +146,26 @@ Memory Types
...
-:math:`\EXCEPTION~[t^n] \to []`
-...............................
+Exception Types
+~~~~~~~~~~~~~~~
 
-* The :ref:`event attribute <syntax-eventattr>` must be |EXCEPTION|.
+:math:`[t^n] \to []`
+....................
 
 * The :ref:`function type <syntax-functype>` :math:`[t^n] \to []` must be :ref:`valid <valid-functype>`.
```
rossberg:
```suggestion
* The :ref:`function type <syntax-functype>` :math:`[t_1^n] \to [t_2^m]` must be :ref:`valid <valid-functype>`.

* The type sequence :math:`t_2^m` must be empty.
```

proposals/Exceptions.md
```diff
@@ -117,21 +117,22 @@ Each event has an `attribute` and a `type`. Currently, the attribute can only
...
 values may be added when other events are added to WebAssembly.
 
-The type of an event is denoted by an index to a function signature defined in
-the `type` section. The parameters of the function signature defines the list of
-values associated with the exception event. The result type must be 'void'.
-
-An `event tag` is a value to distinguish different events, while an `event
-index` is a numeric name to refer to an (imported or defined) event tag within a
-module (see [event index space](#event-index-space) for details).
+To allow for such a future extension possibility, we reserve a bit in the binary
```
rossberg:
```suggestion
To allow for such a future extension possibility, we reserve a byte in the binary
```

proposals/Exceptions.md
```diff
@@ -117,21 +117,22 @@ Each event has an `attribute` and a `type`. Currently, the attribute can only
...
 An `exception` is an internal construct in WebAssembly . WebAssembly exceptions
-are defined in the `event` and import sections of a module. Each event (with an
-exception attribute) defines an `exception`. The event index is also called the
-`exception index`. Similarly, the corresponding event tag is called an
-`exception tag`.
+are defined in the `exception` and import sections of a module.
+
+The type of an exception is denoted by an index to a function signature defined in
+the `type` section. The parameters of the function signature define the list of
+values associated with the exception. The result type must be 'void'.
```
rossberg:
```suggestion
values associated with the exception. The result type must be empty.
```

document/core/exec/instructions.rst
```diff
@@ -1630,7 +1630,9 @@ Control Instructions
 
 5. Let :math:`L` be the label whose arity is :math:`m` and whose continuation is the end of the |TRY| instruction.
 
-4. :ref:`Execute <exec-catchn>` the administrative instruction :math:`\CATCHN_m` by :ref:`entering <exec-instr-seq-enter>` the block :math:`\instr_1^\ast` with label :math:`L`.
+6. Let :math:`H` be the exception handler whose arity is :math:`m` and whose continuation is the beginning of :math:`\instr_2^\ast`.
+
+7. :ref:`Install <exec-install-handler>` the exception handler `H` for the block :math:`\val^n~\instr_1^\ast` with label :math:`L`.
```
rossberg: I'd separate this into installing the handler and entering the block, corresponding to the two separate administrative instructions it reduces to:
```suggestion
7. :ref:`Enter <exec-handler-enter>` the exception handler `H`.

8. :ref:`Enter <exec-instr-seq-enter>` the block :math:`\val^n~\instr_1^\ast` with label :math:`L`.
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
 
 The following auxiliary rules define the semantics of handling thrown exceptions
-inside :ref:`throw contexts <syntax-ctxt-throw>`.
+inside :ref:`throw contexts <syntax-ctxt-throw>`, such as installing, searching for,
+and exiting an :ref:`exception handler <syntax-handler>` :math:`\CATCHN_n`.
+
+.. _exec-install-handler:
+
+Installing an exception handler :math:`H` for a block :math:`\instr^\ast` with label :math:`L`
+..............................................................................................
```
```suggestion
Entering an exception handler :math:`H`
.......................................
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+1. Let :math:`H` be the exception handler :math:`\CATCHN_m{\instr'^\ast}`.
+
+2. Push :math:`H` on the stack.
+
+3. :ref:`Enter <exec-instr-seq-enter>` the block :math:`\instr^\ast` with label :math:`L`.
+
```
```suggestion
1. Push :math:`H` onto the stack.
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+.. note::
+   No formal reduction rule is needed for installing an exception handler
+   because it is an :ref:`administrative instruction <syntax-instr-admin>`
+   which exception throwing instructions will search for.
+
+
+Exiting an exception handler with arity :math:`m` from a block with label :math:`L`
+...................................................................................
```
```suggestion
Exiting an exception handler
............................
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+When the block is exited without a throw :ref:`triggering <exec-handle-exn>`
```
```suggestion
When the end of a :ref:`try <syntax-try>` instruction is reached without a jump, exception or trap, then the following steps are performed.
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+the exception handler, then the following steps are performed.
+
+1. Assert: due to validation, there are :math:`m` values on the top of the stack.
```
```suggestion
1. Let :math:`m` be the number of values on the top of the stack.
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+2. Pop the values :math:`\val^m` from the stack.
+
+3. Assert: due to :ref:`validation <valid-instr-seq>`, the exception handler :math:`H` is now on the top of the stack.
```
```suggestion
3. Assert: due to :ref:`validation <valid-instr-seq>`, the handler :math:`H` is now on the top of the stack.
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+4. Pop the exception handler from the stack.
```
```suggestion
4. Pop the handler from the stack.
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+5. Push :math:`\val^m` back to the stack.
+
+6. Jump to the position after the |END| of the originating |TRY| instruction associated with the label :math:`L`.
```
```suggestion
6. Jump to the position after the |END| of the originating |TRY| instruction associated with the handler :math:`H`.
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+.. math::
+   ~\\[-1ex]
+   \begin{array}{lcl@{\qquad}l}
+   \CATCHN_m\{instr^\ast\}~\val^m~\END &\stepto& \val^m
+   \end{array}
+
+
+.. _exec-search-handler:
```
```suggestion
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+.. _exec-throwaddr:
+
+Searching for an exception handler for the :ref:`exception address <syntax-exnaddr>` :math:`a`
+..............................................................................................
```
```suggestion
Throwing an exception with :ref:`exception address <syntax-exnaddr>` :math:`a`
..............................................................................
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+When a throw or a rethrow occurs, labels and call frames are popped if necessary,
+until an exception handler is found on the top of the stack.
+
+1. Assert: due to validation, :math:`S.\SEXNS[a]` exists.
+
+2. Let :math:`[t^n] \to []` be the type of the exception instance :math:`S.\SEXNS[a]`.
```
```suggestion
2. Let :math:`[t^n] \to [t'^m]` be the :ref:`exception type <syntax-exntype>` :math:`S.\SEXNS[a].\EITYPE`.
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+3. Assert: due to validation, there are :math:`n` values on the top of the stack.
```
```suggestion
3. Assert: due to :ref:`validation <valid-try>`, there are :math:`n` values on the top of the stack.
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+4. Pop the :math:`n` values :math:`\val^n` from the stack.
+
+5. While the stack is not empty and the top of the stack is not an exception handler, do:
+
+   a. Pop the top element from the stack.
+
+6. The stack is now either empty or there is an exception handler on the top.
```
```suggestion
6. Assert: The stack is now either empty or there is an exception handler on the top.
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+.. _exec-handle-exn:
```
```suggestion
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+7. If there is an exception handler :math:`\CATCHN_m\{\instr^\ast\}` on the top of the stack, then:
+
+   a. Assert: the last element popped from the stack was the label :math:`L` whose continuation is the end of the originating |TRY| instruction.
+
+   b. Pop the exception handler from the stack.
```
```suggestion
   a. Pop the exception handler from the stack.

   b. Let :math:`L` be the label whose arity is :math:`m` and whose continuation is the end of the |TRY| instruction associated with the handler.
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+   c. Put the label :math:`L` back on the stack.
```
```suggestion
   c. Push the label :math:`L` on the stack.
```

document/core/exec/runtime.rst
```diff
@@ -608,6 +675,53 @@ This definition allows to index active labels surrounding a :ref:`branch <syntax
...
+.. note::
+   For example, catching a simple :ref:`throw <exec-throw>` in a :ref:`try-catch block <exec-try>` would be as follows.
+
+   .. math::
+      \begin{array}{ll}
+      & \hspace{-5ex} S;~F;~\val^n~(\TRY~\X{bt}~\THROW~x~\CATCH~\RETURN~\END) \\
+      \stepto & S;~F;~\CATCHN_1\{\RETURN\}~(\LABEL_1 \{\}~\val^n~\THROWADDR~a~\END)~\END \\
+      \end{array}
+
+   :ref:`Handling <exec-handle-exn>` the thrown exception address :math:`a` in the throw context
```
```suggestion
   :ref:`Handling <exec-handler-throwaddr>` the thrown exception address :math:`a` in the throw context
```

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+
+   d. Enter the block :math:`(\REFEXNADDR~a~\val^n) \instr^\ast` with label :math:`L`.
+
```
```suggestion
   d. Enter the block :math:`\instr^\ast` with label :math:`L`.

   e. Push the :ref:`exception reference <syntax-refexnaddr>` :math:`(\REFEXNADDR~a~\val^n)` to the stack.
```

rossberg (commented):
Sorry for the delay, vacation. ;)

This looks pretty good! I have added a number of suggestions, most prominently for factoring out entering a block from entering a handler, which should make things slightly simpler.

document/core/exec/instructions.rst
```diff
@@ -1885,6 +1998,110 @@ When the end of a block is reached without a jump or trap aborting it, then the
    Therefore, execution of a loop falls off the end, unless a backwards branch is performed explicitly.
 
 
+.. index:: exception handling, throw context
+   pair: handling; exception
+
+.. _exec-catchn:
+.. _exec-exn-handling:
```
```suggestion
```

document/core/exec/instructions.rst
```diff
@@ -1885,6 +1998,110 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+2. Push :math:`H` on the stack.
+
+3. :ref:`Enter <exec-instr-seq-enter>` the block :math:`\instr^\ast` with label :math:`L`.
+
+.. note::
+   No formal reduction rule is needed for installing an exception handler
+   because it is an :ref:`administrative instruction <syntax-instr-admin>`
+   which exception throwing instructions will search for.
+
+
```
rossberg, line 2026:
```suggestion
.. _exec-handler-exit:

```

document/core/util/macros.def
```diff
@@ -903,8 +899,7 @@
 .. |MITYPE| mathdef:: \xref{exec/runtime}{syntax-meminst}{\K{type}}
 .. |MIDATA| mathdef:: \xref{exec/runtime}{syntax-meminst}{\K{data}}
 
-.. |EVIATTR| mathdef:: \xref{exec/runtime}{syntax-eventinst}{\K{eventattr}}
-.. |EVITYPE| mathdef:: \xref{exec/runtime}{syntax-eventinst}{\K{type}}
+.. |EXNITYPE| mathdef:: \xref{exec/runtime}{syntax-exninst}{\K{type}}
```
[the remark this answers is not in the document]
ioannad: EITYPE is already used for an element instance (L. 907)

document/core/valid/types.rst
```diff
@@ -146,28 +146,26 @@ Memory Types
...
-~~~~~~~~~~~
-
-:math:`\EXCEPTION~[t^n] \to []`
-...............................
+Exception Types
+~~~~~~~~~~~~~~~
 
-* The :ref:`event attribute <syntax-eventattr>` must be |EXCEPTION|.
+:math:`[t^n] \to []`
+....................
```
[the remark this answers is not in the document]
ioannad: I'm adding your suggestion to the batch and will fix the prose below in the next commit.

document/core/exec/instructions.rst
```diff
@@ -1998,23 +2000,105 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
-.. _exec-throwaddr:
+.. _exec-exn-handling:
 
 Exception Handling
 ~~~~~~~~~~~~~~~~~~
 
 The following auxiliary rules define the semantics of handling thrown exceptions
-inside :ref:`throw contexts <syntax-ctxt-throw>`.
+inside :ref:`throw contexts <syntax-ctxt-throw>`, such as installing, searching for,
+and exiting an :ref:`exception handler <syntax-handler>` :math:`\CATCHN_n`.
```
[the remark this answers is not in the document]
ioannad: It seems I missed this suggestion, I am incorporating it in the next commit.

document/core/valid/instructions.rst
```diff
@@ -955,19 +955,19 @@ Control Instructions
...
+* The exception :math:`C.\CEXNS[x]` must be :math:`[t^\ast]\to[]`.
 
 * Then the instruction is valid with type :math:`[\EXNREF]\to[\EXNREF]`
 
 .. math::
    \frac{
      C.\CLABELS[l]=[t^\ast]
    \qquad
-     C.\CEVENTS[x]=\EXCEPTION~[t^\ast]\to[]
+     C.\CEXNS[x]=[t^\ast]\to[]
```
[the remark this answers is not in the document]
ioannad: Not sure what happened here, in my current version I only see the wrongly indented `\qquad`, which I fix in the next commit.

document/core/valid/instructions.rst
```diff
@@ -955,19 +955,19 @@ Control Instructions
 
 * The label :math:`C.\CLABELS[l]` must be defined in the context.
 
-* The event :math:`C.\CEVENTS[x]` must be defined in the context.
+* The exception :math:`C.\CEXNS[x]` must be defined in the context.
 
 * Let :math:`[t^\ast]` be the :ref:`result type <syntax-resulttype>` :math:`C.\CLABELS[l]`.
 
-* The event :math:`C.\CEVENTS[x]` must be :math:`\EXCEPTION~[t^\ast]\to[]`.
+* The exception :math:`C.\CEXNS[x]` must be :math:`[t^\ast]\to[]`.
```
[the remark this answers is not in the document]
ioannad: This is saying that the parameters of the exception type are the same as the parameters of the label. 
To make this more clear, I rephrase this in the next commit to:

* Assert: the exception :math:`C.\CEXNS[x]` is :math:`[t'^\ast]\to[]`.

* The type sequence :math:`t'^\ast` must be the same as the type sequence :math:`t^\ast`.

document/core/util/macros.def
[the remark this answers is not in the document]
rossberg: Ah, right. Maybe we should turn that around, since exns are the more canonical thing now, and moreover, the type in an elem inst isn't even the inst's type. So perhaps rename that to EIELEMTYPE?

document/core/valid/types.rst
ioannad: Again, not sure why I missed putting this in the batch of suggestions to apply. :grimacing: Applying it in the next commit.

document/core/valid/instructions.rst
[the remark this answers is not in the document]
rossberg: I believe GH suggestions are still broken in some cases.

document/core/util/macros.def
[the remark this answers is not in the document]
ioannad: Sounds good, it'll be in the next commit.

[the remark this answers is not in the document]
ioannad: Btw, I suppose do the same for ETYPE? so ETYPE-->EELEMTYPE and EXNTYPE-->ETYPE?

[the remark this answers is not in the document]
rossberg: Yes, makes sense.

ioannad:
@rossberg I think all the comments should be now implemented, and as usual, the rebuilt spec is [here](https://ioannad.github.io/exception-handling/core/).

document/core/binary/types.rst
```diff
@@ -178,15 +178,16 @@ Global Types
...
 
 .. math::
    \begin{array}{llclll}
    \production{exception type} & \Bexntype &::=&
-     ft{:}\Bfunctype &\Rightarrow& \hex{00}~ft \\
+     \hex{00}~~ft{:}\Bfunctype &\Rightarrow& ft \\
    \end{array}
 
-The |Bfunctype| of an exception must have void result.
+The |Bfunctype| of an exception must have empty result.
```
rossberg:
```suggestion
The |Bfunctype| of an exception must have an empty result.
```

document/core/exec/instructions.rst
```diff
@@ -2002,49 +2004,43 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
 
+.. _exec-handler-exit:
 
-Exiting an exception handler with arity :math:`m` from a block with label :math:`L`
-...................................................................................
+Exiting an exception handler
+............................
 
-When the block is exited without a throw :ref:`triggering <exec-handle-exn>`
+When the end of a :ref:`try <syntax-try>` instruction is reached without a jump, exception or trap, then the following steps are performed.
```
rossberg: I just noted that a similar language tweak is probably necessary to the intro sentences to the sections on exiting a block and returning from a function (including exceptions in the enumerations of possibilities).

document/core/exec/runtime.rst
```diff
@@ -321,16 +321,14 @@ Exception Instances
 ~~~~~~~~~~~~~~~~~~~
 
 An *exception instance* is the runtime representation of an :ref:`exception <syntax-exn>` definition.
-It records the :ref:`function type <syntax-functype>` of its :ref:`type <syntax-exntype>`.
+It records type :ref:`type <syntax-exntype>` of the exception.
```
```suggestion
It records the type :ref:`type <syntax-exntype>` of the exception.
```

document/core/exec/runtime.rst
```diff
@@ -538,7 +536,7 @@ Exception handlers
 ..................
 
 Like labels, exception handlers carry the return arity :math:`n` of the
-respective try-catch block, and their associated branch *target*, which is
+respective try block, and their associated branch *target*, which is
```
```suggestion
respective |TRY| block, and their associated branch *target*, which is
```

document/core/syntax/types.rst
```diff
@@ -207,7 +207,8 @@ Exception Types
    \production{exception type} &\exntype &::=& \functype \\
    \end{array}
 
-The result type of an exception's function type |functype| must be void.
+Because :ref:`exceptions <syntax-exn>` have a :ref:`valid <valid-exn>` |exntype|,
+it is an invariant of the semantics that the function type :math:`\functype` has empty result type.
 The parameters of |functype| define the list of values associated with the exception.
```
rossberg:
```suggestion
The parameters of |functype| define the list of values associated with the exception.
Furthermore, it is an invariant of the semantics that every |functype| in a :ref:`valid <valid-exntype>` exception type has an empty result type.

.. note::
   Future versions of WebAssembly may allow non-empty result types in exceptions.
```

document/core/exec/instructions.rst
```diff
@@ -2002,49 +2004,43 @@ When the end of a block is reached without a jump or trap aborting it, then the
    pair: handling; exception
 
 .. _exec-catchn:
```
rossberg: Nit: `_exec-catch` (overlooked this one).

rossberg (commented):
Great, thanks! Below a few things I overlooked before.

One further thing I just noted is that the section on typing administrative instructions in appendix/properties is missing the new cases.

document/core/exec/runtime.rst
ioannad: I'll change that to 

> It records the :ref:`type <syntax-exntype>` of the exception.

document/core/syntax/types.rst
ioannad: For some reason I can't add this to the batch, I'll make the change in the next commit.

ioannad:
@rossberg: I applied your latest suggestions and implemented the rest of the comments, and rebuilt the [fork github page](https://ioannad.github.io/exception-handling/core/)

document/core/exec/instructions.rst
```diff
@@ -2027,7 +2027,7 @@ Entering an exception handler :math:`H`
 Exiting an exception handler
 ............................
 
-When the end of a :ref:`try <syntax-try>` instruction is reached without a jump, exception or trap, then the following steps are performed.
+When the end of a :ref:`try <syntax-try>` instruction is reached without a jump, exception, or trap, then the following steps are performed.
```
rossberg, line 2030: I think in this particular place you do _not_ want to include exceptions. ;)

document/core/exec/runtime.rst
```diff
@@ -715,7 +715,7 @@ Throw contexts allow matching the program context around a throw instruction up
    When a throw occurs, execution halts until that throw is the continuation of a throw context in a catching try block.
 
    In this particular case, the exception reference is returned normally, as opposed to being thrown as the result of a plain
-   :math:`\THROW~x` without a catching try-block.
+   :math:`\THROW~x` without a catching try block.
```
```suggestion
   :math:`\THROW~x` without a catching |TRY| block.
```

rossberg (commented):
Looks very good. Bu have you seen my comment regarding the missing typing rules for the new administrative instructions?

document/core/exec/instructions.rst
ioannad: Ooops, slip of the hand. Thanks for spotting it! :blush:

ioannad:
> Looks very good. Bu have you seen my comment regarding the missing typing rules for the new administrative instructions?

Oh, sorry I missed implementing that comment. On it!

ioannad:
@rossberg please ignore my last comment, I was confused by the way github sometimes shows the PR additions after some rounds of reviews and extra commits.
It seems I included the typing rules already in the first commit of this PR: https://github.com/ioannad/exception-handling/blob/06d5b3ff9ae58d56b66a1b6129850f2e457b3309/document/core/appendix/properties.rst#administrative-instructions
Or did I misunderstand your comment and there is something else missing?

document/core/appendix/properties.rst
```diff
@@ -554,6 +577,67 @@ To that end, all previous typing judgements :math:`C \vdash \X{prop}` are genera
    }
 
 
+.. index:: exception address, exception type, function type, value type, exception tag
```
ioannad, line 580: @rossberg : are these the typing rules you are referring to?

document/core/exec/modules.rst
```diff
@@ -277,6 +316,34 @@ New instances of :ref:`functions <syntax-funcinst>`, :ref:`tables <syntax-tablei
...
+5. Append :math:`\exninst` to the |SEXNS| of :math:`S`.
+
+5. Return :math:`a`.
+
+.. math::
+   \begin{array}{rlll}
+   \allocexn(S, \exn, \module) &=& S', \exnaddr \\[1ex]
+   \mbox{where:} \hfill \\
+   \exnaddr &=& |S.\SEXNS| \\
+   \exninst &=& \{ \EITYPE~\functype \} \\
```
rossberg:
```suggestion
   \exninst &=& \{ \EITYPE~\module.\MTYPES[\exn.\ETYPE] \} \\
```

document/core/exec/modules.rst
```diff
@@ -277,6 +316,34 @@ New instances of :ref:`functions <syntax-funcinst>`, :ref:`tables <syntax-tablei
...
+
+4. Let :math:`\exninst` be the :ref:`exception instance <syntax-exninst>` :math:`\{ \EITYPE~\functype \}`.
+
+5. Append :math:`\exninst` to the |SEXNS| of :math:`S`.
+
+5. Return :math:`a`.
+
+.. math::
+   \begin{array}{rlll}
+   \allocexn(S, \exn, \module) &=& S', \exnaddr \\[1ex]
```
rossberg: This signature doesn't match its use in embedding.rst, which passes merely an exntype -- which is necessary, because that's outside any module. So I think this function needs to use that, and allocmodule needs adjusting.

rossberg (approved):
Thanks for the great work! I stumbled over a couple more minor things, but I approved anyway, so you don't need to wait for me looking again.

document/core/appendix/properties.rst
ioannad: The same happened to me when I looked the first time. :)

document/core/exec/instructions.rst
ioannad: @rossberg: on a second look, these are the instructions for exiting the try without an exception being thrown. If an exception is thrown then we are in the case "Throwing an exception ...", right?

rossberg: Right.

ioannad:
Ok, so now all the comments should be implemented, except the first two lines of "Exiting an exception handler" which now reads (first line is the same, second is a clarification):

> When the end of a :ref:`try <syntax-try>` instruction is reached without a jump, exception, or trap, then the following steps are performed.
> If an exception is thrown before the end of the |TRY| instruction, then we are in the case of :ref:`throwing <exec-throwaddr>` an exception.

ioannad:
Thank you so much for all your comments and suggestions @rossberg! :pray:

rossberg:
> When the end of a :ref:`try <syntax-try>` instruction is reached without a jump, exception, or trap, then the following steps are performed.
> If an exception is thrown before the end of the |TRY| instruction, then we are in the case of :ref:`throwing <exec-throwaddr>` an exception.

Hm, that doesn't sound quite right. The ~instruction sequence is~ steps are _not_ performed when an exception occurred instead.

rossberg:
Ah, sorry, I was misparsing. Scratch the previous comment.

document/core/exec/instructions.rst
```diff
@@ -1885,6 +2000,103 @@ When the end of a block is reached without a jump or trap aborting it, then the
...
+   because it is an :ref:`administrative instruction <syntax-instr-admin>`
+   that the :ref:`try <syntax-try>` instruction reduced to directly.
+
+.. _exec-handler-exit:
+
+Exiting an exception handler
+............................
+
+When the end of a :ref:`try <syntax-try>` instruction is reached without a jump, exception, or trap, then the following steps are performed.
+If an exception is thrown before the end of the |TRY| instruction, then we are in the case of :ref:`throwing <exec-throwaddr>` an exception.
```
```suggestion
If an exception is thrown before the end of the |TRY| instruction, then the case of :ref:`throwing <exec-throwaddr>` an exception applies.
```
Or perhaps just remove the sentence -- it's kind of redundant, and we don't have such a note for trap either.

ioannad: sentence removed! :)

aheejin:
Sorry, but given that we might have to change the spec significantly, is merging this now a good idea?

Ms2ger:
I think it's better to have the current proposal merged so we can iterate on it. If we end up fundamentally changing everything about the proposal, we could still revert this change; if not, we at least have something to build on.

PR #136, opened 2020-10-06
Organize old proposals

aheejin:
This renames existing old proposals by adding a version number to
filenames and adds the current (=soon-to-be-previous) proposal to the
old/folder as well.


tlively (approved):
Would it be better to move the explainer rather than copying it? Kind of strange to have two copies of the same document in the repo, even if it's temporary.

aheejin:
But then the next PR will be not a diff against the current version but a whole new document, which I don't think we want?

tlively:
Good point. LGTM!

Issue #133, opened 2020-10-05
Rename catch_br


rossberg:
The name `catch_br` is misleading, since it is not a branch in the regular sense. In particular, `catch_br $l` jumps to a different place than `br $l`. Possible suggestion: `catch_delegate`.

aheejin:
I don't mind either way. Do people have objections or other suggestions for the name? (Do we need a vote for this? I guess not, right?)

ioannad:
> The name `catch_br` is misleading, since it is not a branch in the regular sense. In particular, `catch_br $l` jumps to a different place than `br $l`. Possible suggestion: `catch_delegate`.

I also don't mind changing the name, but I must say I didn't realise that `catch_br $l` branches to a different place than `br $l`. I though we just have to validate that `$l` is indeed a `try` block. Is this not the case?

aheejin:
Yes, I think we should verify it too. `$l` should be attached to a `try`, and `catch_br` (or `catch_delegate`) should transfer, or delegate, the exception searching path to the corresponding `catch`/`unwind` of `try` specified by the label.

Whether `$l` is attached to `try` of `catch` is just a matter of text syntax, because in binary the argument will be an immediate integer.

EDIT: I think this comment is misleading and incorrect; I posted a new explanation here: https://github.com/WebAssembly/exception-handling/issues/133#issuecomment-704525238

rossberg:
@ioannad, a regular `br` jumps past the whole try statement, while `catch_br` goes to its catch/unwind part. The former also carries along potential arguments (depending on the type of the label), whereas the latter is agnostic of the label type.

RossTate:
Taking the "skip over" interpretation of `catch_br` in #130, `catch_br` should not specify what `try` to delegate to; it should specify how many `try`s to skip over. This would also provide an answer to how to specify the caller as whom to delegate to.

rossberg:
It does, but it counts other control constructs as well. That makes finding the target try in the control stack O(1) in an implementation instead of O(N), and doesn't require a separate try stack.

RossTate:
That makes sense. Note then that with the "skip over" interpretation, there is no need for the target to be a `try`. That is, `catch_br` is essentially saying "for any exceptions thrown from in here, pretend their continuation does not include the n+1 immediately enclosing control constructs".

rossberg:
[...]

Issue #131, opened 2020-10-02
Could the different `try` variants be combined?


tlively:
Kudos to @ioannad for the great [write up](https://github.com/WebAssembly/exception-handling/issues/87#issuecomment-702323595) in #87!

One thing that stood out to me was the large number of `try` variants, and I wanted to ask about them in a new issue to avoid cluttering up that other thread. We currently have the following:

```
instr ::= ... | throw x | rethrow
        | try bt instr* (catch exnidx instr*)+ end
        | try bt instr* (catch exnidx instr*)* catch_all instr* end
        | try bt instr* catch_br l
        | try bt instr* unwind instr* end
```

It's clear that try-catch_br needs to be its own separate variant, but I wonder if it would be simpler to combine the other variants:

```
instr ::= ... | throw x | rethrow
        | try bt instr* (catch exnidx instr*)* (catch_all instr*)? (unwind instr*)? end
        | try bt instr* catch_br l
```

One quirk of this formulation is that it would allow for `try bt intr* end`, which is completely useless. I imagine it would just have the semantics of a normal block. The other obvious change is that this formulation would allow a `try` to have both catches and an unwind. I imagine the semantics of that would be similar to the semantics of this:

```
try
  try
    ...
  catch
    ...
  end
unwind
  ...
end
```

Specifically, if the `catch` body rethrows, the `unwind` body is still run afterwards. So this would just allow for saving a level of `try` nesting.

I don't feel strongly about this at all since it is just a surface syntax change, but I think it might be a nice simplification. I'm curious to hear what other people think and whether this change would complicate anything I've overlooked.

ioannad:
Thank you so much @tlively!

I am in favour of your suggestion. I think it would simplify the formal writeup a lot, and, at least at first glance I can't see any complications it could create.

aheejin:
I'd like to keep `try`-`catch` and `try`-`unwind` separate. I don't have a very strong argument to back it, but I think it is generally clearer; when both of them are present, it may not be very obvious which has to be visited first. But yeah, it can be well defined and it may not be a problem.

[...]

Issue #124, opened 2020-07-29
Unwinding as a modifier to branching


RossTate:
I think there's a way to view unwinding as a modifier to branching. But before I get into that, I first want to make sure I'm on the same page as everyone with respect to branching.

## Branching

Contrary to what their names suggest, branching instructions like `br` do not correspond at the assembly level to jumping instructions. Rather, that's only half of the story. The other half has to do with the stack.

Suppose an engine has some type that is managed using reference counting&mdash;call it `refcounted`. Then consider the branch in the following code that calls some function `$foo : [] -> [refcounted]`:
```
(block $target ;; [] -> []
  (call $foo)
  (br $target)
)
```
This branch does not compile to simply a jump to the assembly-code location corresponding to `$target`. It also cleans up the portion of the stack that is not relevant to `$target`. In this case, that involves decrementing the reference count of the `refcounted` value returned by `$foo` that is sitting on the stack. In a single-threaded engine, that might in turn result in freeing the corresponding memory space and consequently require recursively decrementing the counts of any values contained therein. Depending on how large this structure is and how much of it is no longer necessary, this might take a while. So a simple instruction like `br` can actually execute a fair amount behind the scenes depending on how the engine manages resources and the stack.

All this is to illustrate that a label like `$target` more accurately corresponds to an assembly-code location *and* a location within the stack, and likewise an instruction like `br $target` more accurately corresponds to "clean up the stack up to the stack location corresponding to `$target` and then jump to the assembly-code location corresponding to `$target`".

Note, though, that "clean up" here only corresponds to the engine's resources on the stack. But what about the application's resources? That's where unwinding comes into play.

## Unwinding

For the sake of this discussion, I am going to say that unwinders are specified using `try instr1* unwind instr2* end`, which executes `instr*` but indicates that `instr2* : [] -> []` should be used to "unwind the stack", i.e. to perform application-level clean up. In a second, I'll get to how one causes the stack to be "unwound" rather than just "cleaned up".

Now consider some surface-level code using the common `finally` construct:
```
while (true) {
  File file = open(...);
  try {
    try {
      if (...)
        break;
    } catch (SomeException) {}
  } finally {
    close(file);
  }
}
```
[...]

PR #137, opened 2020-10-07
Update explainer for the new spec

aheejin:
This updates the explainer text according to the new spec we agreed in
the 09-15-2020 CG meeting. Some of the text is taken from [the first
version](https://github.com/WebAssembly/exception-handling/blob/master/proposals/old/Exceptions-v1.md)
of the proposal. I renamed `catch_br` to `delegate` based on the
perceived consensus in #133.

There are some minor issues that need clarification, some of which is
currently being discussed in the repo:
- `rethrow`'s immmediate (or label) argument. I made it match the status
  in the first proposal for now. I believe keeping it makes it more
  versatile and lets us handle more languages.
- It is not clear whether `rethrow` and `delegate`'s label (=immediate)
  argument should count non-try block-like structures, such as blocks
  and loops. It is currently being discussed for `delegate`.
- `delegate`'s opcode should be decided. We exhausted all control-flow
  opcodes that begin with `0x0`.


proposals/Exceptions.md
```diff
@@ -143,42 +93,85 @@ Exception indices are used by:
...
+try blocktype
+  instruction*
+unwind
+  instruction*
+end
+```
+
+The `unwind` block is meant to contain cleanup instructions, such as
+destructors, in case any instruction in the corresponding try block throws.
+Currently the `unwind` instruction has the same semantics as the `catch_all`
```
ioannad: I'm surprised and confused with these semantics of `unwind`.

If `try ... unwind ... end` is semantically equivalent to `try ... catch_all ... rethrow end`, then I can't see why we are adding this instruction. Can't we just postpone this `try ... unwind ... end` to a future two-phase unwinding proposal? In fact, if the `unwind` semantics will change later, couldn't that create backwards compatibility problems? 

Until this document, I had understood `unwind` more as a primitive related to common-lisp's `unwind-protect`, which btw [seems to be related to]( https://stackoverflow.com/a/12999898 ) Java's and Python's `finally` instead of `catch_all`.

proposals/Exceptions.md
```diff
@@ -143,42 +93,85 @@ Exception indices are used by:
...
+```
+try blocktype
+  instruction*
+unwind
+  instruction*
+end
+```
+
+The `unwind` block is meant to contain cleanup instructions, such as
+destructors, in case any instruction in the corresponding try block throws.
```
ioannad: What are these "destructors" that the `unwind` block should contain? Is this a special category of Wasm instructions?

proposals/Exceptions.md
ioannad: Also it seems that `unwind` could spark a very long discussion, for example when common-lisp was being specified, `unwind-protect` related discussions spanned over years ([see this link](http://www.lispworks.com/documentation/HyperSpec/Issues/iss151_w.htm), at the bottom of the page). 

If we can avoid adding this instruction now, we could have these discussions in a different proposal.

RossTate: @ioannad It's good to point out that there are good reasons to be wary of adding control constructs. Fortunately, though, in the case of `unwind`, the control construct has been tried out in a wide variety of languages (with `finally` being a close relative) and its semantics are fairly well understood. It seems to have been found to be particularly useful in multi-language settings for reasons akin to why it was added here to make modules forwards compatible with two-phase exception handling. And nowadays its semantics are fairly well understood, at least at the level that WebAssembly operates at. (This wasn't the case back when Common Lisp was being specified, and some of the complications in the referenced discussion have to do with higher-level considerations.)

I cannot say the same for `catch_all` or `rethrow` though, so rather than removing `unwind` due to the above equivalence, my suggestion would be to remove `catch_all` instead (if you were to remove anything at all).

proposals/Exceptions.md
RossTate: Destructors here are from C++. They are C++'s analog of `finally`. However, their semantics slightly differ across C++ versions. In older versions, if a destructor throws an exception during "normal" clean up, then the exception is propagated, but if a destructor throws an exception during "excpetional" clean up, i.e. unwinding caused by an exception being thrown, then the program is supposed to terminate.

proposals/Exceptions.md
aheejin: @ioannad They have the same semantics, and they are planned to have different semantics in two-phase. In single-phase unwinding, both `catch_all`'s and `unwind`'s bodies run as we search up and unwind the stack. In two-phase, `catch` bodies run after we finish all two-phase unwinding process, while `unwind` bodies run as cleanups in the second phase, and their `end` instruction act as "resuming the second phase" instruction.

I think what you said also makes sense that if two are equivalent, why adding them now? I don't have a strong opinion either way myself. It was taken from @RossTate's suggestion, and I agreed with that for the reason that it'd be good if we make the current proposal and the follow-on future two-phase proposal as similar as possible, so that toolchains targeting EH proposals don't need to change where to put their cleanup code when they switch to the next proposal. (But there will be some other changes necessary for toolchains.)

@RossTate As I said in the other issue, `catch_all` has its uses. If we remove that, `unwind` cannot replace it when two-phase arrives. These days, I am almost getting an impression that you are trying to remove every single instruction that you don't need yourself for your future proposals. (By the way, I don't want this PR to be another discussion for removal of instructions.)

proposals/Exceptions.md
aheejin: @ioannad `unwind` blocks contain routines for cleanups that need to run when the stack is unwound. In C++ that's mostly destructors, and `finally` can belong here in compilers where normal-pass `finally` and exception-pass `finally` are compiled separately. But we don't have support for any languages that have `finally` construct yet. I'll change this term to "C++ destructors" to make the example clearer.

proposals/Exceptions.md
```diff
@@ -143,42 +93,85 @@ Exception indices are used by:
...
+Currently the `unwind` instruction has the same semantics as the `catch_all`
+instruction that it catches all exceptions. In case an exception is caught by
+the `unwind` block, it becomes the catching block.
+
+The `end` instruction at the end of `unwind` block is special that it
+automatically rethrows the current exception.
+
+The current plan is, try-unwind blocks will have a different semantics from
+that of try-catch blocks when we introduce two-phase unwinding as a follow-on
+proposal in the future.
```
rossberg: It would really help if this plan could be sketched in this explainer somehow (or some separate doc to point to). Because to be honest, I still don't understand it, nor how the current semantics of `unwind` could be changed/generalised later in a backwards-compatible manner.

proposals/Exceptions.md
rossberg: > though, in the case of `unwind`, the control construct has been tried out in a wide variety of languages

@RossTate, can you point us to any such language? Constructs like `finally` or `unwind-protect` seem much more common-place and semantically well-understood.

proposals/Exceptions.md
ioannad: @aheejin I suppose here is the right place to make [this comment](https://github.com/WebAssembly/exception-handling/issues/87#issuecomment-707683679). 

I am not sure I understand how one can embed C++ instructions (destructors) in a WebAssembly code block, or how we can specify a WebAssembly instruction as containing instructions from another language (C++).

tlively: I believe the point of mentioning C++ destructors is to describe the intended use case of this feature (and its semantics by rough analogy to that use case), not to actually involve C++ in the formal semantics. "Destructors" are not and should not become a Wasm concept in this proposal.

proposals/Exceptions.md
RossTate: What is called `unwind` here is called `fault` in the .NET CLI. Like `finally`, `fault`/`unwind` is executed whenever the stack (frame) is unwound (as opposed to just basic cleanup like deallocated). The only difference between `finally` and `fault` is that the former is also executed by local control, e.g. `br` and `return`. We need to make the distinction in this proposal because the semantics of C++ destructors is different for the two kinds of paths.

rossberg: AFAIK, `fault` is not used by C#. Do we have any information where it is used? Managed C++?

Wouldn't the distinction needed by C++ be aided better by a `finally` that passes in a Boolean to indicate whether it's regular or exceptional finalisation?

RossTate: From what I understand, you would have to make significant changes to LLVM to get that to happen, including changing how it represents CFGs and exception handling. My take is that `finally` is a surface-level construct, and that `unwind` is part of its lower-level counterpart, with code duplication or #124 providing ways to support the rest of its lower-level counterpart.

proposals/Exceptions.md
aheejin: @ioannad Destructors are just function calls. Like,
```cpp
void test() {
  Object o;
  ... do some work ...
}
```
In this code, the compiler will automatically generate calls like `call ~Object()` (the real function name will be some mangled name) at the end of this function. So they are just plain function calls that need to run as cleanup whenever we exit the scope. This destructor should run both when exit the scope normally and exceptionally, but many compilers, such as LLVM, compile those two path separately, because the two paths have some different things to handle.

Anyway, this was used as an example, but if people think this term is confusing, I can remove it.

By the way, I was maybe not doing a very good job on explaining this when we were having a meeting. Did you think `unwind` is the same as `finally` then?

proposals/Exceptions.md
rossberg: Well, the LLVM backend gets by without any use of finally/unwind right now, so, shrug? And of course it is trivial to express fault a.k.a. unwind with such a flagged finally if needs be.

I'd still be interested in knowing which .NET producers actually use fault.

RossTate: > Well, the LLVM backend gets by without any use of finally/unwind right now, so, shrug?

Every function call in LLVM has an argument specifying how to unwind it.

> I'd still be interested in knowing which .NET producers actually use fault.

It's how .NET addresses the problem above that C++ destructors have different semantics on the two kinds of exit paths.

aheejin: @rossberg 

About LLVM: It compiles destructors for the normal path and the exception path separately, because the exception path contains some different logic, such as, if another exception arises within there, it does not throw but calls `std::terminate` or something. This is C++, but it also compiles `finally` away for some languages that have `finally`, such as Objective-C.

In the CFG level, LLVM has an `invoke` instruction, which is a variant of `call` instruction. `invoke` takes two successors: a normal destination, where we should go when the call returns normally, and unwind destination, where we should unwind to when the call throws. CFG's control flow logic is very direct and simple; all are just go-to edges (or unwind edges) and they are directly encoded. Our wasm backend lowers this to `try`-`catch`. 

So my point is, if we introduce `finally` as a replacement of `unwind`, at least for LLVM it is not usable for us. And I am not 100% sure if I understood your fragmented stack argument, but I think `finally` has all the same problem anyway, no?

We can still use `catch` for cleanup code as we have so far, but for 2PEH we need to distinguish user catch handler from cleanup code. That doesn't have to be `unwind`, but what I'm saying is we need a way.

rossberg: Yes, I can see that it's not useful with the specific choices LLVM made. But other compilers could benefit. If there is pressure to add finally, then we shouldn't end up having both unwind and finally, thus the idea of a simple generalisation that expresses both.

> And I am not 100% sure if I understood your fragmented stack argument, but I think finally has all the same problem anyway, no?

That is unrelated. That problem problem occurs with the search phase, not the unwind phase. The latter runs on the top of the stack, since all other stack frames are already unwound at that point.

mf-RDP:
Aheejin,

I just want to ask a few questions:
- When, do you guess, can we expect to see this in LLVM/clang as a (partially) working, first implementation, can we expect Q1/2021 or later?
- For clang++: will the current required WASM<>ABI implementation (_Unwind_CallPersonality, __cxa_begin_catch etc.) stay the same? I really hope so, also at least for the catchpad data layout in data segments.

Many Thanks
Sascha

aheejin:
@mf-RDP 
> * When, do you guess, can we expect to see this in LLVM/clang as a (partially) working, first implementation, can we expect Q1/2021 or later?

Can't guarantee an exact timeline, but we hope we can get it work with LLVM/Clang/Emscripten at least within Q1 2021.

> * For clang++: will the current required WASM<>ABI implementation (_Unwind_CallPersonality, __cxa_begin_catch etc.) stay the same? I really hope so, also at least for the catchpad data layout in data segments.

I think they will stay mostly the same. At least so far I don't plat to change them.

aheejin:
It's been a long discussion, so let's merge this finally.

The following are the updated changes after the initial version. I listed relevant discussions within ().
- Removed dependences on the reference types proposal and the multivalue proposal. The multivalue proposal was previously listed as dependent because 1. `try` is basically a `block`, so it can have multivalue input/output 2. `br_on_exn` can extract multiple values from a `block`. We don't have `br_on_exn` anymore, and I'm not sure 1 is a strong enough reason to make it a dependence.
- Mention `rethrow` cannot occur within `unwind` (#142)
- Mention some runtimes, especially web VMs, can attach stack traces to the exception object, implying stack traces are not required for all VMs
- Update label/validation rules for `delegate` and `rethrow` (#146)
- Finalize opcodes for `delegate` (0x18) and `catch_all` (0x19) (#145 and #147)

One thing that does not seem to be fully resolved is the typing rule for `try`-`unwind`. I previously thought it would have the same rule as `try`-`catch`, but @rossberg [suggested](https://github.com/WebAssembly/exception-handling/pull/143#discussion_r574473077) the `unwind` part should have the type []->[] regardless of `try` part. @RossTate What's your take on this, as who originally proposed the instruction? (But I think we can elaborate more on this part even after this PR is merged and this is not necessarily a blocker for merging this.)

---

The repo does not allow me to add some people as reviewers. I guess it only lists people who are in the WebAssembly org in the reviewer list? I think several people who already commented on this PR will be notified. CC'ing other people (who are not on the possible reviewer list) working on this: @thibaudmichaud @takikawa

proposals/exception-handling/Exceptions.md
```diff
@@ -499,10 +537,17 @@ throws, and rethrows as follows:
...
-| `br_on_exn` | `0x0a` |  relative_depth : `varuint32`, index : `varuint32` | Branches to the given label and extracts data within `exnref` on top of stack if it was created using the corresponding exception `index` |
+| `rethrow` | `0x09` | relative_depth : `varuint32` | Pops the `exnref` on top of the stack and throws it |
 
 The *sig* fields of `block`, `if`, and `try` operators are block signatures
 which describe their use of the operand stack.
+
+Note that the textual `catch_all` instruction is implemented using the
+`else` operator. Since the `else` operator is always unambiguous in the binary
+format, there is no need to tie up a separate opcode for the `catch_all`
+instruction.
```
takikawa: I think this text about `catch_all` can be deleted now that it's assigned opcode `0x19`.

takikawa:
I read through the new explainer text and it looks good to me, I agree with merging it and sorting out remaining issues such as `try-unwind` typing after.

proposals/exception-handling/Exceptions.md
```diff
@@ -143,42 +93,80 @@ Exception indices are used by:
...
   instruction*
 end
 ```
 
-A try block ends with a `catch block` that is defined by the list of
-instructions after the `catch` instruction.
+A try-catch block contains zero or more `catch` blocks and zero or one
+`catch_all` block. All `catch` blocks should precede the `catch_all` block, if
+any. The `catch`/`catch_all` instructions (within the try construct) are called
+the _catching_ instructions. There should be at least `catch` or `catch_all`
```
tlively:
```suggestion
the _catching_ instructions. There should be at least one `catch` or `catch_all`
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
+Then in case of a try-catch block, tagged catch blocks are tried in the
+order they appear in the catching try block, until one matches. If a matched
+tagged catch block is found, control is transferred to the body of the catch
+block, and the data fields of the exception are pushed back onto the stack.
+
+Otherwise, control is transferred to the body of the `catch_all` block, if any.
+However, unlike tagged catch blocks, the constructor arguments are not copied
+back onto the operand stack. If no tagged catch blocks were matched, and the
+catching try block doesn't have a `catch_all` block, the exception is rethrown
+to the next enclosing try block.
```
tlively: There might not be an enclosing try block, but that isn't a problem.

```suggestion
catching try block doesn't have a `catch_all` block, the exception is rethrown.
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
-the caught exception.
+In case of a try-unwind block, the control is transfered to the body of the
+`unwind` block. After executing instructions within the `unwind` block, the
+exception is automatically rethrown when the control reaches the `end`
+instruction. As in the case of the `catch_all` block, the exception arguments
+are not copied onto the operand stack.
+
+If the selected catch/unwind block does not throw an exception, it must yield
+the value(s) expected by the corresponding catching try block. This includes
+popping the caught exception.
```
tlively: It is also possible for a catch block to do nothing with the exception and have the correct type if the type of the exception matches the type of the try.

```suggestion
If the selected catch/unwind block does not throw an exception, it must yield
the value(s) expected by the corresponding catching try block.
```

tlively (approved):
I have just a few comments, but otherwise LGTM!

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
 
 ### Rethrowing an exception
 
-The `rethrow` instruction takes the exception associated with the `exnref` on
-top of the stack, and rethrows the exception. A rethrow has the same effect as a
-throw, other than an exception is not created. Rather, the referenced exception
-on top of the stack is popped and then thrown. The `rethrow` instruction traps
-if the value on the top of the stack is null.
+The `rethrow` instruction can only appear in the body of a catch/catch_all block
+but not within the body of an unwind block. It always re-throws the exception
```
thibaudmichaud: I think saying that `rethrow` cannot appear in an `unwind` block is too restrictive, consider the following example:
```wasm
try
...
catch
  try
  ...
  unwind
    rethrow 1
  end
end
```
The `rethrow` appears in the `unwind` block, but IIUC this is ok because we never rethrow its exception. Should the rule rather be something like "the `rethrow` immediate can only be a `catch` or `catch_all` block"?

ioannad: I still don't understand why we are making this restriction. IIUC, for `N > 0` the following code

```
try bt
  instr1*
unwind
  instr2*
  rethrow N
  instr3*
end
```

is equivalent to

```
try bt
  instr1*
unwind
  instr2*
  br 0
  instr3*
end
rethrow N-1
```

For N=0, 

```
try bt
  instr1*
unwind
  instr2*
  rethrow 0
  instr3*
end
```

is equivalent to 


```
try bt
  instr1*
unwind
  instr2*
end
```

In both cases, `instr3*` is redundant. 

So I don't understand what situation is prevented with this restriction. Am I (still) misunderstanding the semantics of `unwind`?

(cc. @RossTate)

ioannad: Clarification: I do agree with @thibaudmichaud that, if we do add a restriction, then it should be on the rethrow target (i.e., that rethrow cannot target an unwind block).

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
-the caught exception.
+In case of a try-unwind block, the control is transfered to the body of the
+`unwind` block. After executing instructions within the `unwind` block, the
+exception is automatically rethrown when the control reaches the `end`
+instruction. As in the case of the `catch_all` block, the exception arguments
+are not copied onto the operand stack.
+
+If the selected catch/unwind block does not throw an exception, it must yield
+the value(s) expected by the corresponding catching try block. This includes
+popping the caught exception.
```
rossberg, line 236: Maybe it's worth mentioning here that catch and unwind blocks can also contain branches that abort them. In the particular case of unwind, this implies that the exception is not rethrown.

Unwind actually creates an ambiguity here. What happens for:
```
try
  ...
unwind
  br 0
end
```
Is this interpreted as exiting only the unwind block (and thus rethrow) or the try block altogether (thus aborting the rethrow)? This problem exists for no other construct, so it is indicative of the special unwind-end semantics being rather alien to Wasm's control structure (i.e., too high-level).

proposals/exception-handling/Exceptions.md
rossberg: Agreed with @thibaudmichaud and @ioannad. The restriction is stated the wrong way. It should simply say that the rethrow target must be a catch block, which automatically implies that it has to be properly nested into one.

I also agree that there is no particular reason to disallow rethrow targetting unwind.

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
-br_on_exn label except_index
+try $l1
+  ...
+catch 1
+  ...
+  block
+    ...
+    try $l2
+      ...
+    catch 2
```
```suggestion
    catch  ;; $l2
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
+      ...
+      try $l3
+        ...
+      catch 3
```
```suggestion
      catch  ;; $l3
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
 
-The `br_on_exn` instruction is a conditional branch that checks the exception
-tag of an exception on top of the stack, in the form of:
+For example consider the following:
 
 ```
-br_on_exn label except_index
+try $l1
+  ...
+catch 1
```
```suggestion
catch  ;; $l1
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
+    ...
+    try $l2
+      ...
+    catch 2
+      ...
+      try $l3
+        ...
+      catch 3
+        ...
+        rethrow N (or label)
```
```suggestion
        rethrow N  ;; (or label)
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
+In this example, `N` is used to disambiguate which caught exception is being
+rethrown. It could rethrow any of the three caught expceptions. Hence,
+`rethrow 0` corresponds to the exception caught by `catch 3`, `rethrow 1`
+corresponds to the exception caught by `catch 2`, and `rethrow 3` corresponds
+to the exception caught by `catch 1`. In wat format, the argument for the
+`rethrow` instructions can also be written as a label, like branches. So
+`rethrow 0` in the example above can also be written as `rethrow $l3`.
+
+Note that `rethrow 2` is not allowed because it does not reference a `try`
+instruction. Rather, it references a `block` instruction, so it will result in a
```
```suggestion
Note that `rethrow 2` is not allowed because it does not reference a catch
block. Rather, it references a `block` instruction, so it will result in a
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
+validation failure.
+
+Note that the example below is a validation failure:
+```
+try $l1
+  try $l2
+    rethrow $l2 (= rethrow 0)
```
```suggestion
    rethrow $l2  ;; (= rethrow 0)
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
+catch
+end
+```
+The `rethrow` here references `try $l2`, but the `rethrow` is not within its
+`catch` block.
+
+### Try-delegate blocks
+
+Try blocks can also be used with the `delegate` instruction. A try-delegate
+block contains an `delegate` instruction with the following form:
```
```suggestion
block contains a `delegate` instruction with the following form:
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
-to the label referenced by the instruction (In the binary form, the label will
-be converted to a relative depth immediate, like other branch instructions), and
-while doing that, pops the `exnref` value from the stack and instead pushes the
-exception's argument values on top of the stack. In order to use these popped
-values, the block signature of the branch target has to match the exception
-types - because it receives the exception arguments as branch operands. If the
-exception tag does not match, the `exnref` value remains on the stack. For
-example, when an `exnref` contains an exception of type (i32 i64), the target
-block signature should be (i32 i64) as well, as in the following example:
+The `delegate` instruction does not have an associated body, so try-delegate
```
```suggestion
The `delegate` clause does not have an associated body, so try-delegate
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
-      br $end
-    ...
-  end $l1
-  ;; handler for $e1
-end $end
+try $l1
+catch 1
+  try
+    call $foo
+  delegate $l1 (= delegate 0)
```
```suggestion
  delegate $l1  ;; (= delegate 0)
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
+takes a label defined by a construct in which they are enclosed, and delegates
+exception handling to a catch/unwind block specified by the label. For example,
+consider this code:
 
 ```
-block $l (result i32 i64)
+try $l1
+  try
+    call $foo
+  delegate $l1 (= delegate 0)
```
```suggestion
  delegate $l1  ;; (= delegate 0)
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
+catch_all
   ...
 end
 ```
 
-This can now be used to construct handler switches in the same way `br_table`
-is used to construct regular switch:
+If `call $foo` throws, searching for a catching block first finds `delegate`,
+and because it delegates exception handling to catching instructions associated
+with `$label0`, it will be next checked by the outer `catch` and then
```
```suggestion
with `$l1`, it will be next checked by the outer `catch` and then
```

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
+`catch_all` instructions.
+When the specified label within a `delegate`
+instruction does not correspond to a `try` instruction, it is a validation
+failure.
```
rossberg: Technically, this is an unnecessary restriction as well. We could specify it as simply being equivalent to a rethrow inside the target block, which makes it agnostic to what sort of construct it is. That also increases flexibility, because it allows to delegate to an enclosing handler that's not in lexical scope, which is more in line with the dynamically scoped nature of exception handlers.

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
-### Stack traces
+#### Stack traces
 
 When an exception is thrown, the runtime will pop the stack across function
-calls until a corresponding, enclosing try block is found. It may also associate
-a stack trace that can be used to report uncaught exceptions. However, the
-details of this is left to the embedder.
+calls until a corresponding, enclosing try block is found. Some runtimes,
+especially web VMs may also associate a stack trace that can be used to report
+uncaught exceptions. However, the details of this is left to the embedder.
```
```suggestion
uncaught exceptions. However, the details of this are left to the embedder.
```

proposals/exception-handling/Exceptions.md
```diff
@@ -326,20 +381,20 @@ document](https://github.com/WebAssembly/spec/blob/master/document/core/text/ins
 The following rules are added to *instructions*:
 
 ```
-  try blocktype instruction* catch instruction* end |
+  try blocktype instruction* (catch instruction*)+ (catch_all instruction*)+ end |
```
rossberg: This grammar would require at least one catch and one catch_all, and allows multiple of the latter. I think you want
```suggestion
  try blocktype instruction* (catch instruction*)* (catch_all instruction*)? end |
```
This allows a try with no catch blocks, which I think is preferable anyway. If you want to rule that out, you'll need to complicate the grammar.

proposals/exception-handling/Exceptions.md
aheejin: Good catch, thanks. Removed.

proposals/exception-handling/Exceptions.md
aheejin: @tlively Done.
@rossberg What would that mean then? Wouldn't that be just the same as a `block`?

proposals/exception-handling/Exceptions.md
aheejin: I agree that text is confusing. Removed.

proposals/exception-handling/Exceptions.md
aheejin: Done.

proposals/exception-handling/Exceptions.md
```diff
@@ -143,42 +93,80 @@ Exception indices are used by:
...
+catch_all
   instruction*
 end
 ```
 
-A try block ends with a `catch block` that is defined by the list of
-instructions after the `catch` instruction.
+A try-catch block contains zero or more `catch` blocks and zero or one
+`catch_all` block. All `catch` blocks should precede the `catch_all` block, if
+any. The `catch`/`catch_all` instructions (within the try construct) are called
```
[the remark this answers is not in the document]
aheejin: Then do you think `unwind`, `catch_all` and `end` all should be a clause too?

proposals/exception-handling/Exceptions.md
aheejin: We discussed this at length in some of comments in #143 and #146, and it seems people there, including me, agreed that validation failure is clearer. I also would like to strongly advocate for this, because assuming exception can slip until it meets any enclosing try will make toolchain optimization, such as in Binaryen, very difficult.

proposals/exception-handling/Exceptions.md
aheejin: I think try-catch is a little different case then just having a list of of something (e.g. functions) that can be empty. It was designed as a two-parts instruction from the beginning, and while I don't think it's a groundbreaking change, many parts of toolchain have to treat try without catch as a special case, because most parts are written under the assumption there will be two parts.

Also I think this is a kind of big change that has not been discussed so far, so I think it warrants a separate discussion thread..?

rossberg: There are still two parts, but one is a list, and the list can have any length, including length 0. :) I don't think this is different from any other similar case, where we have deliberately refrained from making length 0 a special case -- there's not a single occurrence of `X^+` in the entire Wasm grammar. :)

proposals/exception-handling/Exceptions.md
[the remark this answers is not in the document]
rossberg: Catch, unwind delegate. I'd call end an opcode. But this admittedly is nitpicking, so feel free to ignore.

proposals/exception-handling/Exceptions.md
```diff
@@ -143,42 +93,80 @@ Exception indices are used by:
...
+the _catching_ instructions. There should be at least `catch` or `catch_all`
+block within a try-catch block.
+
+The _body_ of the try block is the list of instructions before the first
+catching instruction. The _body_ of each catch block is the sequence of
+instructions following the corresponding catching instruction before the next
+catching instruction (or the `end` instruction if it is the last catching
+block).
+
+The `catch` instruction has an exception tag associated with it. The tag
```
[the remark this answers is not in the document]
rossberg: Yes, discussing elsewhere is fine by me.

proposals/exception-handling/Exceptions.md
rossberg: Forbidding rethrow would make sense for finally, because that has nothing to rethrow if the block was entered regularly, but no real point for unwind (unless one assumes we will be adding yet some other form of abort that is neither a trap nor an exception, but that's hardly desirable). But it's conservative to disallow it, so okay to leave it out for now.

proposals/exception-handling/Exceptions.md
```diff
@@ -205,99 +193,166 @@ flushed, the embedder defines how to handle uncaught exceptions. Otherwise, the
...
+catching try block doesn't have a `catch_all` block, the exception is rethrown
+to the next enclosing try block.
 
 If control is transferred to the body of a catch block, and the last instruction
 in the body is executed, control then exits the try block.
 
-If the selected catch block does not throw an exception, it must yield the
-value(s) expected by the corresponding catching try block. This includes popping
-the caught exception.
+In case of a try-unwind block, the control is transfered to the body of the
```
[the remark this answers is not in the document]
rossberg: Sounds good!

proposals/exception-handling/Exceptions.md
aheejin: Yeah, maybe that doesn't need to be treated as a special case. But anyway, I think we should make this a separate issue post first for more visibility than just changing this here and merging it in.

proposals/Exceptions.md
ioannad: It seems I have forgotten to answer this comment, apologies @aheejin !

> Anyway, this was used as an example, but if people think this term is confusing, I can remove it.

To me it *was* confusing because I am thinking of the Wasm spec as self-standing, and I don't really think about translations from C++, as I have not worked on any tooling, just on the SpiderMonkey implementation. This might be just me though, so I wouldn't mind either way now that I understand the intention behind it. You pushed a commit which removes this anyway, which I think is fine.

> By the way, I was maybe not doing a very good job on explaining this when we were having a meeting.

I don't think we talked about unwind much, so apologies for not having asked more questions there.

> Did you think unwind is the same as finally then?

In a way I did; I thought unwind is the same as finally when an exception is thrown, this is why I [initially wrote a formal spec overview](https://github.com/WebAssembly/exception-handling/issues/87#issuecomment-702323595) with that behaviour for unwind. When I understood that this is equivalent to `catch_all ... rethrow 0 end` (modulo the actual behaviour of unwind's `rethrow` in a future 2-phase-unwinding proposal), I simplified the spec to the one in #143 by "hardcoding" this equivalence for now.

proposals/exception-handling/Exceptions.md
ioannad: @rossberg IIUC also having `br`, `br_if`, `br_table`, or `return` among the instructions of an `unwind` clause would abort the unwinding process.

proposals/exception-handling/Exceptions.md
ioannad: I just want to add that issue #131 by @tlively also asks whether a try without further clauses could be allowed or not.

proposals/exception-handling/Exceptions.md
RossTate: A `rethrow` that "targets" `unwind` is problematic because `unwind` can fire due to things besides exceptions. But a `rethrow` that is simply a `throw` of a caught exception should be allowed within `unwind`, just like `throw` is. (For supporting Common Lisp, it is necessary that exceptions be throwable from within `unwind` clauses.)

rossberg: @ioannad, but branches do not invoke intermediate unwind clauses, IIUC.

aheejin: I clarified the rethrow-unwind relationship to match the rule we agreed on here: https://github.com/WebAssembly/exception-handling/pull/137/commits/4d6c635ce79413fd9f48abe87050d1926575a986

aheejin:
Thanks for all the comments. I think I addressed most of them. One suggestion was whether we should allow catchless trys, but I think that discussion can continue in a separate issue and can be merged to the explainer later if we decide to make changes.

I'll merge this shortly if no one has any more comments.

dschuff (approved):

ioannad (approved):

rossberg (approved):

PR #151, opened 2021-04-21
[interpreter] Add event section

thibaudmichaud:
This change adds support for the event section in the interpreter. This is based on the explainer and on existing implementations (V8, wabt). The changes are mostly mechanical and follow the syntax and logic of function declarations, with the additional rule that the result type must be empty.


interpreter/text/arrange.ml
```diff
@@ -359,12 +359,20 @@ let data i seg =
   Node ("data $" ^ nat i, segment_mode "memory" dmode @ break_bytes dinit)
 
 
+(* Events *)
+
+let event off i e =
+  Node ("event",
```
rossberg: This could insert a symbolic name for better readability, analogous to the other forms of definition.

interpreter/text/parser.mly
```diff
@@ -842,6 +845,61 @@ memory_fields :
...
+    { let FuncType (ins, out) = $5 in FuncType ($3 @ ins, out) }
+  | LPAR PARAM bind_var value_type RPAR event_fields_import  /* Sugar */
+    { let FuncType (ins, out) = $6 in FuncType ($4 :: ins, out) }
+
+event_fields_import_result :  /* Sugar */
+  | /* empty */ { FuncType ([], []) }
+  | LPAR RESULT value_type_list RPAR event_fields_import_result
+    { let FuncType (ins, out) = $5 in FuncType (ins, $3 @ out) }
+
+event_types :
```
rossberg: Nit: `event_type`, since it represents a single event type. Also, I think you can define this as simply
```
event_type :
  | func_type
    { fun c -> EventType ($1 c) }
```
(assuming you added the `EventType` constructor as suggested above).

interpreter/binary/decode.ml
```diff
@@ -555,6 +556,7 @@ let import_desc s =
   | 0x01 -> TableImport (table_type s)
   | 0x02 -> MemoryImport (memory_type s)
   | 0x03 -> GlobalImport (global_type s)
+  | 0x04 -> ignore (vu32 s); EventImport (at var s)
```
rossberg: You need to check that the value is 0, otherwise it won't be forward-compatible. Also, a single zero byte is enough for an unused extension hook (since it's upward-compatible to LEB), and is what we did elsewhere.

In similar other places, the decoder uses the `zero` function for this.

Moreover, I think you want to introduce an `event_type` decoding function for this, since the same it's reused for event definitions below. Something like:
```
let event_type s =
  zero s;
  let x = at var s in
  EventType x
```

interpreter/binary/decode.ml
```diff
@@ -592,6 +594,13 @@ let memory s =
 let memory_section s =
   section `MemorySection (vec (at memory)) [] s
 
+(* Event section *)
+
+let event s =
+  ignore (vu32 s); var s
```
rossberg: As mentioned above, this should probably invoke an `event_type` decoder function.

interpreter/binary/encode.ml
```diff
@@ -426,6 +426,7 @@ let encode m =
       | TableImport t -> u8 0x01; table_type t
       | MemoryImport t -> u8 0x02; memory_type t
       | GlobalImport t -> u8 0x03; global_type t
+      | EventImport x -> u8 0x04; vu32 0x00l; var x
```
rossberg: Similar to the decoder, I'd introduce an `event_type` encoding function.

rossberg (commented):
Ah, this is great. It looks good overall, up to some details of the factorisation, see comments.

Sorry, I didn't know that you were up to this, otherwise I would have pointed you to the PR at effect-handlers/wasm-spec#4, where I already implemented event sections (though with some extensions for the continuation proposal). That PR also includes basic parts of the exception semantics, so feel free to steal form it to avoid duplicate work. It covers only a subset of the exception proposal's instruction set, though.

(One difference with your PR is that I did not use the indirection through I type index in the representation of event types. That was just laziness on my part, and is a TODO in the repo. Your encoding is more accurate.)

interpreter/runtime/instance.ml
```diff
@@ -15,6 +16,7 @@ type module_inst =
 and func_inst = module_inst ref Func.t
 and table_inst = Table.t
 and memory_inst = Memory.t
+and event_inst = func_type
```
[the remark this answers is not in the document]
thibaudmichaud: Done

interpreter/syntax/types.ml
```diff
@@ -10,11 +10,13 @@ type 'a limits = {min : 'a; max : 'a option}
 type mutability = Immutable | Mutable
 type table_type = TableType of Int32.t limits * ref_type
 type memory_type = MemoryType of Int32.t limits
+type event_type = func_type
```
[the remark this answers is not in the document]
thibaudmichaud: Done

interpreter/text/arrange.ml
thibaudmichaud: Done

interpreter/text/parser.mly
thibaudmichaud: Good point, I simplified this even further by removing the new rule completely and using func_type directly.

interpreter/binary/decode.ml
thibaudmichaud: Done

interpreter/binary/decode.ml
thibaudmichaud: Done

interpreter/binary/encode.ml
thibaudmichaud: Done

thibaudmichaud:
Thanks for the review! I was also not aware you were working on this, I will definitely take some inspiration.
I only plan to implement what is necessary to convert EH spec tests into JS tests for now, not the full runtime semantics  yet. I can revisit the priority of that when EH is at phase 3 if no one else takes it.

interpreter/syntax/types.ml
```diff
@@ -10,11 +10,13 @@ type 'a limits = {min : 'a; max : 'a option}
 type mutability = Immutable | Mutable
 type table_type = TableType of Int32.t limits * ref_type
 type memory_type = MemoryType of Int32.t limits
+type event_type = EventType of int32
```
rossberg, line 13: For consistency, can you change the other occurrences to `int32` as well?

interpreter/binary/encode.ml
```diff
@@ -118,6 +121,8 @@ let encode m =
     let memory_type = function
       | MemoryType lim -> limits vu32 lim
 
+    let event_type x = vu32 0x00l; var x
```
rossberg, line 124:
```suggestion
    let event_type x = u8 0x00; var x
```

interpreter/binary/encode.ml
```diff
@@ -426,6 +428,7 @@ let encode m =
       | TableImport t -> u8 0x01; table_type t
       | MemoryImport t -> u8 0x02; memory_type t
       | GlobalImport t -> u8 0x03; global_type t
+      | EventImport x -> u8 0x04; event_type x
```
rossberg:
```suggestion
      | EventImport t -> u8 0x04; event_type t
```

interpreter/runtime/event.ml
```diff
@@ -0,0 +1,10 @@
+open Types
```
rossberg, line 1: Even if its kind of trivial, can you also add an .mli file for this module?

interpreter/runtime/event.ml
```diff
@@ -0,0 +1,10 @@
...
+
+type event = {ty : func_type}
```
rossberg, line 3: Hm, it's a bit odd that this is func_type not event_type. I can see why, but it would break down once an event type has additional attributes.

The underlying problem is that there are two notions of event_type: the syntactic one (with a typeidx) and the semantic one (with the function type inline). Depending on context, we want one or the other. This is not a situation we've had before, so I don't have any great suggestion right now. We can resolve this later.

interpreter/syntax/ast.ml
```diff
@@ -150,6 +150,8 @@ and memory' =
   mtype : memory_type;
 }
 
+type event = int32 Source.phrase
```
rossberg:
```suggestion
type event = event' Source.phrase
and event' =
{
  etype : var;
}
```

test/core/exceptions.wast
```diff
@@ -0,0 +1,11 @@
+;; Test exception handling support
+
+(module
+  (event)
+  (event (param i32))
```
rossberg: Testing a few more syntactic variations (e.g. with symbolic names) couldn't harm.

test/core/exceptions.wast
```diff
@@ -0,0 +1,11 @@
+;; Test exception handling support
```
rossberg: Nit: Maybe name this file `event.wast`, since it's testing event definitions (exception-related instructions should go to separate files).

rossberg (approved):
Sorry for the delay. This LGTM modulo a few nits.

There is an open question regarding the proper definition and use of `event_type` (see comment below), but I think we can resolve that later.

interpreter/syntax/types.ml
thibaudmichaud: Done

interpreter/binary/encode.ml
thibaudmichaud: Done

interpreter/binary/encode.ml
thibaudmichaud: Done

interpreter/runtime/event.ml
thibaudmichaud: Done

interpreter/runtime/event.ml
thibaudmichaud: Ack. Regarding the syntactic vs semantic `event_type`: I don't see how this is too different from functions, where the type can also be represented directly or through the type ID? It makes sense to me that if one uses the "semantic" encoding, the other does too.

interpreter/syntax/ast.ml
thibaudmichaud: Done

test/core/exceptions.wast
thibaudmichaud: Done. I did much more extensive testing with my next PR which includes the new EH opcodes. I'll send it for review after this one is merged.
In the meantime I added the relevant tests to this PR as suggested.

test/core/exceptions.wast
thibaudmichaud: Done

interpreter/runtime/event.ml
rossberg: The difference is the extra indirection. For function types, the type section is just mapping a typeidx to a func type, and the functype itself is a closed type description. Once looked up, we can use that directly. But for event types, the definition itself refers to another typeidx, so isn't self-contained. This becomes clearer if we assume that an event type has additional ingredients, say, a hypothetical event_mode (as per the reserved zero byte). Then for the syntax, we'd need 
```event_type = EventType of typeidx * event_mode```
But but when e.g. defining extern_type, that index is meaningless (because it's used outside any module) and you want `ExternEventType of event_type'` where
```event_type' = EventType' of func_type * event_mode```
FWIW, a similar problem also arises with typed reference types, so the syntactic vs semantic distinction becomes rather ubiquitous.

**Review by rossberg — APPROVED** [review]

File: interpreter/runtime/event.ml (hunk 12 above)
Comment (thibaudmichaud) [reply interpreter/runtime/event.ml:3]: I agree, and for this reason I think types.ml should contain the resolved type.
We already have the typeidx in the AST:
```
type event = event' Source.phrase
and event' =
{
  etype : var;
}
```
which makes sense for encoding, decoding and pretty-printing. For type checking, imports/exports and everything else, the resolved type is easier to work with. WDYT about simply reverting back to `type event_type = EventType of func_type`?

rossberg: Actually, both `etype` and `EventImport` ought to carry an `event_type`, not a `var`. Otherwise, they in turn don't factorise correctly, e.g., when event types were to consist of more than just a var.

I'm afraid we can't hack around the two facets, they're inherent. It's relatively localised here, but becomes ubiquitous with concrete reference types. In the function references proposal I currently resolve it by actually distinguishing syntactic and semantic types. They differ by how variables are represented: for syntactic types it's just an index, with semantic types they become a pointer to the actual definition.

But I suggest to not worry about it for now. I want to play with streamlining this a bit and can fix it later. This needs a representation that cleanly carries over to the paper spec.

Issue #150, opened 2021-03-31
Details of the JS API for exception handling


takikawa:
Now that the semantics for the current iteration of the proposal are fairly solid and two browser implementations are underway, I was wondering if this is a good time to figure out the details for the JS API for exceptions.

@Ms2ger has written up [draft JS API spec text](https://github.com/WebAssembly/exception-handling/pull/86) which provides a good baseline to expand on. This draft was made during the previous iteration of the semantics, so it needs an update to remove `exnref`. But it already describes `WebAssembly.Exception` and `WebAssembly.RuntimeException` that are discussed below and additional API features can be incorporated into it.

Here are some of the aspects that the final JS API could/should likely provide, along with some design questions on how these aspects should work:

  * Provide an interface for exported exception types such as `WebAssembly.Exception`.
    - Previous discussions have generally assumed this should exist, and be called `WA.Exception`.
    - Should it support the [type reflection API](https://github.com/WebAssembly/js-types) proposal? i.e., a `type()` method on exceptions.
    - Should it be possible to construct these in JS (given a function type or just a list of parameters from the type reflection API), to create a unique exception tag on the JS side? (which a Wasm module could import)
  * Provide an interface for Wasm exceptions caught in JS such as `WebAssembly.RuntimeException`.
    - A previous discussion reached [a consensus](https://github.com/WebAssembly/exception-handling/issues/109#issuecomment-625411897) to call this `WebAssembly.RuntimeException` and to keep it separate from `WebAssembly.RuntimeError`.
    - Should it allow access to exception values? (e.g., `values()` or [`getArg()`](https://github.com/WebAssembly/exception-handling/issues/109#issuecomment-639447310) method, using `ToJSValue` to coerce Wasm values to JS as usual)
    - If values are accessible, it should require a capability (i.e., the exception type export) to access. See [this comment](https://github.com/WebAssembly/exception-handling/issues/109#issuecomment-639447310) for why.
    - As discussed in the link in the previous bullet, should it have a method to let you check the tag against an exception type?
    - Should JS be allowed to construct these? (e.g., given an exception type, and values that would be coerced via `ToWasmValue`)

In terms of features, IMO it would make sense to allow accessing the values (requiring the exception type for access) and to allow constructing exception objects. Not adding this functionality won't actually prevent JS hosts from doing these operations (both limitations can be hacked around by constructing a new Wasm module) but adding it will make it more convenient.

[...]

Issue #109, opened 2020-04-23
[js-api] Should rethrow/br_on_exn trap on undefined?


aheejin:
We decided to make `rethrow` and `br_on_exn` trap on a null value in #90. Should we do the same when a given `exnref` is an undefined value thrown from JavaScript?

Raised by @backes in https://bugs.chromium.org/p/v8/issues/detail?id=10128#c12.

RossTate:
I think wasm programs should be able to expect that they can `rethrow` any `exnref` that was caught by a `try-catch`.

RossTate:
Oh shoot, apparently that's already not the case. From what I can tell of the JS API in #86, which lines up with the linked Chromium issue, if JavaScript throws `null` then the `exnref` caught by a surrounding `try-catch` in wasm will be `ref.null`.

aheejin:
This is not about null but about undefined I think? And #86 does not seem to say anything about neither null nor undefined case. (And #86 was written before we decided on #93, so it needs revision too, but it's not relevant with this issue)

RossTate:
I was trying to figure out what the interaction of wasm exception handling and JS throws was in more detail so that I could understand the problem more, and #86 was the most recent description I could find. In particular, it says that a JS value being thrown is converted to a wasm value via ToWebAssemblyValue. I then noticed that ToWebAssemblyValue converts a JS null to `ref.null`, hence the follow-up comment above. (It looks like #93 doesn't change that observation.)

So I don't know what to do about my first comment. It seems like a useful principle to always be able to assume you can rethrow an exception you caught. But I didn't realize at the time that the principle is incompatible with the existing design, so I'm at a loss for suggestions. My original thinking was that the semantics of `rethrow` and `br_on_exn` should be independent of what JS value was thrown, but it already depends on whether the JS value was null or not, so I'm not sure what general principle to apply. Sorry for my thinking out loud.

RossTate:
Hmm, I realized that the JS-interop strategy here has the ability to turn any wasm reference into an `exnref`. First, give the wasm reference to JS code so that it goes through `ToJSValue`, then throw the resulting reference in JS code. When this exception enters a wasm stack frame, it will go through `ToWebAssemblyValue`, which is nearly the inverse of `ToJSValue`. Then wasm code can catch that value as an `exnref`.

[...]

PR #154, opened 2021-05-25
[js-api] Add description of interface additions to overview document

takikawa:
Based on the discussion in #150, this PR adds a description of new proposed JS API interfaces for exceptions to the overview document.

This is a draft and I'm open to any suggestions. I commented in #150 about what other points we might adjust or discuss.

The overview is also based on @Ms2ger's formal spec PR #86.

### DIFF
--- a/proposals/exception-handling/Exceptions.md
+++ b/proposals/exception-handling/Exceptions.md
@@ -395,6 +395,36 @@ specially mark non-catchable exceptions.
 be intercepted in JS, and types of exceptions generated from stack overflow and
 out of memory are implementation-defined.)
 
+#### API additions
+
+The following additional classes are added to the JS API in order to allow JavaScript to interact with WebAssembly exceptions:
+
+  * `WebAssembly.Exception`
+  * `WebAssembly.RuntimeException`.
+
+The `WebAssembly.Exception` class represents an exception defined in the event section and exported from a WebAssembly module. It allows querying the type of an exception following the [JS type reflection proposal](https://github.com/WebAssembly/js-types/blob/master/proposals/js-types/Overview.md). Constructing an instance of `Exception` creates a fresh exception tag, and the new exception can be passed to a WebAssembly module as an import.
+
+The `WebAssembly.RuntimeException` class represents an exception thrown from WebAssembly, or an exception that is constructed in JavaScript and is to be thrown to a WebAssembly exception handler. The `RuntimeException` constructor accepts an `Exception` argument and a sequence of arguments for the exception's data fields. The `Exception` argument determines the exception tag to use. The data field arguments must match the types specified by the `Exception`'s type. The `is` method can be used to query if the `RuntimeException` matches a given `Exception`'s exception tag. The `getArg` method allows access to the data fields of a `RuntimeException` if an `Exception` is passed with a matching exception tag. This last check ensures that without access to a WebAssembly module's exported exception, the associated data fields cannot be read.
+
+More formally, the added interfaces look like the following:
+
+```
+[LegacyNamespace=WebAssembly, Exposed=(Window,Worker,Worklet)]
+interface Exception {
+  constructor(ExceptionType type);
+  ExceptionType type();
+};
+
+[LegacyNamespace=WebAssembly, Exposed=(Window,Worker,Worklet)]
+interface RuntimeException {
+  constructor(Exception exceptionType, sequence<any> payload);
+  any getArg(Exception exceptionType, unsigned long index);
+  boolean is(Exception exceptionType);
+};
+```
+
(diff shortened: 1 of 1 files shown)


proposals/exception-handling/Exceptions.md
```diff
@@ -395,6 +395,36 @@ specially mark non-catchable exceptions.
...
+
+The `WebAssembly.RuntimeException` class represents an exception thrown from WebAssembly, or an exception that is constructed in JavaScript and is to be thrown to a WebAssembly exception handler. The `RuntimeException` constructor accepts an `Exception` argument and a sequence of arguments for the exception's data fields. The `Exception` argument determines the exception tag to use. The data field arguments must match the types specified by the `Exception`'s type. The `is` method can be used to query if the `RuntimeException` matches a given `Exception`'s exception tag. The `getArg` method allows access to the data fields of a `RuntimeException` if an `Exception` is passed with a matching exception tag. This last check ensures that without access to a WebAssembly module's exported exception, the associated data fields cannot be read.
```
aheejin: > The `getArg` method allows access to the data fields of a `RuntimeException` if an `Exception` is passed with a matching exception tag. This last check ensures that without access to a WebAssembly module's exported exception, the associated data fields cannot be read.

Why is this check necessary?

proposals/exception-handling/Exceptions.md
```diff
@@ -395,6 +395,36 @@ specially mark non-catchable exceptions.
...
+
+More formally, the added interfaces look like the following:
+
+```
```
aheejin:
```suggestion
```WebIDL
```

proposals/exception-handling/Exceptions.md
```diff
@@ -395,6 +395,36 @@ specially mark non-catchable exceptions.
...
+interface Exception {
+  constructor(ExceptionType type);
+  ExceptionType type();
+};
+
+[LegacyNamespace=WebAssembly, Exposed=(Window,Worker,Worklet)]
+interface RuntimeException {
+  constructor(Exception exceptionType, sequence<any> payload);
+  any getArg(Exception exceptionType, unsigned long index);
+  boolean is(Exception exceptionType);
```
aheejin: How about `tag` instead of `exceptionType`?

aheejin (commented):
Thanks a lot for doing this! Sorry for the delayed reply. This looks good to me, with one thing to consider, as I also commented in https://github.com/WebAssembly/exception-handling/issues/150#issuecomment-829853628: I'm OK with `Exception` / `RuntimeException` names here, but should we rename `event` to `exception` altogether in the core spec? What do you think, @rossberg @ioannad ?

proposals/exception-handling/Exceptions.md
takikawa: I believe the issue is that it would give JS capabilities that the Wasm semantics is not intended to have. Specifically Wasm code cannot ever extract the values for a thrown exception if it doesn't have the tag for it. But if JS could do this via `getArg` then Wasm could also do it by calling out to JS.

(original discussion of this check is here https://github.com/WebAssembly/exception-handling/issues/109#issuecomment-639447310)

ioannad:
@aheejin 

> Thanks a lot for doing this! Sorry for the delayed reply. This looks good to me, with one thing to consider, as I also commented in [#150 (comment)](https://github.com/WebAssembly/exception-handling/issues/150#issuecomment-829853628): I'm OK with `Exception` / `RuntimeException` names here, but should we rename `event` to `exception` altogether in the core spec? What do you think, @rossberg @ioannad ?

Sounds good to me! Since we have the attribute-bit in the binary format of an exception (event), in case we do add other events in future proposals, I think it's simpler to just use the term "exception" in this proposal.

aheejin:
> @aheejin
> 
> > Thanks a lot for doing this! Sorry for the delayed reply. This looks good to me, with one thing to consider, as I also commented in [#150 (comment)](https://github.com/WebAssembly/exception-handling/issues/150#issuecomment-829853628): I'm OK with `Exception` / `RuntimeException` names here, but should we rename `event` to `exception` altogether in the core spec? What do you think, @rossberg @ioannad ?
> 
> Sounds good to me! Since we have the attribute-bit in the binary format of an exception (event), in case we do add other events in future proposals, I think it's simpler to just use the term "exception" in this proposal.

I'm not sure if I understand. Are we still planning to add other kinds of events to the event section? If so and the attribute bit signals the intent, we should consider rather keeping the name 'event', because exception can be considered as a subtype of event but not vice versa.

What I'm saying is, we would better be consistent - if we are quite sure we are not going to add other kinds of events to the current event section, we can rename it as the exception section and use the same names in the JS API. If we plan to add other kinds of events in the event section or are not sure, we may need to keep the name, and using 'exception' in the JS API becomes weird, because, is that API going to only handle exceptions and not other kinds of potential future events?

takikawa:
I updated this PR to match the proposal in #161 for tags.

aheejin (approved):
Thank you!

PR #155, opened 2021-06-02
[js-api] Add spec tests for proposed API additions

takikawa:
This PR adds preliminary web-platform / JS API spec tests for the proposed exception handling API from #154 and discussed in #150.


test/js-api/runtimeexception/getArg.any.js
```diff
@@ -0,0 +1,58 @@
...
+  const exn = new WebAssembly.Exception({ "parameters": [] })
+  const runtimeExn = new WebAssembly.RuntimeException(exn);
+  for (const value of outOfRangeValues) {
+    assert_throws_js(TypeError, () => runtimeExn.getArg(exn, value));
+  }
+}, "Getting out-of-range argument");
+
+test(() => {
+  const exn = new WebAssembly.Exception({ "parameters": ["i32"] })
+  const runtimeExn = new WebAssembly.RuntimeException(exn, 42);
```
Ms2ger: #154 says the second argument should be a `sequence` (i.e., an array-like), so this should throw. Do we want variadic arguments instead?

test/js-api/runtimeexception/is.any.js
```diff
@@ -0,0 +1,33 @@
...
+  const runtimeExn = new WebAssembly.RuntimeException(exn);
+  for (argument of invalidValues) {
+    assert_throws_js(TypeError, () => runtimeExn.is(argument));
+  }
+}, "Invalid exception argument");
+
+test(() => {
+  const exn1 = new WebAssembly.Exception({ "parameters": ["i32"] })
+  const exn2 = new WebAssembly.Exception({ "parameters": ["i32"] })
+  const runtimeExn = new WebAssembly.RuntimeException(exn1, 42);
```
Ms2ger: Ditto

test/js-api/runtimeexception/toString.any.js
```diff
@@ -0,0 +1,18 @@
+// META: global=window,dedicatedworker,jsshell
+
+test(() => {
+  const argument = { "parameters": [] };
+  const exception = new WebAssembly.Exception(argument);
+  const runtimeException = new WebAssembly.RuntimeException(exception);
```
Ms2ger: Same: you'd need to pass an empty array.

test/js-api/runtimeexception/constructor.any.js
```diff
@@ -0,0 +1,47 @@
+// META: global=window,dedicatedworker,jsshell
+// META: script=/wasm/jsapi/assertions.js
+
+test(() => {
+  assert_function_name(WebAssembly.RuntimeException, "RuntimeException", "WebAssembly.RuntimeException");
+}, "name");
+
+test(() => {
+  assert_function_length(WebAssembly.RuntimeException, 1, "WebAssembly.RuntimeException");
```
Ms2ger: Should be two.

Ms2ger (commented):
These should all have the `.tentative.` flag until the proposal merges into the main spec.

test/js-api/runtimeexception/getArg.any.js
takikawa: Thanks for catching this, I had implemented this wrong in my prototype with variadic args and it leaked into the tests. I think it makes more sense to use a sequence argument here as the spec draft says.

(should be fixed in latest commit)

takikawa:
This PR is now updated to match the recent event to tag renaming in the spec.

Ms2ger (approved):
Anything blocking this?

test/js-api/tag/type.tentative.any.js
```diff
@@ -0,0 +1,21 @@
+// META: global=window,dedicatedworker,jsshell
+// META: script=/wasm/jsapi/assertions.js
+
+function assert_type(argument) {
+    const tag = new WebAssembly.Tag(argument);
```
aheejin: It looks some functions use two spaces while others use four spaces. What's our rule for other JS spec tests? It'd better be consistent.

aheejin (commented):
Thanks for doing this! Sorry for the late reply. Is there any reason file names have 'tentative' in them, even if they are tentative? (Aren't all files tentative in a sense, unless it is standardized..?) Is that a rule in the spec repo?

Anyway can you wait for a little more while before merging?

takikawa:
> Is there any reason file names have 'tentative' in them, even if they are tentative? (Aren't all files tentative in a sense, unless it is standardized..?) Is that a rule in the spec repo?

I'm not sure exactly what the rules are, @Ms2ger would know better what the standard practice is here I think, but I think the idea was to keep it tentative until the proposal spec is merged into the main spec.

The test framework docs says this about the `tentative` flag:

> .tentative : Indicates that a test makes assertions not yet required by any specification, or in contradiction to some specification. This is useful when implementation experience is needed to inform the specification. It should be apparent in context why the test is tentative and what needs to be resolved to make it non-tentative.

BTW, I just noticed it's also possible to set the whole directory as tentative. Would it be better if I did that for these tests?

> Anyway can you wait for a little more while before merging?

That's ok with me. I guess it would make sense to wait until the JS API spec text is merged too, is there anything else it should block on?

test/js-api/tag/type.tentative.any.js
takikawa: Thanks for noticing that, I think 2-space is the norm based on other tests. I've just reformatted all the files I added with prettier to be consistent with 2-space. @Ms2ger is that fine for spec test style? (would be happy to apply a different linter too if there's a recommended one)

Ms2ger:
These tests are all going into WPT as well, as soon as they're merged here, so are marked as tentative until phase 5. I forgot about marking the directory as tentative; no preference either way on that.

test/js-api/exception/constructor.tentative.any.js
```diff
@@ -0,0 +1,62 @@
...
+}, "length");
+
+test(() => {
+  assert_throws_js(TypeError, () => new WebAssembly.Exception());
+}, "No arguments");
+
+test(() => {
+  const argument = new WebAssembly.Tag({ parameters: [] });
+  assert_throws_js(TypeError, () => WebAssembly.Exception(argument));
+}, "Calling");
```
aheejin, line 23: Why does this throw?

test/js-api/exception/type.tentative.any.js
```diff
@@ -0,0 +1,21 @@
+// META: global=window,dedicatedworker,jsshell
+// META: script=/wasm/jsapi/assertions.js
+
+function assert_type(argument) {
+    const exception = new WebAssembly.Exception(argument);
```
aheejin: The indentation for this file is 4 spaces, unlike other files

aheejin (commented):
Sorry for the delayed reply and thanks for your patience. Also sorry I'm not very familiar with the spec tests convention, so I might ask easy questions.

- For the usage of 'tentative': Then are other wasm proposal repos all using 'tentative' for their JS API tests until they reach Phase 5?
- How can I run these JS API tests?

test/js-api/exception/constructor.tentative.any.js
Ms2ger: The issue here is the absence of the `new` keyword, which leads to `NewTarget` being **undefined**, which causes an exception in step 1.2 of the ["create an interface object"](https://heycam.github.io/webidl/#create-an-interface-object) algorithm.

test/js-api/tag/constructor.tentative.any.js
```diff
@@ -0,0 +1,49 @@
...
+}, "length");
+
+test(() => {
+  assert_throws_js(TypeError, () => new WebAssembly.Tag());
+}, "No arguments");
+
+test(() => {
+  const argument = { parameters: [] };
+  assert_throws_js(TypeError, () => WebAssembly.Tag(argument));
+}, "Calling");
```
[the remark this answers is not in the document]
Ms2ger: Ditto.

Ms2ger (commented):
> For the usage of 'tentative': Then are other wasm proposal repos all using 'tentative' for their JS API tests until they reach Phase 5?

If I did the review, they should. I can try to find a place to document this as well.

> How can I run these JS API tests?

Once they're in WPT, they'll be published at wpt.live ([example](http://wpt.live/wasm/jsapi/global/constructor.any.html)).

aheejin (approved):
Thanks for the explanation!

PR #157, opened 2021-06-07
Allow catch/delegate-less trys

### LINKED ISSUE #131: Could the different `try` variants be combined?

aheejin:
This was suggested here:
https://github.com/WebAssembly/exception-handling/issues/131#issuecomment-784048075

Closes #131.

### DIFF
--- a/proposals/exception-handling/Exceptions.md
+++ b/proposals/exception-handling/Exceptions.md
@@ -122,7 +122,7 @@ A try-catch block contains zero or more `catch` blocks and zero or one
 `catch_all` block. All `catch` blocks must precede the `catch_all` block, if
 any. The `catch`/`catch_all` instructions (within the try construct) are called
 the _catching_ instructions. There may not be any `catch` or `catch_all` blocks
-after a `try`, at which point the whole `try` block is effectively a regular
+after a `try`, in which case the whole `try` block is effectively a regular
 block.
 
 The _body_ of the try block is the list of instructions before the first
--- a/proposals/exception-handling/Exceptions.md
+++ b/proposals/exception-handling/Exceptions.md
@@ -121,8 +121,9 @@ end
 A try-catch block contains zero or more `catch` blocks and zero or one
 `catch_all` block. All `catch` blocks must precede the `catch_all` block, if
 any. The `catch`/`catch_all` instructions (within the try construct) are called
-the _catching_ instructions. There should be at least one `catch` or `catch_all`
-block within a try-catch block.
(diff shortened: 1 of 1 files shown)


proposals/exception-handling/Exceptions.md
```diff
@@ -121,8 +121,9 @@ end
 A try-catch block contains zero or more `catch` blocks and zero or one
 `catch_all` block. All `catch` blocks must precede the `catch_all` block, if
 any. The `catch`/`catch_all` instructions (within the try construct) are called
-the _catching_ instructions. There should be at least one `catch` or `catch_all`
-block within a try-catch block.
+the _catching_ instructions. There may not be any `catch` or `catch_all` blocks
+after a `try`, at which point the whole `try` block is effectively a regular
```
tlively: language nit:

```suggestion
after a `try`, in which case the whole `try` block is effectively a regular
```

tlively (approved):

ioannad (approved):

rossberg (approved):
Thanks!

takikawa (approved):

PR #160, opened 2021-06-09
[interpreter] Parse and convert EH opcodes

thibaudmichaud:
Please take a look. This adds support for EH opcodes in the parser, AST, encoder, decoder and formatter, and adds core spec tests. This follows the latest discussions that seem to have reached agreement: unwind is not supported (#156), and try blocks may have no handlers attached to them (#157).

This can already be used to convert the tests to JS, so this should be a useful starting point to ensure consistency across JS engines. The interpreter itself is still missing validation and execution.

### DIFF
--- a/interpreter/binary/decode.ml
+++ b/interpreter/binary/decode.ml
@@ -246,7 +246,30 @@ let rec instr s =
     end
 
   | 0x05 -> error s pos "misplaced ELSE opcode"
-  | 0x06| 0x07 | 0x08 | 0x09 | 0x0a as b -> illegal s pos b
+  | 0x06 ->
+    let bt = block_type s in
+    let es = instr_block s in
+    let ct = catch_list s in
+    let ca =
+      if peek s = Some 0x19 then begin
+        ignore (u8 s);
+        Some (instr_block s)
+      end else
+        None
+    in
+    if ct <> [] || ca <> None then begin
+      end_ s;
+      try_catch bt es ct ca
+    end else begin
+      match op s with
+      | 0x0b -> try_catch bt es [] None
+      | 0x18 -> try_delegate bt es (at var s)
+      | b -> illegal s pos b
+    end
+  | 0x07 -> error s pos "misplaced CATCH opcode"
+  | 0x08 -> throw (at var s)
+  | 0x09 -> rethrow (at var s)
+  | 0x0a as b -> illegal s pos b
   | 0x0b -> error s pos "misplaced END opcode"
 
   | 0x0c -> br (at var s)
@@ -263,7 +286,10 @@ let rec instr s =
     let x = at var s in
     call_indirect x y
 
-  | 0x12 | 0x13 | 0x14 | 0x15 | 0x16 | 0x17 | 0x18 | 0x19 as b -> illegal s pos b
+  | 0x12 | 0x13 | 0x14 | 0x15 | 0x16 | 0x17 as b -> illegal s pos b
+
+  | 0x18 -> error s pos "misplaced DELEGATE opcode"
+  | 0x19 -> error s pos "misplaced CATCH_ALL opcode"
 
   | 0x1a -> drop
   | 0x1b -> select None
@@ -499,11 +525,19 @@ let rec instr s =
 and instr_block s = List.rev (instr_block' s [])
 and instr_block' s es =
   match peek s with
-  | None | Some (0x05 | 0x0b) -> es
+  | None | Some (0x05 | 0x07 | 0x0a | 0x0b | 0x18 | 0x19) -> es
   | _ ->
     let pos = pos s in
     let e' = instr s in
     instr_block' s (Source.(e' @@ region s pos pos) :: es)
+and catch_list s =
+  if peek s = Some 0x07 then begin
+    ignore (u8 s);
+    let tag = at var s in
+    let instrs = instr_block s in
+    (tag, instrs) :: catch_list s
+  end else
+    []
 
 let const s =
   let c = at instr_block s in
--- a/interpreter/syntax/ast.ml
+++ b/interpreter/syntax/ast.ml
@@ -114,6 +114,13 @@ and instr' =
   | Unary of unop                     (* unary numeric operator *)
   | Binary of binop                   (* binary numeric operator *)
   | Convert of cvtop                  (* conversion *)
+  | TryCatch of block_type * instr list * (* try *)
+                (var * instr list) list * (* catch exception with tag *)
+                instr list option     (* catch_all *)
+  | TryDelegate of block_type * instr list * (* try *)
+                   var                (* delegate to outer handler *)
+  | Throw of var                      (* throw exception *)
+  | Rethrow of var                    (* rethrow exception *)
 
 
 (* Globals & Functions *)
--- a/interpreter/syntax/free.ml
+++ b/interpreter/syntax/free.ml
@@ -87,6 +87,15 @@ let rec instr (e : instr) =
     memories zero
   | MemoryInit x -> memories zero ++ datas (var x)
   | DataDrop x -> datas (var x)
+  | TryCatch (bt, es, ct, ca) ->
+    let catch (tag, es) = events (var tag) ++ block es in
+    let catch_all = function
+      | None -> empty
+      | Some es -> block es in
+    block es ++ (list catch ct) ++ catch_all ca
+  | TryDelegate (bt, es, x) -> block es ++ events (var x)
+  | Throw x ->  events (var x)
+  | Rethrow x ->  labels (var x)
 
 and block (es : instr list) =
   let free = list instr es in {free with labels = shift free.labels}
--- a/test/core/rethrow.wast
+++ b/test/core/rethrow.wast
@@ -1,0 +1,83 @@
+;; Test rethrow instruction.
+
+(module
+  (event $e0)
+  (event $e1)
+
+  (func (export "catch-rethrow-0")
+    (try
+      (do (throw $e0))
+      (catch $e0 (rethrow 0))
+    )
+  )
+
+  (func (export "catch-rethrow-1") (param i32) (result i32)
+    (try (result i32)
+      (do (throw $e0))
+      (catch $e0
+        (if (i32.eqz (local.get 0)) (then (rethrow 1))) (i32.const 23)
+      )
+    )
+  )
+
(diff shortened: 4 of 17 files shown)


interpreter/syntax/ast.ml
```diff
@@ -114,6 +114,14 @@ and instr' =
   | Unary of unop                     (* unary numeric operator *)
   | Binary of binop                   (* binary numeric operator *)
   | Convert of cvtop                  (* conversion *)
+  | Try of block_type * instr list *  (* try *)
+           handler                    (* handle exception *)
+  | Throw of var                      (* throw exception *)
+  | Rethrow of var                    (* rethrow exception *)
+and handler =
```
rossberg: It's not worth introducing an auxiliary type here (and auxiliary functions for it in various places). Simply make it two separate instructions, TryCatch and TryDelegate. That they share an opcode/keyword is best considered a detail of the binary & textual representation.

interpreter/binary/decode.ml
```diff
@@ -263,7 +271,9 @@ let rec instr s =
     let x = at var s in
     call_indirect x y
 
-  | 0x12 | 0x13 | 0x14 | 0x15 | 0x16 | 0x17 | 0x18 | 0x19 as b -> illegal s pos b
+  | 0x12 | 0x13 | 0x14 | 0x15 | 0x16 | 0x17 as b -> illegal s pos b
+  | 0x18 -> error s pos "misplaced DELEGATE opcode"
+  | 0x19 -> error s pos "misplaced CATCH_ALL opcode"
```
rossberg, line 292:
```suggestion

  | 0x18 -> error s pos "misplaced DELEGATE opcode"
  | 0x19 -> error s pos "misplaced CATCH_ALL opcode"
```

interpreter/binary/decode.ml
```diff
@@ -499,12 +509,39 @@ let rec instr s =
...
   match peek s with
-  | None | Some (0x05 | 0x0b) -> es
+  | None | Some (0x05 | 0x07 | 0x0a | 0x0b | 0x18 | 0x19) -> es
   | _ ->
     let pos = pos s in
     let e' = instr s in
     instr_block' s (Source.(e' @@ region s pos pos) :: es)
 
+and handler s =
+  let rec catch_list () =
```
rossberg: Nit: avoid nesting functions unless they close over local variables in a non-trivial fashion.

interpreter/syntax/free.ml
```diff
@@ -87,10 +87,24 @@ let rec instr (e : instr) =
...
+  | Try (bt, es, h) -> block es ++ handler h
+  | Throw x ->  events (var x)
+  | Rethrow x ->  labels (var x)
 
 and block (es : instr list) =
   let free = list instr es in {free with labels = shift free.labels}
 
+and handler = function
+  | Catch (cs, catch_all) ->
+    let rec catch_list = function
```
rossberg: Same nit here.

test/core/exceptions.wast
```diff
@@ -0,0 +1,429 @@
+;; Test exception handling support
```
rossberg: Can you split this up into files focussing on individual instructions?

throw.wast
rethrow.wast
try_catch.wast
try_delegate.wast

interpreter/syntax/ast.ml
thibaudmichaud: Done.

interpreter/binary/decode.ml
thibaudmichaud: Done.

interpreter/binary/decode.ml
thibaudmichaud: Done.

interpreter/syntax/free.ml
thibaudmichaud: Done.

test/core/exceptions.wast
thibaudmichaud: Done.

takikawa:
I was able to run these tests for SpiderMonkey as well using its test harness for spec tests. After some modifications to account for recent changes (e.g., bare `try` blocks) and test harness issues, I believe SM also agrees on all of these tests as well.

One issue I ran into, however, is that the core tests are using `assert_trap` to test uncaught exception cases. Would it make sense to instead extend the interpreter with a new assertion `(assert_uncaught_exception <action>)` for these cases? (since the spec defines uncaught exceptions as host-dependent behavior rather than a trap necessarily)

rossberg (approved):

interpreter/binary/decode.ml
```diff
@@ -246,7 +246,29 @@ let rec instr s =
...
+    in
+    if ct <> [] || ca <> None then begin
+      end_ s;
+      try_catch bt es ct ca
+    end else
+      begin match op s with
+      | 0x0b -> try_catch bt es [] None
+      | 0x18 -> try_delegate bt es (at var s)
+      | b -> illegal s pos b
+      end
```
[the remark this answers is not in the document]
thibaudmichaud: Done

thibaudmichaud:
> One issue I ran into, however, is that the core tests are using assert_trap to test uncaught exception cases. Would it make sense to instead extend the interpreter with a new assertion (assert_uncaught_exception <action>) for these cases? (since the spec defines uncaught exceptions as host-dependent behavior rather than a trap necessarily)

I think that makes sense and I just updated the PR with that suggestion. The JS implementation of the rule uses a simple try/catch to accept _any_ exception.

I was actually planning to add such a rule later that uses the JS API to also check the type and content of the exception, but for now I think it is sufficient to add the weak form to solve the issue you pointed out.

takikawa (approved):

takikawa:
> I think that makes sense and I just updated the PR with that suggestion. The JS implementation of the rule uses a simple try/catch to accept any exception.

Thanks! I just tried the new commit with `assert_uncaught_exception` and I was able to get it to work with the SpiderMonkey spec test harness.

rossberg:
Ah, wanted to comment on the assertion question as well, but forgot.

I agree, we should have a dedicated assertion for that. But can we just call it `assert_exception` though? Assertions are kind of a unit test thing, so "uncaught" seems off, since that's only meaningful in the context of an entire program.

thibaudmichaud:
> I agree, we should have a dedicated assertion for that. But can we just call it assert_exception though? Assertions are kind of a unit test thing, so "uncaught" seems off, since that's only meaningful in the context of an entire program.

Done, both look fine to me.

test/core/try_catch.wast
```diff
@@ -0,0 +1,206 @@
...
+  (event $e0 (export "e0"))
+  (func (export "throw") (throw $e0))
+)
+
+(register "test")
+
+(module
+  (event $imported-e0 (import "test" "e0"))
+  (func $imported-throw (import "test" "throw"))
+  (event $e0)
```
aheejin, line 13: Assuming that we are changing this to `(tag ...)` (not necessarily in this PR), do we need a text syntax to represent the attribute (exception)? Currently we are tentatively using 
```wasm
(event $e0 (attr 0))
```
in Binaryen, but not sure if this is a good one. (I uploaded a PR to change this in Binaryen to `(tag ..)` in WebAssembly/binaryen#3937, but this `attr` thing is a separate thing anyway)

aheejin (commented):
Thanks! I haven't read the ML implementation part yet but the test cases here match my understanding.

test/core/rethrow.wast
```diff
@@ -0,0 +1,83 @@
...
+
+  (func (export "rethrow-nested") (param i32) (result i32)
+    (try (result i32)
+      (do (throw $e1))
+      (catch $e1
+        (try (result i32)
+          (do (throw $e0))
+          (catch $e0
+            (if (i32.eq (local.get 0) (i32.const 0)) (then (rethrow 1)))
+	    (if (i32.eq (local.get 0) (i32.const 1)) (then (rethrow 2)))
```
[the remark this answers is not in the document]
thibaudmichaud: Done.

test/core/rethrow.wast
```diff
@@ -0,0 +1,83 @@
...
+    )
+  )
+
+  (func (export "rethrow-recatch") (param i32) (result i32)
+    (try (result i32)
+      (do (throw $e0))
+      (catch $e0
+        (try (result i32)
+	  (do (if (i32.eqz (local.get 0)) (then (rethrow 2))) (i32.const 42))
+	  (catch $e0 (i32.const 23))
```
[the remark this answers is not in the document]
thibaudmichaud: Done.

test/core/try_delegate.wast
```diff
@@ -0,0 +1,115 @@
...
+(assert_return (invoke "delegate-no-throw") (i32.const 1))
+
+(assert_return (invoke "delegate-throw" (i32.const 0)) (i32.const 1))
+(assert_return (invoke "delegate-throw" (i32.const 1)) (i32.const 2))
+
+(assert_exception (invoke "delegate-throw-no-catch"))
+
+(assert_return (invoke "delegate-merge" (i32.const 1) (i32.const 0)) (i32.const 2))
+(assert_exception (invoke "delegate-merge" (i32.const 2) (i32.const 0)))
+(assert_return (invoke "delegate-merge" (i32.const 1)) (i32.const 2))
```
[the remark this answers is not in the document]
thibaudmichaud: It should be ok but in this case that was a mistake, thanks for noticing! Fixed.

test/core/try_catch.wast
rossberg: As argued elsewhere, I don't think it is necessary to manifest the (so far vacuous) attribute in the text format for the time being. It's not even necessary to put it in the semantics. All we need right now is a reserved zero byte in the binary format, until we know better what possible distinctions are needed in the future.

thibaudmichaud: So if we eventually need a different tag kind and go with @aheejin's syntax, exception tags would be a special case as shown below?
```
(tag)          ;; an exception tag
(tag (attr 0)) ;; an exception tag
(tag (attr 1)) ;; some other tag
```
I am fine with that, just want to clarify.

aheejin: Sounds good. I'll remove the `attr` attribute from the Binaryen text format then.

ioannad (approved):
I agree with @aheejin that the change from `event` to `tag` could be done in a different PR, so everything LGTM. Sorry for the delayed review!

PR #161, opened 2021-06-11
Replace 'event' with 'tag' in explainer

### LINKED ISSUE #159: Event vs. exception?
In the core spec, we are currently using the term 'event' to represent a tag that can be used to match an exception object in `catch` instructions. An event contains a signature, in which the result type is always void. Two different event can contain the same signature.

The term 'event' was suggested years ago out of concern that there will be more kinds of events that will be added to the event section. See https://github.com/WebAssembly/exception-handling/blob/master/proposals/exception-handling/old/Exceptions-v2-Level-1%2BN.md#level-3 for examples of thought then.

Recently JS API addition proposal was suggested in #154 and it uses the term `exception` to represent an exception, which I think fine, but the consistency in the terminology in the core spec and the JS API comes to mind. Also https://github.com/WebAssembly/exception-handling/issues/150#issuecomment-814854657 has several links to previous comments on possible event vs. exception terminology.

I am OK with either, but I think we should be consistent on the terminology in the core spec and the JS API spec.

Do we still plan to add other kinds of events to the event section? Are other stack switching or coroutine primitives planning to use the section?

cc @rossberg @ioannad @takikawa @RossTate @tlively

aheejin:
It was suggested to change the term 'event' to 'tag' to reference the
section name and the entries within the section. The suggested reasons
were they can be used something other than events in future, and the
term event can be confusing with other concepts on the web.

Closes #159.


proposals/exception-handling/Exceptions.md
```diff
@@ -452,42 +444,42 @@ So the list of all sections will be:
...
 The export section is extended to reference exception types by extending an
 `export_entry` as follows:
 
-If the `kind` is `Event`:
+If the `kind` is `Tag`:
 
 | Field | Type | Description |
 |-------|------|-------------|
-| `index` | `varuint32` | the index into the corresponding event index space |
+| `index` | `varuint32` | the index into the corresponding exception index space |
```
tlively:
```suggestion
| `index` | `varuint32` | the index into the corresponding tag index space |
```

I would expect the tag section and tag imports to define a tag index space.

proposals/exception-handling/Exceptions.md
```diff
@@ -452,42 +444,42 @@ So the list of all sections will be:
...
 
-The event section declares a list of event types as follows:
+The tag section declares a list of tag types as follows:
 
 | Field | Type | Description |
 |-------|------|-------------|
-| count | `varuint32` | count of the number of event to follow |
-| type | `event_type*` | The definitions of the event types |
+| count | `varuint32` | count of the number of tags to follow |
+| type | `tag_type*` | The definitions of the tag types |
```
tlively, line 450: Is `tag_type` defined anywhere?

proposals/exception-handling/Exceptions.md
```diff
@@ -418,28 +417,21 @@ Each exception type has the fields:
...
+[definition](#tag-section)
 
 ### Module structure
 
 #### High-level structure
 
-A new `event` section is introduced and is named `event`. If included, it must
-appear immediately after the memory section.
+A new `tag` section is introduced. If included, it must appear immediately after
+the memory section.
```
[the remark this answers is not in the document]
aheejin: Thanks; removed.

proposals/exception-handling/Exceptions.md
aheejin: While creating this PR, I found multiple occurrences that we had missed when we had replaced 'exception' with 'event' before, and I tried to replace 'event' with 'tag' in this PR, so I missed those remaining 'exception's.. It was written as `exception_type`. I changed it to `tag_type` now.

proposals/exception-handling/Exceptions.md
```diff
@@ -45,12 +45,12 @@ instruction. Thrown exceptions are handled as follows:
...
+### Exception handling
 
 This proposal adds exception handling to WebAssembly. Part of this proposal is
 to define a new section to declare exceptions. However, rather than limiting
 this new section to just defining exceptions, it defines a more general format
-that allows the declaration of other forms of events.
+that allows the declaration of other forms of events in future.
 
 In general, an event handler allows one to process an event generated by a block
 of code. Events suspend the current execution and look for a corresponding event
```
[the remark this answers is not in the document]
aheejin: Thanks for the suggestion; I think the distinction makes the text a lot clearer. Fixed many parts of the existing text.

proposals/exception-handling/Exceptions.md
```diff
@@ -45,50 +45,40 @@ instruction. Thrown exceptions are handled as follows:
...
-that allows the declaration of other forms of events.
+`tag` that allows the declaration of other forms of typed tags in future.
 
-In general, an event handler allows one to process an event generated by a block
-of code. Events suspend the current execution and look for a corresponding event
-handler. If found, the corresponding event handler is run. Some event handlers
-may send values back to the suspended instruction, allowing the originating code
-to resume.
+WebAssembly tags are defined in a new `tag` section of a WebAssembly module. The
+tag section is a list of declared tags associated with the module.
```
rossberg: Nit: "associated with the module" isn't very clear.
```suggestion
tag section is a list of declared tags that are created fresh each time the module is instantiated.
```

proposals/exception-handling/Exceptions.md
```diff
@@ -45,50 +45,40 @@ instruction. Thrown exceptions are handled as follows:
...
+attribute.
 
 ### Exceptions
 
-An `exception` is an internal construct in WebAssembly. WebAssembly exceptions
-are defined in the event and import sections of a module.
+An `exception tag` is a value to distinguish different exceptions, while an
+`exception tag index` is a numeric name to refer to an (imported or defined)
+exception tag within a module (see [tag index space](#tag-index-space) for
+details). Exception tags are defined in the tag and import sections of a module.
```
rossberg: Nit (because an import is not a definition in itself):
```suggestion
details). Exception tags are declared in the tag and import sections of a module.
```

proposals/exception-handling/Exceptions.md
```diff
@@ -45,50 +45,40 @@ instruction. Thrown exceptions are handled as follows:
...
 
-The type of an exception is denoted by an index to a function signature defined
-in the `type` section. The parameters of the function signature define the list
-of values associated with the exception. The result type must be empty.
+An `exception` is an internal construct in WebAssembly that represents a runtime
+object that can be thrown. A WebAssembly exception is defined by an exception
```
```suggestion
object that can be thrown. A WebAssembly exception consists of an exception
```

proposals/exception-handling/Exceptions.md
```diff
@@ -45,50 +45,40 @@ instruction. Thrown exceptions are handled as follows:
...
+tag and its runtime arguments.
 
-An `exception tag` is a value to distinguish different exceptions, while an
-`exception index` is a numeric name to refer to an (imported or defined)
-exception tag within a module (see [exception index
-space](#exception-index-space) for details).
+The type of an exception tag is denoted by an index to a function signature
+defined in the `type` section. The parameters of the function signature define
+the list of values associated with the tag. The result type must be empty.
```
```suggestion
the list of argument values associated with the tag. The result type must be empty.
```

proposals/exception-handling/Exceptions.md
```diff
@@ -398,48 +386,39 @@ document](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md).
...
 |-----------|-------|
 | Exception | 0     |
 
-Each exception type has the fields:
+Each tag type has the fields:
 
 | Field | Type | Description |
 |-------|------|-------------|
-| `attribute` | `varuint32` | The attribute of an exception. |
+| `attribute` | `varuint32` | The attribute of a tag. |
```
rossberg, line 404: Obviously a separate issue, but I'm just noticing this while I'm here: I'd conservatively make this a `uint8` for the time being, since we do not know yet how exactly this is gonna be used in the future, and a zero `uint8` is still forward compatible with `varuint32`. That's what we did in other places anyway.

rossberg (approved):
Looks good, mostly nits.

proposals/exception-handling/Exceptions.md
aheejin: I can do that as a followup PR. But why `uint8` better than `varuint32` here?

takikawa (approved):

proposals/exception-handling/Exceptions.md
rossberg: It's both more conservative and faster to decode. No reason to bother with allowing multibyte representations of 0 before we even know that we'll need them.

PR #168, opened 2021-06-28
Clarify catchless try in explainer

### LINKED ISSUE #164: Behavior for try-delegate targeting catch-less try?
Something I overlooked when reviewing #157 was what the behavior of `try-delegate` should be when it targets a catchless `try` block. For example:

```
try $l
  try
  delegate $l
end
```

I think I didn't see discussion of this particular case anywhere, sorry if I missed it. A semantics that might make sense is that this will just delegate to the next enclosing catch block (effectively rethrows from `try $l`). Any thoughts on how this should be specified?

aheejin:
- Removes the sentence that catchless `try` is the same as a regular
`block`, because it can be targetted by `delegate`.
- Improves a few sentences
- Adds examples for catchless try

Closes #164.

### DIFF
--- a/proposals/exception-handling/Exceptions.md
+++ b/proposals/exception-handling/Exceptions.md
@@ -115,8 +115,8 @@ A try-catch block contains zero or more `catch` blocks and zero or one
 `catch_all` block. All `catch` blocks must precede the `catch_all` block, if
 any. The `catch`/`catch_all` instructions (within the try construct) are called
 the _catching_ instructions. There may not be any `catch` or `catch_all` blocks
-after a `try`, in which case the whole `try` block is effectively a regular
-block.
+after a `try`, in which case the `try` block does not catch any exceptions and
+rethrows them.
 
 The _body_ of the try block is the list of instructions before the first
 catching instruction. The _body_ of each catch block is the sequence of
@@ -258,6 +258,18 @@ end
 The `rethrow` here references `try $l2`, but the `rethrow` is not within its
 `catch` block.
 
+Also, targetting a label of `catch`-less `try` or `try`-`delegate` is also a
+validation failure, because `rethrow` is not within a `catch` or `catch_all`.
+```
+try $l1
+  rethrow $l1  ;; validation failure!
+delegate
+
+try $l2
+  rethrow $l2  ;; validation failure!
+end
+```
+
 ### Try-delegate blocks
 
 Try blocks can also be used with the `delegate` instruction. A try-delegate
@@ -271,9 +283,8 @@ delegate label
 
 The `delegate` clause does not have an associated body, so try-delegate blocks
 don't have an `end` instruction at the end. The `delegate` instruction takes a
-label defined by a construct in which they are enclosed, and delegates exception
-handling to a catch block specified by the label. For example, consider this
-code:
+try label and delegates exception handling to a `catch`/`catch_all`/`delegate`
+specified by the `try` label. For example, consider this code:
 
 ```
 try $l1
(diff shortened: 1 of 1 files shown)


proposals/exception-handling/Exceptions.md
```diff
@@ -115,8 +115,8 @@ A try-catch block contains zero or more `catch` blocks and zero or one
 `catch_all` block. All `catch` blocks must precede the `catch_all` block, if
 any. The `catch`/`catch_all` instructions (within the try construct) are called
 the _catching_ instructions. There may not be any `catch` or `catch_all` blocks
-after a `try`, in which case the whole `try` block is effectively a regular
-block.
+after a `try`, in which case the `try` block does not catch any exceptions and
+rethrows them.
```
```suggestion
after a `try`, in which case the `try` block does not catch any exceptions.
```
(I think the rethrow part is a bit off, since technically, a non-catch isn't the same as a rethrow.)

proposals/exception-handling/Exceptions.md
```diff
@@ -258,6 +258,18 @@ end
 The `rethrow` here references `try $l2`, but the `rethrow` is not within its
 `catch` block.
 
+Also, targetting a label of `catch`-less `try` or `try`-`delegate` is also a
```
rossberg: I still don't see the relevance of the "catch-less" here, I think it's misleading: the example is a validation failure regardless of whether you add a catch clause or not. Both examples are invalid simply because rethrow does not reference a catch-block, no other reason.

proposals/exception-handling/Exceptions.md
```diff
@@ -258,6 +258,18 @@ end
...
+validation failure, because `rethrow` is not within a `catch` or `catch_all`.
```
rossberg: This part is confusing as well: even if there was a surrounding catch somewhere it would still be invalid. Whether a catch _exists_ is irrelevant, it only matters whether the rethrow _references_ a catch.

proposals/exception-handling/Exceptions.md
aheejin: I guess the whole added paragraph here about `rethrow` is not really meaningful and also misleading as you suggested. This PR's goal was to clarify about catchless try and its relationship w/ `delegate`, so I'll delete this whole paragraph and example.

proposals/exception-handling/Exceptions.md
aheejin: Deleted.

proposals/exception-handling/Exceptions.md
```diff
@@ -241,9 +240,9 @@ caught by `catch 1`. In wat format, the argument for the `rethrow` instructions
 can also be written as a label, like branches. So `rethrow 0` in the example
 above can also be written as `rethrow $l3`.
 
-Note that `rethrow 2` is not allowed because it does not reference a catch
-block. Rather, it references a `block` instruction, so it will result in a
-validation failure.
+Note that `rethrow 2` is not allowed because it does not refer to a `try` label
+with a catch block. Rather, it references a `block` instruction, so it will
```
```suggestion
from within its catch block. Rather, it references a `block` instruction, so it will
```

proposals/exception-handling/Exceptions.md
```diff
@@ -296,17 +294,22 @@ correspond to a `try` instruction, it is a validation failure.
...
-Here `delegate` is trying to delegate to `catch 1`, which exists before the
-`delegate`. The `delegate` instruction can only delegate to `catch`/`catch_all`
-blocks in a `try` or to another `delegate` below the `delegate` itself.
+Here `delegate` is trying to delegate to `catch`, which exists before the
+`delegate`. The `delegate` instruction can only target `try` label whose
+catching instructions (`catch`/`catch_all`/`delegate`) come after the
+`delegate` instruction.
+
+`delegate` can also target `catch`-less `try`, in which case the effect is the
+same as the `try` has catches but none of the catches are able to handle the
```
```suggestion
same as if the `try` has catches but none of the catches are able to handle the
```

rossberg (approved):

PR #173, opened 2021-07-13
Remove .type() test from js-api tests exceptions folder

takikawa:
This test should have been removed during the events/exception to tag renaming, but looks like it was left in by mistake.

The spec overview and consensus was that the `type()` method is only on the `WebAssembly.Tag` object, and not `WebAssembly.Exception`. The test for the tag object should be fine.


Ms2ger (approved):
Oops

PR #174, opened 2021-07-14
[js-api] Add note about shipping Tag#type().

### LINKED ISSUE #172: Moving WebAssembly.Tag.type to type reflection?
Are there known use cases that require WebAssembly.Tag.type to be part of the exception handling proposal? If not I would be in favor of moving it to the type reflection proposal.

The argument is that exception handling has more support and is advancing faster, so we should not make it depend on a less mature proposal. In particular if changes are made to the type reflection API later, it might be too late to reflect these changes in WebAssembly.Tag.type without breaking backwards compatibility, and would introduce inconcistencies.
The two proposals are technically both at phase 2, but AFAIK type reflection is much less active, and exception handling is close to reaching phase 3 and 4.

cc @gahaas who pointed this out to me.

Ms2ger:
Fixes #172.


thibaudmichaud (approved):
Thanks! Looks good to me.

PR #177, opened 2021-07-16
Clarify delegate targetting function-level block

### LINKED ISSUE #176: Inconsistency in documentation for try-delegate behavior for function body
In #175 there's some discussion of what the intended behavior of `try-delegate` with a label pointing to the function body is supposed to be (https://github.com/WebAssembly/exception-handling/pull/175#discussion_r670152661).

My understanding was it should be valid and rethrow the exception out of the function, as discussed in #146 (see example there, especially for `delegate 3`) and as covered in the core tests [in this test case](https://github.com/WebAssembly/exception-handling/blob/master/test/core/try_delegate.wast#L48) which I'll just inline below:

```wasm
  (func (export "delegate-to-caller")
    (try (do (try (do (throw $e0)) (delegate 1))) (catch_all))
  )
```

There's an inconsistency though with the overview in https://github.com/WebAssembly/exception-handling/blob/master/proposals/exception-handling/Exceptions.md as it says:

> When the specified label within a delegate instruction does not correspond to a try instruction, it is a validation failure.

Assuming we agree on the behavior, we could change the overview to say "does not correspond to a try instruction or function body" (and an example could be helpful) and possibly adjust some other prose.

aheejin:
This clarifies that `delegate` can target the function-level block,
which means the exception is thrown to the caller. Also added a few more
examples.

AFAIK this is consistent with the current implementations in the
toolchain, V8, and Spidermonkey.

Fixes #176.

### DIFF
--- a/proposals/exception-handling/Exceptions.md
+++ b/proposals/exception-handling/Exceptions.md
@@ -310,20 +310,22 @@ catching instructions (`catch`/`catch_all`/`delegate`) come after the
 same as if the `try` has catches but none of the catches are able to handle the
 exception.
 
-If `delegate`'s immediate argument is the depth of block-like structures
-(blocks, loops, and trys) + 1, i.e., the function-level block, it is considered
-to be delegating the exception to the caller of the current function. For
-example:
+If `delegate`'s immediate argument is the depth of the outermost block + 1, i.e.
+the function-level block, the function-level block is considered as a
+`catch`-less `try` and it delegates the exception handling to the caller of the
+current function. For example:
 ```
-try $l1
-  try
-    call $foo
-  delegate 1  ;; delegate to the caller
-catch
-  ...
-catch_all
-  ...
-end
+(func $test
+  try $l1
+    try
+      call $foo
+    delegate 1  ;; delegate to the caller
+  catch
+    ...
+  catch_all
+    ...
+  end
+)
 ```
 In case `foo` throws, `delegate 1` here delegates the exception handling to the
 caller, i.e., the exception escapes the current function.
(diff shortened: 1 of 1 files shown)


proposals/exception-handling/Exceptions.md
```diff
@@ -311,6 +310,39 @@ catching instructions (`catch`/`catch_all`/`delegate`) come after the
 same as if the `try` has catches but none of the catches are able to handle the
 exception.
 
+If `delegate`'s immediate argument is the depth of block-like structures
+(blocks, loops, and trys) + 1, i.e., the function-level block, it is considered
```
thibaudmichaud: I don't understand what you mean by: "the depth of block-like structures (...) + 1" and how this is related to the function-level block case. Or did you mean that in this case the depth is *not* a block-like structure?

proposals/exception-handling/Exceptions.md
```diff
@@ -311,6 +310,39 @@ catching instructions (`catch`/`catch_all`/`delegate`) come after the
...
+```
+try $l1
+  try
+    call $foo
+  delegate 1  ;; delegate to the caller
+catch
+  ...
+catch_all
+  ...
+end
```
thibaudmichaud: To make the example complete I think it should be wrapped in a function, otherwise it is ambiguous what the outer scope is.

proposals/exception-handling/Exceptions.md
thibaudmichaud: I understand what you meant now. What do you think about this phrasing: "the depth of the outermost block + 1, i.e. the function-level block (...)"

proposals/exception-handling/Exceptions.md
aheejin: Done

proposals/exception-handling/Exceptions.md
aheejin: Done. Also added that it will be considered as a catchless try.

aheejin:
Closing this because we have decided to change the spec for `delegate` in #176.

Issue #146, opened 2021-01-26
Match delegate's immediate and label argument to those of branches


aheejin:
I'd like to suggest a slight change to the current seemingly discussed spec in #143 so that we can treat the target of `br`s and `delegate`s in a similar way in the wast format. I think this will help reading and writing the wast format and spec tests.

I'm not actually very sure if this will be a suggestion to  _change_, because after discussions w/ @thibaudmichaud I realized my understanding was different from his and what I'd like to suggest here was actually already more in line with his V8
implementation. (Now I think we two are on the same page.)

So what I'd like to achieve is we can write the argument of `delegate` as labels in the wast format, as we do for `br`/`br_if`s, and make `delegate` can syntactically target a `try`, which semantically in turn goes to that `try`'s corresponding `catch`. `delegate`'s immediate argument is translated to the labels in the same way we do for `br`s. Let me illustrate that with an example. I annotated `catch`es with their corresponding `try`'s labels just to make it more readable. The example looks a bit long, but please bear with me:
```wasm
try $labelA
catch ($labelA)
  block $labelB
    try $labelC
      br 0         (= br $labelC)
      br 1         (= br $labelB)
      br 2         (= br $labelA)
      try
      delegate 0   (= delegate $labelC, which delegates to 'catch ($labelC)')
      try
      delegate 1   (= delegate $labelB, which is a block, so validation failure)
      try
      delegate 2   (= delegate $labelA, which delegates to 'catch ($labelA)', but 'catch ($labelA)' is above this instruction, so validation failure)
      try
      delegate 3   (= delegate to the caller)
    catch ($labelC)
      br 0         (= br $labelC)
      br 1         (= br $labelB)
      br 2         (= br $labelA)
      try
      delegate 0   (= delegate $labelC, which delegates to 'catch ($labelC)', but 'catch ($labelC)' is above this instruction, so validation failure)
      try
      delegate 1   (= delegate $labelB, which is a block, so validation failure)
      try
      delegate 2   (= delegate $labelA, which delegates to 'catch ($labelA)', but 'catch ($labelA)' is above this instruction, so validation failure)
      try
      delegate 3   (= delegate to the caller)
    end  ;; try $labelC
  end ;; block $labelB
end ;; try $labelA
```

(catches can be either `catch` or `catch_all`, it doesn't really matter here. `unwind` will be treated the same.)

[...]

PR #178, opened 2021-09-07
Reflect spec changes in delegate immediate operand

### LINKED ISSUE #176: Inconsistency in documentation for try-delegate behavior for function body

aheejin:
This adds the changes decided in #176 and adds a modified version from
https://github.com/WebAssembly/exception-handling/issues/146#issue-794482909
for clarification. (The example is not the same by the way; the `catch`
from the outermost `try` has been moved)

Closes #176.


proposals/exception-handling/Exceptions.md
```diff
@@ -288,28 +288,104 @@ end
...
+  end
+catch ($l0)
+end
+```
 
-Note that the example below is a validation failure:
+`delegate` can only target `catch`/`catch_all`s that are below the current
+instruction, i.e., it can only delegate downwards. It also targets all
+`catch`/`catch_all`s of a `try` as a whole and does not target individual
+`catch`/`catch_all` within a `try`. Here is another example:
```
rossberg: Nit: Delegate is no longer targeting catch after this change, it simply targets a block. I would perhaps rephrase this accordingly and say that all properties of Wasm's structured control flow labels apply, i.e., you can only target labels of outer blocks.

proposals/exception-handling/Exceptions.md
```diff
@@ -288,28 +288,104 @@ end
...
+
+Here the `delegate` is targeting `catch ($l1)`, which exists before the
+`delegate`. So in case an exception occurs, it propagates out and ends up
+targetting `catch ($l0)`, if the catch has a matching tag. If not, it will
+propagate further out. Even if the `catch_all` is below the `delegate`,
+`delegate` targets catches of a `try` as a whole and does not target an
+individual `catch`/`catch_all`, so it doesn't apply.
+
+If `delegate`'s immediate argument is (the depth of the outermost block + 1), it
+is considered to target the function-level block, so it in effect delegates the
```
```suggestion
If `delegate` targets the implicit function body block, then in effect it delegates the
```

rossberg (approved):
Thanks, LGTM, merely some nits.

proposals/exception-handling/Exceptions.md
```diff
@@ -288,28 +288,104 @@ end
...
 and because it delegates exception handling to catching instructions associated
 with `$l1`, it will be next checked by the outer `catch` and then `catch_all`
-instructions. When the specified label within a `delegate` instruction does not
-correspond to a `try` instruction, it is a validation failure.
+instructions.
+
+`delegate` can also target `catch`-less `try`s or non-`try` block constructs
+like `block`s or `loop`s, in which case the delegated exception is assumed to
+propagate to the outer scope and will be caught by the next matching
+try-catches. In the examples, catches are annotated with `($label_name)` to
```
thibaudmichaud: I haven't seen this case mentioned explicitly yet, I think it fits here.
```suggestion
propagate to the outer scope and will be caught by the next matching
try-catches, or rethrown to the caller if there is no outer try block. In the examples, catches are annotated with `($label_name)` to
```

proposals/exception-handling/Exceptions.md
aheejin: True, but the example below is, the `delegate` targets `$l1`, which is outside of the instruction because the label is attached to the `try`, but because it is below `catch` so it ends up not targeting `$l1` but rather `$l0`, and this was what I tried to explain. Do you think simply describing that `delegate` target labels of outer blocks can explain this case?

proposals/exception-handling/Exceptions.md
```diff
@@ -274,10 +274,10 @@ try label and delegates exception handling to a `catch`/`catch_all`/`delegate`
 specified by the `try` label. For example, consider this code:
 
 ```
-try $l1
+try $l0
   try
     call $foo
-  delegate $l1  ;; (= delegate 0)
+  delegate $l0 (= delegate 0)
```
takikawa: This is a minor nit, but I liked this better with the comment `;;` (for consistency with the rest of the doc).

proposals/exception-handling/Exceptions.md
```diff
@@ -288,28 +288,103 @@ end
...
+  catch
+    ...
+  catch_all
+    ...
+  end
+)
+```
+In case `foo` throws, `delegate 1` here delegates the exception handling to the
+caller, i.e., the exception escapes the current function. If the immediate is
+greater than the depth of the outermost block + 1, it is a validation failure.
```
takikawa: Should this be "greater than or equal to"? (or "depth of the outermost block" without +1?)

takikawa (approved):

proposals/exception-handling/Exceptions.md
aheejin: Done

proposals/exception-handling/Exceptions.md
aheejin: Hmm, yeah, "the depth of the outermost block" is confusing. I tried to rephrase it to
> If the immediate is greater than the number of block nesting including the implicit function-level block, it is a validation failure.

Do you think it's better?

takikawa: I think there's still some potential for confusion if someone interprets "number of block nesting" as counting from 1 and not 0. (In the example right above here you might count nestings of `try` and function body as 2, but immediates >=2 are not allowed here) Perhaps it could explicitly say it's counting from zero?

aheejin: Sorry, yeah, it should change to "greater than equal to". Intuitively the level of nesting is a number > 0 when there's any, and that's what I meant too. I changed it to
> If the immediate is greater than or equal to the number of block nesting including the implicit function-level block, it is a validation failure. In this example, any number equal to or greater than 2 is now allowed.

aheejin: Oh... now allowed -> not allowed...

proposals/exception-handling/Exceptions.md
aheejin: @rossberg Do you still think this needs fixing?

rossberg: I would say s.th like this:

"""
Like branches, `delegate` can only target outer blocks, and effectively rethrows the exception in that block. Consequently, delegating to a specific `catch` or `catch_all` handler requires targeting the respective label from within the associated `try` block. Delegating to a label from within a `catch` block does delegate the exception to the next enclosing handler -- analogous to performing a `throw` within a `catch` block, that handler is no longer active at that point.
"""

aheejin: Done

aheejin:
@takikawa I'm not very familiar with what the status of newly added CI is, but is it expected to fail? This PR doesn't modify anything with the interpreter or the spec tests though.

takikawa:
@aheejin Yes, it should be ok that these two things (ref interpreter & core spec) are failing, and it should be ok to just merge. Sorry for the confusion.

The issue with the interpreter CI is that the interpreter changes haven't been merged yet so the exception tests don't pass (BTW, did you or anyone else want to look at that PR more or should I merge it as it has one approval already?), but it should be ok after it's merged.

The core spec fails most likely because it's out of sync with the upstream spec. This should be fixed once it's rebased.

PR #181, opened 2021-09-09
Cherry-pick CI changes from upstream spec repo

takikawa:
This PR cherry-picks two CI related commits from https://github.com/WebAssembly/spec in order to enable CI for pull requests and automatic rendering of the spec documents. It's a cherry-pick rather than a full rebase as there are a lot of merge conflicts at the moment (which might be better resolved as part of updating the formal spec).

There's also a commit to adjust the CI badge for this repo and add a link in the README to the rendered spec.


Ms2ger (approved):
It's unfortunate that a full merge is hard at this point, but we should at least have CI working. Thanks.

PR #182, opened 2021-09-15
Cherry-pick commits from upstream spec to make CI pass

takikawa:
This PR cherry-picks a minimal set of commits from the upstream spec repo that should make the CI core spec build pass.


aheejin (approved):
I'm not very familiar with how Sphinx works, but if it passes I think it should be fine. Thanks!

PR #186, opened 2021-10-28
Merge upstream spec tests

takikawa:
This PR merges the latest tests from https://github.com/WebAssembly/spec so that it will be possible to merge this proposal's tests into https://github.com/WebAssembly/testsuite/

(a full rebase would be fine too but that's blocked on the formal document where there are merge conflicts)


aheejin:
Thanks! But isn't the recommended way of syncing upstream is to pick commits directly from the spec repo? https://github.com/WebAssembly/proposals/blob/master/howto.md#syncing-with-upstream-changes

takikawa:
Thanks that's a good point, I'll change this to cherry-picking the appropriate test commits.

takikawa:
After looking into this a bit, it's difficult to cherry-pick the commits for this as the relevant commits have a bunch of other things too (like edits to the spec document). It might be better to wait on the more general merge that will be done with the formal spec update instead of trying to update the tests now.

I'm going to close this for now (also it turns out there are test failures too as we have to merge some interpreter changes to make this work).

sbc100:
> Thanks! But isn't the recommended way of syncing upstream is to pick commits directly from the spec repo? https://github.com/WebAssembly/proposals/blob/master/howto.md#syncing-with-upstream-changes

Hmm.. I don't see any thing about cherry-picking there.. its says to merge changes from upstream using `git merge`...

aheejin:
Yeah it doesn't tell us to cherry-pick, but I think it tells to merge commits as they are, rather than squashing them and creating new commits..? I don't remember details but @takikawa once tried to merge all pending commits from spec in another PR to make the CI work but couldn't do it because of some problem, so he had to cherrypick. So I thought if we can't sync all the commits here (if we can, that's better), at least cherry-picking some commits would let us sync remaining commits later. Do you think that will be a problem?

sbc100:
Its probably worth doing to the full merge and deal with the conflicts now.. since it will only get harder the longer you delay.  But its up to you obviously.

rossberg:
Agreed with @sbc100, cherry-picking now will likely just create a bigger mess later. Better to try keeping up with upstream.

Issue #189, opened 2021-11-08
Preserve identity of exceptions from JS when passing through wasm


Ms2ger:
Some discussion starting at https://github.com/WebAssembly/exception-handling/issues/183#issuecomment-955090191 already.

Proposal: attach an optional `externref` to the core exception, and make sure it's propagated by `catch`; `rethrow`, but not expose it to wasm code. Then hosts that don't need it, should be able to avoid storing it at all. The JS API spec can then extract that value when it gets an exception from wasm, or create a new one if there isn't one.

PR #197, opened 2022-02-04
Update explainer on stack trace support in JS API

aheejin:
This updates the explainer on
- We add stack trace option to `Exception` class in the JS API
- `Exception` can contain an optional `externref` to carry a stack trace
- `Exception` is not a subclass of JS `Error`

The WebIDL syntax is suggested by @eqrion in
https://github.com/WebAssembly/exception-handling/issues/183#issuecomment-998899547
and
https://github.com/WebAssembly/exception-handling/issues/183#issuecomment-1003178801.

Addresses #183 and #189.


aheejin:
@eqrion Thanks for the suggestion! It looks I can't add you to reviewers, so cc'ing you here.

eqrion:
Thank you, this looks good to me!

aheejin:
Gentle ping :)

Ms2ger (approved):

aheejin:
@Ms2ger I have a request :) Can you let me merge my own PRs going forward? Especially when there are multiple reviewers, sometimes I'd like to wait a little more. I also generally prefer letting authors merge PRs themselves in case they have the right access to the repo, which is what we do in our wasm toolchain repos (emscripten and binaryen)

Ms2ger:
Apologies, I'll try to remember that

proposals/exception-handling/Exceptions.md
```diff
@@ -455,27 +448,51 @@ access to the data fields of a `Exception` if a matching tag is given. This last
...
+`Error` and it can be used to represent normal control flow constructs,
+`traceStack` field can be set when we use it to represent errors. The format of
+stack trace strings conform to the [WebAssembly stack trace
+conventions](https://webassembly.github.io/spec/web-api/index.html#conventions).
+When `ExceptionOption` is not provided or it does not contain `traceStack`
+entry, `traceStack` is considered `false` by default.
+
+To preserve stack trace info when crossing the JS to Wasm boundary, `Exception`
+can internally contain an optional `externref` value containing a stack trace
+string, which is propagated when caught by `catch` and rethrown by `rethrow`.
```
rossberg, line 464: Hm, is this normative text or is it just an implementation note? IIUC, "internally" means that the reference is not observable in Wasm. Does it even matter then whether it is an externref?

Ms2ger: I suspect it is less error-prone to acknowledge the extra data that needs to be preserved in the core spec, rather than monkey-patching it from the outside. I'm not sure if that means we need to spell out that it's an externref, though.

rossberg: @Ms2ger, I'm not sure how we could do that without actually devolving into spec'ing how JS stack traces work – presumably, the trace also needs to be extended in certain situations. Giving the handwavy-ness of stack traces in JS itself, we can only get that wrong. I'd rather not go there.

aheejin: @rossberg Do you not prefer taking the optional stack trace on/off option in `WebAssembly.Exception` constructor, or you are OK with that part but doesn't want to say anything about how it is achieved and we delegate that entirely to toolchains?

rossberg: @aheejin, the latter. Or at least I would keep the "how" slightly more abstract.
```suggestion
To preserve stack trace info when crossing the JS to Wasm boundary, exceptions
can internally contain a stack trace, which is propagated when caught by `catch` and rethrown by `rethrow`.
```

aheejin: Done in #203.

PR #199, opened 2022-02-10
Fix links in the header for js and web api specs.

Ms2ger:


document/js-api/index.bs
```diff
@@ -1,11 +1,10 @@
 <pre class='metadata'>
-Title: WebAssembly JavaScript Interface
+Title: WebAssembly JavaScript Interface: Exception Handling
 Shortname: wasm-js-api
 Group: wasm
 Status: ED
 Level: 1
-TR: https://www.w3.org/TR/wasm-js-api-1/
```
aheejin, line 7: Are these `TR`s not necessary anymore?

aheejin (approved):

document/js-api/index.bs
Ms2ger: Can go either way for the fork, I think. Lemme know what you prefer.

aheejin: If either way is fine maybe let's keep it to be consistent with the main spec?

PR #200, opened 2022-02-10
[js-api] Few small fixes

aheejin:
These are few small fixes or typos. Some fixes can be wrong if I'm
miunderstanding something, in which case please let me know!

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -1107,7 +1107,7 @@ Note: Exported Functions do not have a \[[Construct]] method and thus it is not 
                 1. Let |payload| be |v|.\[[Payload]].
             1. Otherwise,
                 1. Let |type| be the [=JavaScript exception tag=].
-                1. Let |payload| be [=ToWebAssemblyValue=](|v|.\[[Payload]], [=externref=]).
+                1. Let |payload| be [=ToWebAssemblyValue=](|v|, [=externref=]).
             1. [=WebAssembly/Throw=] with |type| and |payload|.
         1. Otherwise, return |result|.\[[Value]].
     1. Let |store| be the [=surrounding agent=]'s [=associated store=].
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -443,7 +443,8 @@ The verification of WebAssembly type requirements is deferred to the
             1. Let [=external value|table=] |tableaddr| be |externval|.
             1. Let |table| be [=create a Table object|a new Table object=] created from |tableaddr|.
             1. Let |value| be |table|.
-        1. If |externtype| is of the form [=externtype/tag=] |attribute| <var ignore>tagtype</var>,
+        1. If |externtype| is of the form [=externtype/tag=] |attribute| <var ignore>functype</var>,
+            1. Assert: |attribute| is [=tagtype/attribute/exception=].
             1. Assert: |externval| is of the form [=external value/tag=] |tagaddr|.
             1. Let [=external value/tag=] |tagaddr| be |externval|.
             1. Let |tag| be [=create a Tag object|a new Tag object=] created from |tagaddr|.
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -443,12 +443,11 @@ The verification of WebAssembly type requirements is deferred to the
             1. Let [=external value|table=] |tableaddr| be |externval|.
             1. Let |table| be [=create a Table object|a new Table object=] created from |tableaddr|.
             1. Let |value| be |table|.
-        1. If |externtype| is of the form [=externtype/tag=] |attribute| <var ignore>functype</var>,
-            1. Assert: |attribute| is [=tagtype/attribute/exception=].
```
Ms2ger, line 447: Why remove this? (Do we still have the attribute thing? If no, needs some more changes.)

document/js-api/index.bs
```diff
@@ -1101,12 +1100,12 @@ Note: Exported Functions do not have a \[[Construct]] method and thus it is not
         1. [=Clean up after running script=] with |relevant settings|.
         1. Assert: |result|.\[[Type]] is <emu-const>throw</emu-const> or <emu-const>normal</emu-const>.
         1. If |result|.\[[Type]] is <emu-const>throw</emu-const>, then:
-            1. If |v| [=implements=] {{Exception}},
-                1. Let |type| be |v|.\[[Type]].
-                1. Let |payload| be |v|.\[[Payload]].
+            1. If |result| [=implements=] {{Exception}},
```
Ms2ger: *result* is a completion record, so can't be an object. I think a "Let *v* be *result*.[[Value]]" is missing.

Ms2ger (changes requested):
Sorry for the delay.

document/js-api/index.bs
aheejin: Oh we do still have the attribute thing, but I was just wondering if "attribute tagtype" made sense. Didn't really mean to remove the assertion though. How can we fix this?
```
        1. If |externtype| is of the form [=externtype/tag=] <var ignore>tagtype</var>,
            1. Assert: |externtype|.|attribute| is [=tagtype/attribute/exception=].
```
Would this be better? Does `|externtype|.|attribute|` make sense?

document/js-api/index.bs
aheejin: Isn't that just "Let _v_ be _result_"? Below we are using `v.[[Type]]` and `v.[[Payload]]`. I don't think `result.[[Value]].[[Type]]` makes sense..? Then if v == result, do we need another variable `v`?

Ms2ger: Why not? *result* is a completion record, which has a [[Value]] field. The [[Value]] field of *result* contains a WebAssembly.Exception object, which has [[Type]] and [[Payload]] fields.

document/js-api/index.bs
Ms2ger: Not really - the existing prose is the exact same pattern as the

> If *externtype* is of the form global *mut* *globaltype*,

step above - binding two new variables with the values of the *externtype* variant. This is what Rust would express as
```rust
match externtype {
  Func(_functype) => ...,
  Global(_mut, _globaltype_) => ...,
  Tag(attribute, _tagtype) => ...,
}
```

aheejin: I see, thanks. I'll revert this.

document/js-api/index.bs
aheejin: I thought this meant `[[Type]]` and `[[Payload]]` exist within [the completion record](https://tc39.es/ecma262/#sec-completion-record-specification-type), because it looks this line says so: https://github.com/WebAssembly/exception-handling/blob/ca2b7155837c136ce8b3c95c7e49afdd60974016/document/js-api/index.bs#L1043
It looks the completion record is extended to contain a tag and a payload. 

But as you said, `WebAssembly.Exception` also has `[[Type]]` and `[[Payload]]` field. Do these two fields exist within both entities, `WebAssembly.Exception` and the completion record?

Ms2ger: I don't follow. `func_invoke` is a wasm algorithm, so returns wasm values, not completion records.

aheejin: OK, I reverted this too.

Ms2ger (approved):
Thanks!

aheejin:
Thanks! This PR didn't matter, but please let me merge my PRs next time! :)

Ms2ger:
:facepalm: Thanks for reminding me!

PR #204, opened 2022-02-16
Small fixes on formal spec overview/examples

aheejin:
- Removes a link next to `[1]`. The link is also in the footnote so it's
  not strictly necessary, and adding a link right next to `[1]` makes
  `[]` disappear, so it appears as just `1`, which isn't probably what
  the author intended.
- Both of `throw/catch x` and `throw/catch $x` are being used. Unified
  them into `throw/catch $x`.
- Fixed a location of `$label2` in a test.
- Removes `_m` from `catch` and `caught` administrative instructions in
  Exceptions-formal-examples.md. They were only removed in the overview.
- Removes trailing whitespaces

### DIFF
--- a/proposals/exception-handling/Exceptions-formal-examples.md
+++ b/proposals/exception-handling/Exceptions-formal-examples.md
@@ -20,7 +20,7 @@ Note that the block contexts and throw contexts given for the reductions are the
 
 The only example with an almost full reduction trace, and all new instructions. Such explicit reduction steps are only shown in Example 4 and Example 5, to highlight the reduction step of the administrative `delegate`.
 
-In the following reduction, we don't show the first 4 steps, which just reduce the several `try`s and the `throw x` to their respective administrative instructions. The type of the tag `$x` here is `[]→[]`.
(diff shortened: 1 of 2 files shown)


proposals/exception-handling/Exceptions-formal-examples.md
```diff
@@ -153,11 +153,11 @@ Use the trivial throw context `T` again, this time to match the throw to the `ca
 Location of a rethrown exception. Let `x, y, z` be tag indices of tags with type `[t_x]→[]`, `[t_y]→[]`, and `[t_z]→[]` respectively. Let `val_p : t_p` for every `p` among `x, y, z`.
 
 ```
-try
+try $label2
```
aheejin, line 156: I think`$label2` should be here, given that it points to this `try`'s corresponding `catch`.. No?

ioannad (approved):
LGTM! Thank you, sorry I forgot to update this file!

aheejin:
Next time I'd appreciate if let me merge my own PRs! :) (Not a big deal for sure, but I don't usually use commit messages provided by Github that's the list of all commits within the PR.)

ioannad:
Apologies, will do so from now on.

PR #143, opened 2020-11-10
formal spec overview for the 3rd exception handling proposal

ioannad:
This is an attempt to formally describe @aheejin's 3rd proposal, which she presented to the Wasm CG, and which was voted to be the new EH proposal, on September 18, 2020. This is not formal spec that I have developed, but a formal description of this 3rd proposal.

This is a much reworked form of the [first attempt on this formal spec overview](https://github.com/WebAssembly/exception-handling/issues/87#issuecomment-702323595) edited accordingly. It is in the form of a document as @tlively [requested](https://github.com/WebAssembly/exception-handling/issues/142#issuecomment-724246117), to make discussion on specific points easier.

This formal spec overview is roughly in the style of the 2nd proposal's [formal spec overview](https://github.com/WebAssembly/exception-handling/issues/87#issuecomment-517216137) by @rossberg.

I would be happy if anyone could point out things that are wrong or that I misunderstood.

PR #205, opened 2022-02-25
Formal overview address latest review

ioannad:
This PR attempts to address @rossberg 's  last review on merged PR#143, and makes some extra changes for readability, and documentation.

@rossberg @aheejin 

Continuing here the [last discussion](https://github.com/WebAssembly/exception-handling/pull/143/files#r759907998) from the merged PR #143, because that PR is closed and difficult to follow.

The change to remove the +1 from the typing rule of the administrative delegate (will call it adm.delegate) was related to the decision to allow a plain `(try bt instr* delegate 0)` in the function body. It seems I forgot to update the typing rule for try-delegate itself to be in accord with that decision. 

So I think the correct rule for try-delegate should be the following (previously the 3rd premise was `C.labels[l] = [t'*]`, which was equivalent to `|C.labels| > l`).

```
C ⊢ bt : [t1*]→[t2*]
C, labels [t2*] ⊢ instr* : [t1*]→[t2*]
|C.labels| ≥ l
-------------------------------------------
C ⊢ try bt instr* delegate l : [t1*]→[t2*]
```

This way, a context with an empty labels vector can validate this even for `l=0`.

If we now rewrite the typing rule of adm.delegate in the same style it becomes:

```
S;C, labels [t*] ⊢ instr* : []→[t*]
|C.labels| ≥ l
-----------------------------------------------------
S;C, labels [t*] ⊢ delegate{l} instr* end : []→[t*]
```

So perhaps there was no +1 missing from this rule, but a label too many in the typing rule of try.delegate?

**Question**
1. Do you think this new rule for `try-delegate` addresses your concerns with regard to type preservation?
2. Would it be better (and perhaps equivalent?) to write the typing rule for adm.delegate as follows?

```
S;C ⊢ instr* : []→[t*]
|C.labels| > l
-----------------------------------------------------
S;C ⊢ delegate{l} instr* end : []→[t*]
```

### DIFF
--- a/proposals/exception-handling/Exceptions-formal-overview.md
+++ b/proposals/exception-handling/Exceptions-formal-overview.md
@@ -83,10 +83,12 @@ C ⊢ try bt instr1* (catch x instr2*)* (catch_all instr3*)? end : [t1*]→[t2*]
 
 C ⊢ bt : [t1*]→[t2*]
 C, labels [t2*] ⊢ instr* : [t1*]→[t2*]
-|C.labels| ≥ l
+C.labels[l] = [t0*]
 -------------------------------------------
 C ⊢ try bt instr* delegate l : [t1*]→[t2*]
 ```
+
+Note that `try ... delegate 0` may appear without any explicitly surrounding block, in which case the label 0 refers to the label of the frame. Currently it is not allowed to refer to a label higher than the one of the frame.
 
 ## Execution (Reduction)
 
@@ -209,7 +211,7 @@ S;C, labels [t2*] ⊢ instr1* : []→[t2*]
 S;C, labels [t2*] ⊢ catch{a? instr2*}* instr1* end : []→[t2*]
 
 S;C, labels [t*] ⊢ instr* : []→[t*]
-|C.labels| ≥ l
+C.labels[l+1] = [t0*]
 ------------------------------------------------------
 S;C, labels [t*] ⊢ delegate{l} instr* end : []→[t*]
 
--- a/proposals/exception-handling/Exceptions-formal-overview.md
+++ b/proposals/exception-handling/Exceptions-formal-overview.md
@@ -83,7 +83,7 @@ C ⊢ try bt instr1* (catch x instr2*)* (catch_all instr3*)? end : [t1*]→[t2*]
 
 C ⊢ bt : [t1*]→[t2*]
 C, labels [t2*] ⊢ instr* : [t1*]→[t2*]
-C.labels[l] = [t*]
+|C.labels| ≥ l
 -------------------------------------------
 C ⊢ try bt instr* delegate l : [t1*]→[t2*]
 ```
@@ -209,7 +209,7 @@ S;C, labels [t2*] ⊢ instr1* : []→[t2*]
 S;C, labels [t2*] ⊢ catch{a? instr2*}* instr1* end : []→[t2*]
 
 S;C, labels [t*] ⊢ instr* : []→[t*]
-C.labels[l] = [t0*]
+|C.labels| ≥ l
 ------------------------------------------------------
 S;C, labels [t*] ⊢ delegate{l} instr* end : []→[t*]
 
--- a/proposals/exception-handling/Exceptions-formal-overview.md
+++ b/proposals/exception-handling/Exceptions-formal-overview.md
@@ -210,10 +210,10 @@ S;C, labels [t2*] ⊢ instr1* : []→[t2*]
 -----------------------------------------------------------
 S;C, labels [t2*] ⊢ catch{a? instr2*}* instr1* end : []→[t2*]
 
-S;C, labels [t*] ⊢ instr* : []→[t*]
+S;C ⊢ instr* : []→[t*]
 C.labels[l+1] = [t0*]
 ------------------------------------------------------
-S;C, labels [t*] ⊢ delegate{l} instr* end : []→[t*]
+S;C ⊢ delegate{l} instr* end : []→[t*]
 
 S ⊢ tag a : tag [t0*]→[]
 (val:t0)*


ioannad:
@aheejin  [This PR also makes minor readability and documentation changes.](https://github.com/WebAssembly/exception-handling/pull/143/files#r812476148)

ioannad:
@aheejin about your question in the [last comment of that discussion](https://github.com/WebAssembly/exception-handling/pull/143/files#r812476148):

> If we change the current rule
> 
> ```
> F; val^n (try bt instr* delegate l)
>   ↪ F; label_m{} (delegate{l} val^n instr* end) end
> ```
> 
> to
> 
> ```
> F; val^n (try bt instr* delegate l)
>   ↪ F; delegate(l) (label_m{} val^n instr* end) end
> ```
> 
> meaning, if we put the label created by the try-delegate itself within the scope of `adm.delegate`, can this confusion be helped?

I think the reason for putting the label outside of adm.delegate was so that it's similar to what happens in `try-catch`. I don't know if a different behaviour there would be a different source of confusion, but I don't mind either way.

rossberg:
Well, I don't know, I would very much prefer to avoid non-standard premises talking about the size or depending on the ordering of the context, which is a bit of an abstraction leak. Note that the condition `|C.labels| >= l` is technically equivalent to saying `C.labels[l] = t*` for some `t*`, so it should be fine to keep the latter – we could easily prove a lemma saying that if `C.labels[l+1] = t*` then C.labels[l] = t'*`, which would make your previous formulation work. Yet, either way, it is a bit too hacky for my taste, and I fear such an abstraction leak would get in the way of clean formal reasoning about the context for other purposes.

@aheejin is right that it would be natural and solve the problem to nest label and admin delegate the other way around. You are also right that the current reduction keeps it symmetric with catch. But that was itself a hack, so it may be better to keep that confined to catch. WDYT?

proposals/exception-handling/Exceptions-formal-overview.md
```diff
@@ -130,6 +130,8 @@ C^k ::= 'catch'{ tagaddr? instr* }* B^k 'end'
       | 'delegate'{ labelidx } B^k 'end'
 ```
 
+Note the `C` in `C^k` above stands for `control`, because the related administrative instructions are in some ways modelling [control frame opcodes](https://webassembly.github.io/spec/core/appendix/algorithm.html?highlight=control#data-structures) "on the stack".
```
aheejin:
```suggestion
Note the `C` in `C^k` above stands for `control`, because the related administrative instructions are in some ways modeling [control frame opcodes](https://webassembly.github.io/spec/core/appendix/algorithm.html?highlight=control#data-structures) "on the stack".
```

proposals/exception-handling/Exceptions-formal-overview.md
```diff
@@ -194,25 +196,26 @@ There is a subtle difference though. The instruction `br l` searches for the `l+
...
-S.tags[a].type = [t*]→[]
---------------------------------
+S ⊢ tag a : tag [t*]→[]
+-------------------------------
 S;C ⊢ throw a : [t1* t*]→[t2*]
 
-((S.tags[a].type = [t'*]→[])?
- S;C, labels catch [t*] ⊢ instr'* : [t'*?]→[t*])*
-S;C, labels [t*] ⊢ instr* : []→[t*]
+((S ⊢ tag a : tag [t*]→[])?
```
aheejin:
```suggestion
((S ⊢ tag a : tag [t1*]→[])?
```

aheejin (commented):
Sorry for the late reply.

It would be probably better to split out the other non-`delegate`-related parts into another PR and commit it, so this PR only concerns `delegate`..?

@ioannad  
> So I think the correct rule for try-delegate should be the following (previously the 3rd premise was `C.labels[l] = [t'*]`, which was equivalent to `|C.labels| > l`).

Does this account for the added `label` in `delegate`'s execution semantics, which @rossberg was concerned about?

@rossberg 
> Note that the condition `|C.labels| >= l` is technically equivalent to saying `C.labels[l] = t*` for some `t*`, so it should be fine to keep the latter – we could easily prove a lemma saying that if `C.labels[l+1] = t*` then C.labels[l] = t'*`, which would make your previous formulation work.

Can this handle the the case @ioannad described, in which `delegate N` delegates to the caller? For example,
```wasm
(func $test
  (try
    (do
      ...
    )
    (delegate 0)
  )
)
```
given that there's no label this `delegate` targets within this function? Or, do we implicitly push a label when a function is entered in the formal spec?

How do we handle `br` without a `block`? Is that a validation failure?
```wasm
(func $test
  (br 0)
)
```
Is this a validation failure or it just exits the function?

proposals/exception-handling/Exceptions-formal-overview.md
```diff
@@ -152,34 +154,34 @@ An absent tag address in a `catch` administrative instruction (i.e., `a? = ε`)
 ```
 F; throw x  ↪  F; throw a  (if F.module.tagaddrs[x]=a)
 
-caught{a val^n} B^l[rethrow l] end
-  ↪ caught{a val^n} B^l[val^n (throw a)] end
+caught{a val*} B^l[rethrow l] end
```
aheejin: In the formal spec, when do we use `^n` and when do we use `*`? Can they be used interchangeably? I see both notations in the current (standardized) formal spec and I'm not very familiar with the formal notation, so..

aheejin:
Also I wasn't able to follow every argument from https://github.com/WebAssembly/exception-handling/pull/143/files#r759907998, partly due to I wasn't able to understand [this rule](https://webassembly.github.io/spec/core/appendix/properties.html#xref-exec-runtime-syntax-label-mathsf-label-n-xref-syntax-instructions-syntax-instr-mathit-instr-0-ast-xref-syntax-instructions-syntax-instr-mathit-instr-ast-xref-syntax-instructions-syntax-instr-control-mathsf-end). 
```
S; C |- instr0*: [t1^n] -> [t2*]
S; C, labels [t1^n] |- instr* : [] -> [t2*]
-------------------------------------------------------
S; C |- label_n{instr0*} instr* end : [] -> [t2*]
```
Why is `instr*`'s result `t2*` and not `t1^n`? (And what'd the difference between the notation `^n` and `*`?)
It looks `instr*` should produce `t1^n`, because the label's result type is `t1^n`, and that's what `instr0*` consumes after the `instr*` is executed. Which part am I getting wrong?

rossberg:
@aheejin:
> Or, do we implicitly push a label when a function is entered in the formal spec?

Yes, a function creates an implicit outermost label.

proposals/exception-handling/Exceptions-formal-overview.md
```diff
@@ -36,14 +36,14 @@ mod ::= 'module' ... tag*
 
 ## Validation (Typing)
 
-#### Modification to Labels
+
```
aheejin:
```suggestion
```

aheejin:
For `delegate`, do we allow a label immediate greater than the function-level block? For example,
```wasm
(func $test
  try
  delegate 0 ;; throw to the caller
  try
  delegate N ;; For any N >= 1. Do we allow this?
)
```
In this case `delegate 0` is throwing up to the caller. Do we allow `N` equal to or greater than 1? @ioannad's equation seems to be assuming that we allow this and @rossberg's doesn't seem to. I've been somehow assuming we allow this, but I can't find any past discussions on this. I'm OK either way. For `br` we don't seem to allow `br` greater than 0 in that case:
```wasm
(func $test
  br 0 ;; allowed
  br N ;; N >= 1 is not allowed
)
```

rossberg:
@aheejin, I agree that delegate shouldn't allow unbound labels, but I don't think @ioannad's formulation does that, since it requires `l <= |C.labels|`.

aheejin:
> @aheejin, I agree that delegate shouldn't allow unbound labels, but I don't think @ioannad's formulation does that, since it requires `l <= |C.labels|`.

Ah right, I was confused. Thanks, then I think we should assume we don't allow unbound labels. (Maybe I was the only one who was assuming we were allowing this anyway)

aheejin:
@ioannad Gentle ping :) Are you planning to update this soon?

ioannad:
Apologies for the long silence @aheejin !
I am working on updating the formal overview and spec asap.

proposals/exception-handling/Exceptions-formal-overview.md
ioannad: IIUC we use `^n` when the `n` has a constraint in the rule it is mentioned, but I think in the formal spec document this is not always the case.

This change is to implement exactly this idea, so to only use the `n` if it has a particular constraint, as it does for example in lines 163-164.

I think that this change makes the formal overview a bit more concise, as I see the formal overview as a "cheatsheet" for Wasm exceptions.

ioannad:
@rossberg 

> Note that the condition `|C.labels| >= l` is technically equivalent to saying `C.labels[l] = t*` for some `t*`, so it should be fine to keep the latter – we could easily prove a lemma saying that if `C.labels[l+1] = t*` then C.labels[l] = t'*`, which would make your previous formulation work.

I don't understand why is "`|C.labels| >= l`" technically equivalent to "there exists `t*` such that `C.labels[l] = t*`". If `|C.labels| = l` then there is no `C.labels[l]`. Or does the equivalence somehow involve/assume @aheejin's formulation with the "delegate"-label after the delegate administrative instruction? 

> @aheejin is right that it would be natural and solve the problem to nest label and admin delegate the other way around. You are also right that the current reduction keeps it symmetric with catch. But that was itself a hack, so it may be better to keep that confined to catch. WDYT?

I haven't checked all the details yet, but now I think this should work fine.

rossberg:
@ioannad, oops, you are right, I missed the =. Rather, `|C.labels| >= l` would be equivalent to `C.labels[l-1] = t*`, which is a bit more nasty. Either way, I'd like to avoid going there. :)

ioannad:
Just rebased but I see something went wrong there, fixing...

Still todo:  haven't finished switching to @aheejin's [suggestion](https://github.com/WebAssembly/exception-handling/pull/143/files#r812476148) yet - checking the examples work as expected.

ioannad:
@aheejin, @rossberg,
About the new reduction rule for `try-delegate` (see [comment above](https://github.com/WebAssembly/exception-handling/pull/205#issuecomment-1055173912)), which puts the label after the administrative `delegate{l}`:

I tried to implement this in the interpreter and ran into some issues that I'm not sure how to resolve, although they are probably solvable.

In order to keep things confined and more easily searchable (so to not have a too long discussion in the PR or inline), I will make a new PR dedicated to that change, in which I will also include the changes I tried in the interpreter, and more details. 

To wrap up this PR, I reverted the changes involving the length of `C.labels` and instead I did what the PR title says, that is addressed ["the latest review"](https://github.com/WebAssembly/exception-handling/pull/143/files#r759907998), by adding the +1 to the `C.labels` premise of the typing rule for `delegate{l}`, as requested.

ioannad:
@rossberg, reading again [this comment](https://github.com/WebAssembly/exception-handling/pull/143/files#r811680384) of the original discussion in #143, I am thinking that your objection may be in using `C.labels[l]` instead of `(C labels [t*]).labels[l+1]` in that rule's premise.

Am I reading that comment correctly, so is it the case that you think that the typing rule for `delegate{l}` should be as follows?

```
S;C' ⊢ instr* : []→[t*]
C'.labels[ l+1 ] = [t0*]
------------------------------------------------------
S;C' ⊢ delegate{l} instr* end : []→[t*]
```

Or am I misunderstanding that comment?

aheejin:
> @rossberg, reading again [this comment](https://github.com/WebAssembly/exception-handling/pull/143/files#r811680384) of the original discussion in #143, I am thinking that your objection may be in using `C.labels[l]` instead of `(C labels [t*]).labels[l+1]` in that rule's premise.
> 
> Am I reading that comment correctly, so is it the case that you think that the typing rule for `delegate{l}` should be as follows?
> 
> ```
> S;C' ⊢ instr* : []→[t*]
> C'.labels[ l+1 ] = [t0*]
> ------------------------------------------------------
> S;C' ⊢ delegate{l} instr* end : []→[t*]
> ```
> 
> Or am I misunderstanding that comment?

I'm also wondering about this. Only changing `l` to `l+1` doesn't make sense to me, because as @ioannad described in https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions-formal-overview.md,
> Because `delegate{l}` is always wrapped in its own `label_n{} ... end` [2], with the same lookup as for `br l` we end up breaking inside the `l+1`th surrounding block, and throwing there. 

And what @rossberg said in https://github.com/WebAssembly/exception-handling/pull/143/files#r811680384 was:
> The problem I see is that (4) is not the premise you'll need when arriving at the adm.delegate rule. Instead you'll need
> 
> ```
> (4') C'[l'] = [t'*]
> ```
> 
> with `C' = C, labels [t^n]` and some `l'`, because that's the context you now enter the rule with. The only suitable `l'` that allows you to prove this from (4) is `l+1`, I believe.

So we should change the LHS to `C`? @rossberg

rossberg:
@ioannad, you are right, the first premise needs changing as well – the added label is wrong there, since that's added by the surrounding `label` construct, not by `delegate`.

aheejin:
@ioannad Can we make the change and merge it?

Ms2ger (approved):

PR #216, opened 2022-07-14
Clarify that stack traces are not required

### LINKED ISSUE #215: Ambiguous wording regarding the optionality of stack tracing attachments
https://github.com/WebAssembly/exception-handling/blob/74e7aa69fedac46255bf48e0b8600889270d6037/proposals/exception-handling/Exceptions.md#L452-L454

The use of "can" here makes it seem as if attaching a stack trace is optional, and possibly up to the whim of the host/engine? Can this be reworded to make it obvious as to which way it is intended to be interpreted?

ghost:
Fixes #215.

This should indicate that an engine is still conforming if it does not attach a stack trace, even if `traceStack` is requested by JavaScript code (by setting it to `true`).

### DIFF
--- a/proposals/exception-handling/Exceptions.md
+++ b/proposals/exception-handling/Exceptions.md
@@ -449,22 +449,18 @@ check ensures that without access to a WebAssembly module's exported exception
 tag, the associated data fields cannot be read.
 
 The `Exception` constructor can take an optional `ExceptionOptions` argument,
-which can optionally contain `traceStack` entry. When `traceStack` is `true`,
-JavaScript VMs are permitted to attach a stack trace string to `Exception.stack`
-field, as in JavaScript's `Error` class. `traceStack` serves as a request to the
-WebAssembly engine to attach a stack trace; it is not necessary to honour if 
-`true`,
-but `trace` may not be populated if `traceStack` is `false`. While `Exception` 
-is
-not a subclass of JavaScript's `Error` and it can be used to represent normal 
-control
-flow constructs, `traceStack` field can be set when we use it to represent 
-errors.
-The format of stack trace strings conform to the [WebAssembly stack trace 
-conventions
-](https://webassembly.github.io/spec/web-api/index.html#conventions). When
-`ExceptionOption` is not provided or it does not contain `traceStack` entry,
-`traceStack` is considered `false` by default.
+which can optionally contain `traceStack` entry. When `traceStack` is
+`true`, JavaScript VMs are permitted to attach a stack trace string to
+`Exception.stack` field, as in JavaScript's `Error` class. `traceStack`
+serves as a request to the WebAssembly engine to attach a stack trace; it
+is not necessary to honour if `true`, but `trace` may not be populated if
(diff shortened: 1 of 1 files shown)


proposals/exception-handling/Exceptions.md
```diff
@@ -450,14 +450,16 @@ tag, the associated data fields cannot be read.
 
 The `Exception` constructor can take an optional `ExceptionOptions` argument,
 which can optionally contain `traceStack` entry. When `traceStack` is `true`,
```
ghost: This is described as "optional," but the associated WebIDL doesn't use `optional boolean`, or `(boolean | undefined)`? This wording could be dropped.

proposals/exception-handling/Exceptions.md
```diff
@@ -450,14 +450,16 @@ tag, the associated data fields cannot be read.
...
-JavaScript's `Error` class. While `Exception` is not a subclass of JavaScript's
-`Error` and it can be used to represent normal control flow constructs,
-`traceStack` field can be set when we use it to represent errors. The format of
-stack trace strings conform to the [WebAssembly stack trace
-conventions](https://webassembly.github.io/spec/web-api/index.html#conventions).
-When `ExceptionOption` is not provided or it does not contain `traceStack`
-entry, `traceStack` is considered `false` by default.
+JavaScript VMs are permitted to attach a stack trace string to `Exception.stack`
+field, as in JavaScript's `Error` class. `traceStack` serves as a request to the
+WebAssembly engine to attach a stack trace; it is not necessary to honour if `true`,
```
aheejin: While it is a request because there can be engines that don't have the stack trace capability, I'm not sure if the current text is unclear, or if I prefer to explicitly say "it is not necessary to honour", which is something we certainly don't encourage.

I'm actually not very sure why you find the current text not clear.

ghost: > While it is a request because there can be engines that don't have the stack trace capability, I'm not sure if the current text is unclear, or if I prefer to explicitly say "it is not necessary to honour", which is something we certainly don't encourage.

Well, if anything, the ECMAScript spec makes no demands on engines creating/exposing stack traces, i.e. `Error` `stack` is non-standard, nor do other specs directly demand it, such as `DOMException`, so isn't it a bit weird that the Wasm WG mandates it?

> I'm actually not very sure why you find the current text not clear.

I'm hoping that it more explicitly explains the expected behaviour. Unless one is taking an extremely strict view of the wording, the current text doesn't make it immediately obvious as to how an engine is permitted to behave when given this parameter.

Although, of course, feel free to reject the changes if you don't believe that it is a net positive.

aheejin: Hmm, I see, I think that makes sense. By the way can you wrap the text to 80 cols? It's not a rule kept universally in wasm spec repo, but this file conforms to it, so I think it's better to be consistent.

ghost: Yes, the immediate commit should have wrapped it, can you confirm?

(Although for future reference, which tool do you use, as it may differ?)

aheejin (approved):
Thanks!

PR #217, opened 2022-07-27
[js-api] Specify maximum number of tags

### LINKED ISSUE #212: Implementation-defined limits for tags
The JS-API specifies implementation-defined limits on the structure of a module [1]. It looks like this proposal doesn't specify the maximum amount of tags that may be allowed.

SpiderMonkey allows `1000000` to match all the other top level fields. It looks like V8 does as well.

[1] https://webassembly.github.io/exception-handling/js-api/index.html#limits

aheejin:
It is implemented as 1000000 in both V8 and Firefox.
V8: https://github.com/v8/v8/blob/64aeabbc95576892b61d3e99e60ea74bbf734f80/src/wasm/wasm-limits.h#L36
Firefox: https://github.com/mozilla/gecko-dev/blob/7b9d23ece4835bf355e5195f30fef942d376a1c7/js/src/wasm/WasmConstants.h#L1087

Fixes #212.

cc @eqrion

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -1423,6 +1423,7 @@ In practice, an implementation may run out of resources for valid modules below 
 <li>The maximum number of imports declared in a module is 100000.</li>
 <li>The maximum number of exports declared in a module is 100000.</li>
 <li>The maximum number of globals defined in a module is 1000000.</li>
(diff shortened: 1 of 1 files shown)


thibaudmichaud (approved):
Looks good to me.

PR #180, opened 2021-09-07
Update formal spec to (partially) implement the 3rd EH proposal

ioannad:
Updates formal spec to (partially) implement the 3rd EH proposal.

In particular:

- exnref is removed from reference types
- REFEXNADDR is removed from administrative instructions
- exn and exception are renamed to tag (also when part of a word/macro)
- renamed EITYPE to TAGITYPE, ETYPE to TAGTYPE,
- changed variable names for tags from "et" to "tt" or "tagt", "iet" to "itagt"
- adjusted indexing where needed
- CATCHN is renamed to CATCHadm and its syntax and rules are adjusted
- added CAUGHTadm and DELEGATEadm, both with syntax and rules
- adjusted syntax and rules for TRY-CATCH, using \catchtype in binary and syntax, in order to describe TRY-CATCH compactly
- added syntax and rules for TRY-DELEGATE
- adjusted syntax and rules for RETHROW
- removed dependencies "bulk instructions" and "reference types" from Release
- adjusted/added validation example of control frame with opcode catch/catch_all resp.
- adjusted "Folded Instructions" for TRY-CATCH (didn't add folded TRY-DELEGATE)
- in several places added or adjusted prose, or added TODO items to add prose later

In particular there are still the following items marked as
TODO:

[] to add prose for typing rule of DELEGATEadm
[] to add prose for execution rules of TRY-CATCH, TRY-DELEGATE, and RETHROW
[] to finish prose for "Throwing an exception" (exec/instructions.rst)
[] to add EXCEPTION result value, signifying throwing an exception to the embedder
[] to adjust block contexts by adding CATCHadm, DELEGATEadm, and CAUGHTadm
[] to change labels to include the label type "catch" (in section "Contexts")
[] to add prose for validation rules of TRY-DELEGATE and RETHROW

PR #219, opened 2022-09-01
Modified the throw context example with concrete types but not concrete values.

ioannad:
Addresses @aheejin's request in a previous review comment:
https://github.com/WebAssembly/exception-handling/pull/180#discussion_r955477590

Originally I wrote this using
- `val_{i32} = (i32.const 1)`,
- `val_{f32} = (f32.const 2.0)`, and
- `val_{i64} = (i64.const 3)`,

but the example seemed then really long.
To keep the example relatively short I switched to the current version.

### DIFF
--- a/document/core/exec/runtime.rst
+++ b/document/core/exec/runtime.rst
@@ -732,23 +732,24 @@ Throw contexts allow matching the program context around a throw instruction up 
 .. note::
    For example, catching a simple :ref:`throw <exec-throw>` in a :ref:`try block <exec-try-catch>` would be as follows.
 
-   Assume that :math:`\expand_F(bt) = [t1^n] \to [t2^m]`, for some :math:`n > m` such that :math:`t1^n[(n-m):n] = t2^m`,
-   and that the tag address `a` of :math:`x` corresponds to the tag type :math:`[t2^m] \to []`.
+   Assume that :math:`\expand_F(bt) = [i32 f32 i64] \to [f32 i64]`,
+   and that the tag address `a` of :math:`x` corresponds to the tag type :math:`[f32 i64] \to []`.
+   Let :math:`\val_{i32}`, :math:`\val_{f32}`, and :math:`\val_{i64}` be values of type |i32|, |f32|, and |i64| respectively.
 
    .. math::
       \begin{array}{ll}
-      & \hspace{-5ex} S;~F;~\val^n~(\TRY~\X{bt}~(\THROW~x)~\CATCH~x~\RETURN~\END) \\
-      \stepto & S;~F;~\LABEL_m\{\} (\CATCHadm\{a~\RETURN\}~\val^n~(\THROW~x)~\END)~\END \\
+      & \hspace{-5ex} F;~\val_{i32}~\val_{f32}~\val_{i64}~(\TRY~\X{bt}~(\THROW~x)~\CATCH~x~\END) \\
+      \stepto & F;~\LABEL_2\{\} (\CATCHadm\{a~\epsilon}~\val_{i32}~\val_{f32}~\val_{i64}~(\THROW~x)~\END)~\END \\
       \end{array}
 
-   We denote :math:`\val^n = \val^{n-m} \val^m`.
    :ref:`Handling the thrown exception <exec-throwadm>` with tag address :math:`a` in the throw context
-   :math:`T=[val^{n-m}\_]`, with the exception handler :math:`H=\CATCHadm\{a~\RETURN\}` gives:
+   :math:`T=[val_{i32}\_]`, with the exception handler :math:`H=\CATCHadm\{a~\epsilon\}` gives:
 
    .. math::
       \begin{array}{lll}
-      \stepto & S;~F;~\LABEL_m\{\}~(\CAUGHTadm\{a~\val^m\}~\val^m~\RETURN~\END)~\END & \hspace{9ex}\ \\
-      \stepto & \val^m & \\
+      \stepto & F;~\LABEL_2\{\}~(\CAUGHTadm\{a~\val_{f32}~\val_{i64}\}~\val_{f32}~\val_{i64}~\END)~\END & \hspace{9ex}\ \\
+      \stepto & F;~\LABEL_2\{\}~\val_{f32}~\val_{i64}~\END & \hspace{9ex}\ \\
+      \stepto & \val_{f32}~\val_{i64} & \\
       \end{array}
 
 
--- a/document/core/exec/runtime.rst
+++ b/document/core/exec/runtime.rst
@@ -752,15 +752,12 @@ If no exception :ref:`handler that catches the exception <syntax-handler>` is fo
       \end{array}
 
 
-   When a throw of the form :math:`val^m (\THROWadm~a)` occurs, search for the maximal surrounding throw context :math:`T` is performed,
-   which means any other values, labels, frames, and |CAUGHTadm| instructions surrounding the throw :math:`val^m (\THROWadm~a)` are popped,
+   When a throw of the form :math:`\val^m (\THROWadm~a)` occurs, search for the maximal surrounding throw context :math:`T` is performed,
+   which means any other values, labels, frames, and |CAUGHTadm| instructions surrounding the throw :math:`\val^m (\THROWadm~a)` are popped,
    until a :ref:`handler <syntax-handler>` for the exception is found.
    Then a new |CAUGHTadm| instruction, containing the tag address :math:`a` and the values :math:`\val^m`, is pushed onto the stack.
 
-   If no exception :ref:`handler that catches the exception <syntax-handler>` is found, the computation :ref:`results <syntax-result>` in an uncaught exception result value, which contains the exception's entire throw context.
-
    In this particular case, the exception is caught by the exception handler :math:`H` and its values are returned.
-
 
 
 .. index:: ! configuration, ! thread, store, frame, instruction, module instruction
@@ -800,7 +797,7 @@ Finally, the following definition of *evaluation context* and associated structu
    \begin{array}{llll}
    \production{(evaluation contexts)} & E &::=&
      [\_] ~|~
-     val^\ast~E~\instr^\ast ~|~
+     \val^\ast~E~\instr^\ast ~|~
      \LABEL_n\{\instr^\ast\}~E~\END \\
    \end{array}
 
--- a/document/core/exec/runtime.rst
+++ b/document/core/exec/runtime.rst
@@ -730,10 +730,10 @@ Throw contexts allow matching the program context around a throw instruction up 
    |CAUGHTadm| blocks do not represent active handlers. Instead, they delimit the continuation of a handler that has already been selected. Their sole purpose is to record the exception that has been caught, such that |RETHROW| can access it inside such a block.
 
 .. note::
-   For example, catching a simple :ref:`throw <exec-throw>` in a :ref:`try block <exec-try-catch>` would be as follows.
+   For example, catching a simple :ref:`throw <syntax-throw>` in a :ref:`try block <syntax-try-catch>` would be as follows.
 
    Assume that :math:`\expand_F(bt) = [\I32~\F32~\I64] \to [\F32~\I64]`,
-   and that the tag address `a` of :math:`x` corresponds to the tag type :math:`[\F32~\I64] \to []`.
+   and that the tag address `a` of :math:`x` has tag type :math:`[\F32~\I64] \to []`.
    Let :math:`\val_{i32}`, :math:`\val_{f32}`, and :math:`\val_{i64}` be values of type |I32|, |F32|, and |I64| respectively.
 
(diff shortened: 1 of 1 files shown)


ioannad:
@aheejin is the example clear enough now or do you think it's better to use the concrete values, so `(i32.const 1)` instead of `val_{i32}` etc?

ioannad:
@rossberg, incidentally, this also fixes [your previous comment](https://github.com/WebAssembly/exception-handling/pull/180#discussion_r955923760) about using "We ...".

ioannad:
@rossberg incidentally, this also addresses your [previous comment](https://github.com/WebAssembly/exception-handling/pull/180#discussion_r955919370) about using "let ... be .. ".

document/core/exec/runtime.rst
```diff
@@ -804,7 +805,7 @@ Finally, the following definition of *evaluation context* and associated structu
    \begin{array}{llll}
    \production{(evaluation contexts)} & E &::=&
      [\_] ~|~
-     \val^\ast~E~\instr^\ast ~|~
+     val^\ast~E~\instr^\ast ~|~
```
aheejin: Why remove `\` here?

document/core/exec/runtime.rst
```diff
@@ -732,31 +732,32 @@ Throw contexts allow matching the program context around a throw instruction up
...
-   When a throw of the form :math:`val^m (\THROWadm~a)` occurs, we search for the maximal surrounding throw context :math:`T`,
-   which means we pop any other values, labels, frames, and |CAUGHTadm| instructions surrounding the throw :math:`val^m (\THROWadm~a)`,
+   When a throw of the form :math:`\val^m (\THROWadm~a)` occurs, we search for the maximal surrounding throw context :math:`T`,
+   which means we pop any other values, labels, frames, and |CAUGHTadm| instructions surrounding the throw :math:`\val^m (\THROWadm~a)`,
    until we find an exception handler (corresponding to a try block) that :ref:`handles the exception <syntax-handler>`.
-   We then append the values :math:`val^m:[t^m]` to the tag address :math:`a` into a new |CAUGHTadm| instruction which we push on the stack.
+   We then append the values :math:`\val^m:[t^m]` to the tag address :math:`a` into a new |CAUGHTadm| instruction which we push on the stack.
 
    In other words, when a throw occurs, normal execution halts and exceptional execution begins, until the throw
    is the continuation (i.e., in the place of a :math:`\_`) of a throw context in a catching try block.
```
aheejin: Not strictly related to this PR, but I'm not sure if I understand this sentence.
- What does a hole represent? (I know this is a basic question, but "block context" says it is "the next step of computation is taking place".. Is that also the case here? Which next step are we talking about?)
- This sentence sounds like we have two separate moments of "exception execution begins" and "(until) the throw is the continuation". I thought the exception execution begins when the `throw` is executed... Are they two different things?

aheejin (approved):
LGTM % comments. The concrete example is a lot easier to understand. Thanks!

document/core/exec/runtime.rst
aheejin: By the way this is a general question, so please don't consider this question as a blocker for landing this PR.

document/core/exec/runtime.rst
ioannad: I removed `\` from all the `var_{...}` in the example because there `var` is part of a variable name (could have used anything, like `x`, `y`, `z` ), as opposed to `\var`which is a  [non-terminal symbol](https://webassembly.github.io/spec/core/syntax/conventions.html#grammar-notation). However using the non-terminal symbols as variable names is widely used in the spec, so I guess it might be confusing to change in this example. Putting all the `\` back before merging.

document/core/exec/runtime.rst
ioannad: > This sentence sounds like we have two separate moments of "exception execution begins" and "(until) the throw is the continuation". I thought the exception execution begins when the throw is executed... Are they two different things?

@aheejin this is exactly what I meant, so by "exceptional execution" I meant the execution of the throw, so the popping of values and labels until the first catching try block is reached. I was only trying to describe the execution of a throw instruction, but I see now I overcomplicated it, making it seem as if there are more than one execution steps during "exceptional execution". 

About the hole, I think of the hole as the position of the stack pointer of the Wasm stack machine, but I see how mentioning the hole here can be confusing as well. Thank you for  pointing this out! 

Rephrasing these two lines to:

>    In other words, a thrown exception is caught when it's the continuation of a throw context in a catching try block, i.e., it's inside a catching try block, which is the innermost with respect to the throw. 

Is it clearer like this?

aheejin: > when it's the continuation of a throw context in a catching try block, 

What's 'it' here? It looks to be "a thrown exception" looking at the first part of the sentence, but I'm not sure what it means that an exception is a continuation of a throw context.

document/core/exec/runtime.rst
```diff
@@ -732,34 +732,35 @@ Throw contexts allow matching the program context around a throw instruction up
...
+   :math:`T=[\val_{i32}\_]`, with the exception handler :math:`H=\CATCHadm\{a~\epsilon\}` gives:
 
    .. math::
       \begin{array}{lll}
-      \stepto & S;~F;~\LABEL_m\{\}~(\CAUGHTadm\{a~\val^m\}~\val^m~\RETURN~\END)~\END & \hspace{9ex}\ \\
-      \stepto & \val^m & \\
+      \stepto & F;~\LABEL_2\{\}~(\CAUGHTadm\{a~\val_{f32}~\val_{i64}\}~\val_{f32}~\val_{i64}~\END)~\END & \hspace{9ex}\ \\
+      \stepto & F;~\LABEL_2\{\}~\val_{f32}~\val_{i64}~\END & \hspace{9ex}\ \\
+      \stepto & \val_{f32}~\val_{i64} & \\
       \end{array}
```
aheejin: How many newlines do we usually add between paragraphs? Here we have three... Is that normal?

document/core/exec/runtime.rst
```diff
@@ -732,34 +732,35 @@ Throw contexts allow matching the program context around a throw instruction up
 .. note::
    For example, catching a simple :ref:`throw <exec-throw>` in a :ref:`try block <exec-try-catch>` would be as follows.
```
```suggestion
   For example, catching a simple :ref:`throw <syntax-throw>` in a :ref:`try block <syntax-try-catch>` would be as follows.
```

document/core/exec/runtime.rst
```diff
@@ -732,34 +732,35 @@ Throw contexts allow matching the program context around a throw instruction up
...
 
-   Assume that :math:`\expand_F(bt) = [t1^n] \to [t2^m]`, for some :math:`n > m` such that :math:`t1^n[(n-m):n] = t2^m`,
-   and that the tag address `a` of :math:`x` corresponds to the tag type :math:`[t2^m] \to []`.
+   Assume that :math:`\expand_F(bt) = [\I32~\F32~\I64] \to [\F32~\I64]`,
+   and that the tag address `a` of :math:`x` corresponds to the tag type :math:`[\F32~\I64] \to []`.
```
```suggestion
   and that the tag address `a` of :math:`x` has tag type :math:`[\F32~\I64] \to []`.
```

document/core/exec/runtime.rst
```diff
@@ -732,34 +732,35 @@ Throw contexts allow matching the program context around a throw instruction up
...
+      \stepto & F;~\LABEL_2\{\}~(\CAUGHTadm\{a~\val_{f32}~\val_{i64}\}~\val_{f32}~\val_{i64}~\END)~\END & \hspace{9ex}\ \\
+      \stepto & F;~\LABEL_2\{\}~\val_{f32}~\val_{i64}~\END & \hspace{9ex}\ \\
+      \stepto & \val_{f32}~\val_{i64} & \\
       \end{array}
 
 
 
-   When a throw of the form :math:`val^m (\THROWadm~a)` occurs, we search for the maximal surrounding throw context :math:`T`,
-   which means we pop any other values, labels, frames, and |CAUGHTadm| instructions surrounding the throw :math:`val^m (\THROWadm~a)`,
+   When a throw of the form :math:`\val^m (\THROWadm~a)` occurs, we search for the maximal surrounding throw context :math:`T`,
```
rossberg: Avoid "we" :)

document/core/exec/runtime.rst
```diff
@@ -732,34 +732,35 @@ Throw contexts allow matching the program context around a throw instruction up
...
+   which means we pop any other values, labels, frames, and |CAUGHTadm| instructions surrounding the throw :math:`\val^m (\THROWadm~a)`,
    until we find an exception handler (corresponding to a try block) that :ref:`handles the exception <syntax-handler>`.
-   We then append the values :math:`val^m:[t^m]` to the tag address :math:`a` into a new |CAUGHTadm| instruction which we push on the stack.
+   We then append the values :math:`\val^m:[t^m]` to the tag address :math:`a` into a new |CAUGHTadm| instruction which we push on the stack.
 
-   In other words, when a throw occurs, normal execution halts and exceptional execution begins, until the throw
-   is the continuation (i.e., in the place of a :math:`\_`) of a throw context in a catching try block.
+   In other words, a thrown exception is caught when it's the continuation of a throw context in a catching try block,
```
rossberg: I'm afraid I don't understand this sentence. How can an exception be a continuation?

document/core/exec/runtime.rst
```diff
@@ -732,34 +732,35 @@ Throw contexts allow matching the program context around a throw instruction up
...
+   i.e., it's inside a catching try block, which is the innermost with respect to the throw.
```
```suggestion
   i.e., it is inside a catching try block, which is the innermost with respect to the throw.
```

document/core/exec/runtime.rst
rossberg: Writing "var" in TeX math mode is the same as "v a r", i.e., three variables. among other things, the macros ensure that it's correctly type-set as "\mathit{var}".

rossberg (changes requested):

document/core/exec/runtime.rst
ioannad: By "the thrown exception" I meant the instruction that throws the exception,  i.e., `val* (throw a)`.

I guess it's better to call it "the exception throw" instead, reformulating.

document/core/exec/runtime.rst
ioannad: Thanks for catching this typo!

document/core/exec/runtime.rst
ioannad: Oops! :see_no_evil: fixing.

document/core/exec/runtime.rst
ioannad: @aheejin I just realised this is all superseded by your PR #232 so leaving this for now.

document/core/exec/runtime.rst
ioannad: I just realised this part is superseded by @aheejin 's PR #232, so leaving this as is for now.

ioannad: This is fixed by @aheejin's PR #232 :)

document/core/exec/runtime.rst
ioannad: @aheejin had the [same remark](https://github.com/WebAssembly/exception-handling/pull/219#discussion_r972985641) and I think you are both right that this doesn't make sense. Rephrasing "the thrown exception" to "the exception throw". Is it better like this?

ioannad: (done)

document/core/exec/runtime.rst
ioannad: (done)

document/core/exec/runtime.rst
rossberg: I'm afraid I still don't understand the bit about the exception being the continuation of a context.

document/core/exec/runtime.rst
ioannad: Now that #232 landed, I merged the current main branch, in which this is fixed.

document/core/exec/runtime.rst
ioannad: To be honest, I really can't remember anymore why this is written like this in the first place (2nd EH proposal), and reading this again I agree it doesn't make much sense, so I removed this line in the last commit.

document/core/exec/runtime.rst
ioannad: Tihs suggestion is applied in the last commit.

document/core/exec/runtime.rst
ioannad: This is now entirely removed, after [this discussion](https://github.com/WebAssembly/exception-handling/pull/219#discussion_r992420960).

ioannad:
All review comments should be addressed now, pending for approval.

document/core/exec/runtime.rst
ioannad: This sentence mentioning continuations is now entirely removed.

rossberg (approved):

PR #221, opened 2022-09-06
Formal overview: add uncaught exceptions

ioannad:
Based on this previous discussion:
https://github.com/WebAssembly/exception-handling/pull/143/files#r761207853

Also:
- fixed throw contexts to include situations where `val* T instr*` in general.

### DIFF
--- a/proposals/exception-handling/Exceptions-formal-overview.md
+++ b/proposals/exception-handling/Exceptions-formal-overview.md
@@ -139,9 +139,10 @@ Throw contexts don't skip over handlers (administrative `catch` or `delegate` in
 Throw contexts are used to match a thrown exception with the innermost handler.
 
 ```
-T ::= val* '[_]' instr* | 'label'_n{instr*} T 'end'
-   | 'caught'{ tagaddr val^n } T 'end'
-   | 'frame'_n{F} T end
+T ::= '[_]' | val* T instr*
+   | 'label'_n{instr*} T 'end'
+   | 'caught'{ tagaddr val* } T 'end'
+   | 'frame'_n{F} T 'end'
 ```
 
 Note that because `catch` and `delegate` instructions are not included above, there is always a unique maximal throw context to match the reduction rules. Note that this basically means that `caught{..} instr* end` is not a potential catching block for exceptions thrown by `instr*`. The instruction sequence `instr*` is inside a `catch` or `catch_all` block.
@@ -222,6 +223,11 @@ S;C, labels (catch [t*]) ⊢ instr* : []→[t*]
 S;C, labels [t*] ⊢ caught{a val^n} instr* end : []→[t*]
 ```
 
-## TODO Uncaught Exceptions
+## Uncaught Exceptions
 
-We haven't yet described the formalism for an uncaught exception being the result of evaluation.
+To describe uncaught exceptions we add a new terminal value.
+
+```
+result ::= ... | T[val* (throw tagaddr)]
+```
+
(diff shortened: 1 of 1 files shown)


proposals/exception-handling/Exceptions-formal-overview.md
```diff
@@ -222,6 +223,11 @@ S;C, labels (catch [t*]) ⊢ instr* : []→[t*]
 S;C, labels [t*] ⊢ caught{a val^n} instr* end : []→[t*]
 ```
 
-## TODO Uncaught Exceptions
+## Uncaught Exceptions
+
+To describe uncaught exceptions we add a new terminal value.
```
aheejin: @rossberg suggested we should avoid 'we': https://github.com/WebAssembly/exception-handling/pull/180#discussion_r955923760

proposals/exception-handling/Exceptions-formal-overview.md
```diff
@@ -222,6 +223,11 @@ S;C, labels (catch [t*]) ⊢ instr* : []→[t*]
...
+
+```
+result ::= ... | T[val* (throw tagaddr)]
```
aheejin: What `result` are we talking about? Can we expand `...` if that's not too long?

proposals/exception-handling/Exceptions-formal-overview.md
ioannad: Thank you, it seems to have slipped my mind!

proposals/exception-handling/Exceptions-formal-overview.md
ioannad: This refers to [these result values](https://webassembly.github.io/spec/core/exec/runtime.html#syntax-result). I now expanded `...` and added this link in the document as well.

proposals/exception-handling/Exceptions-formal-overview.md
ioannad: Fixed in the latest commit. By the way I noticed other usages of "we" in the formal overview, will fix those, including some other nits in an upcoming PR.

proposals/exception-handling/Exceptions-formal-overview.md
```diff
@@ -222,6 +223,12 @@ S;C, labels (catch [t*]) ⊢ instr* : []→[t*]
...
 ```
 
-## TODO Uncaught Exceptions
+## Uncaught Exceptions
+
+A new [result](https://webassembly.github.io/spec/core/exec/runtime.html#syntax-result) value is added to describe uncaught exceptions.
+
+```
+result ::= val* | trap
+         | T[val* (throw tagaddr)]
```
aheejin, line 232: - What @rossberg suggested was `val* T[throw e] instr*`... Is this different from `T[val* (throw tagaddr)]`?
- How are the cases, which a throw context contains not only a `throw` and `val`s but other things... such as `label`, `caught`, or `frame` as listed in the throw context section, covered?

ioannad: > What @rossberg suggested was `val* T[throw e] instr*`... Is this different from `T[val* (throw tagaddr)]`?

They were different when @rossberg suggested the former formulation, but in this PR I also changed throw contexts from `T ::= val* '[_]' instr* | ...` to `T ::= '[_]' | val* T instr* | ...`, so now these two things `val* T[throw e] instr*` and `T[val* (throw tagaddr)]` are really the same. 

>  How are the cases, which a throw context contains not only a throw and vals but other things... such as label, caught, or frame as listed in the throw context section, covered?

So now throw contexts surrounded by values and instructions are also throw contexts. Meaning that `val* T instr*` is a throw context too.

rossberg (approved):

aheejin (approved):

PR #224, opened 2022-09-07
Add prose for reduction rule of try-catch-catch_all.

ioannad:

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2637,21 +2637,25 @@ Control Instructions
 
 6. Let :math:`F` be the :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>`.
 
-7. Let :math:`H` be the empty :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm`.
+7. For each catch clause :math:`(\CATCH~x_i~\instr_{2i}^\ast)` do:
 
-8. For each catch clause :math:`(\CATCH~x~\instr_2^\ast)` do:
+   a. Assert: due to :ref:`validation <valid-try-catch>`, :math:`F.\AMODULE.\MITAGS[x_i]` exists.
 
-   a. Let :math:`a_x` be the tag address :math:`F.\AMODULE.\MITAGS[x]`.
+   b. Let :math:`a_i` be the tag address :math:`F.\AMODULE.\MITAGS[x_i]`.
 
-   b. Append :math:`\{a_x~\instr_2^\ast\}` to :math:`H`.
+   c. Let :math:`H_i` be the handler clause :math:`\{a_i~\instr_{2i}^\ast\}`.
 
-9. If there is a catch all clause :math:`(\CATCHALL~\instr_3^\ast)`, then
+8. If there is a catch all clause :math:`(\CATCHALL~\instr_3^\ast)`, then:
 
-   a. Append :math:`\{\epsilon~\instr_3^\ast\}` to :math:`H`.
+    a. Let :math:`H'^?` be the handler clause :math:`\{\epsilon~\instr_3^\ast\}`.
 
-10. :ref:`Enter <exec-instr-seq-enter>` the block :math:`H~(\val^m~\instr_1^\ast)~\END` with label :math:`L`.
+9. Else:
 
-11. :ref:`Enter <exec-handler-enter>` :math:`\val^m~\instr_1^\ast` with exception handler :math:`H`.
+    a. Let :math:`H'^?` be the empty handler clause :math:`\epsilon`.
+
+10. Let :math:`H^\ast` be the :ref:`catching exception handler <syntax-handler>` containing the concatenation of the handler clauses :math:`H_i` and :math:`H'^?`.
+
+11. :ref:`Enter <exec-handler-enter>` the block :math:`\val^m~\instr_1^\ast` with label :math:`L` and exception handler :math:`H`.
 
 .. math::
    ~\\[-1ex]
@@ -2999,12 +3003,14 @@ The following auxiliary rules define the semantics of entering and exiting :ref:
 
 .. _exec-handler-enter:
 
-Entering :math:`\instr` with exception handler :math:`H`
-........................................................
+Entering :math:`\instr^\ast` with label :math:`L` and exception handler :math:`H`
+.................................................................................
 
-1. Push :math:`H` onto the stack.
+1. Push :math:`L` to the stack.
 
-2. Jump to the start of the instruction sequence :math:`\instr^\ast`.
+2. Push :math:`H` onto the stack.
+
+3. Jump to the start of the instruction sequence :math:`\instr^\ast`.
 
 
 .. note::
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2662,7 +2662,7 @@ Control Instructions
    \begin{array}{l}
    F; \val^m~(\TRY~\X{bt}~\instr_1^\ast~(\CATCH~x~\instr_2^\ast)^\ast~(\CATCHALL~\instr_3^\ast)^?~\END
    \quad \stepto \\
-   \qquad F; \LABEL_n\{\}~(\CATCHadm\{a_x~\instr_2^\ast\}^\ast\{\epsilon~\instr_3\ast\}^?~\val^m~\instr_1^\ast~\END)~\END \\
+   \qquad F; \LABEL_n\{\epsilon\}~(\CATCHadm\{a_x~\instr_2^\ast\}^\ast\{\epsilon~\instr_3\ast\}^?~\val^m~\instr_1^\ast~\END)~\END \\
    (\iff \expand_F(\X{bt}) = [t_1^m] \to [t_2^n] \land (F.\AMODULE.\MITAGS[x]=a_x)^\ast)
    \end{array}
 
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2635,11 +2635,14 @@ Control Instructions
 
 5. Pop the values :math:`\val^m` from the stack.
 
-6. Let :math:`H` be the :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm\{a_x~instr_2^\ast\}^\ast\{\epsilon~\instr_3^\ast\}^?`, whose mapping from tag addresses to branch targets corresponds to the |CATCH| and |CATCHALL| clauses, where :math:`a_x` is the tag address of each tag :math:`x`.
+6. Let :math:`H` be the :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm\{a_x~instr_2^\ast\}^\ast\{\epsilon~\instr_3^\ast\}^?`, whose mapping from tag addresses to catch bodies corresponds to the |CATCH| and |CATCHALL| clauses, where :math:`a_x` is the tag address of each tag :math:`x`.
+
+   .. todo::
+       Step 6 needs proper writing out into a loop over x*, which looks up each corresponding a_x in the store.
 
 7. :ref:`Enter <exec-instr-seq-enter>` the block :math:`H~(\val^m~\instr_1^\ast)~\END` with label :math:`L`.
 
-8. :ref:`Install <exec-handler-enter>` the exception handler `H` which contains `\val^m~\instr_1^\ast`.
+8. :ref:`Enter <exec-handler-enter>` the block :math:`\val^m~\instr_1^\ast` with exception handler :math:`H`.
 
 .. math::
    ~\\[-1ex]
@@ -2987,8 +2990,8 @@ The following auxiliary rules define the semantics of entering and exiting :ref:
 
 .. _exec-handler-enter:
 
-Installing an exception handler :math:`H` containing :math:`\instr^\ast`
-........................................................................
+Entering :math:`\instr` with exception handler :math:`H`
+........................................................
 
 1. Push :math:`H` onto the stack.
 
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2646,8 +2646,8 @@ Control Instructions
    \begin{array}{l}
    F; \val^m~(\TRY~\X{bt}~\instr_1^\ast~(\CATCH~x~\instr_2^\ast)^\ast~(\CATCHALL~\instr_3^\ast)^?~\END
    \quad \stepto \\
-   \qquad F; \LABEL_n\{\}~(\CATCHadm\{a~\instr_2^\ast\}^\ast\{\epsilon~\instr_3\ast\}^?~\val^m~\instr_1^\ast~\END)~\END \\
-   (\iff \expand_F(\X{bt}) = [t_1^m] \to [t_2^n] \land (F.\AMODULE.\MITAGS[x]=a)^\ast)
+   \qquad F; \LABEL_n\{\}~(\CATCHadm\{a_x~\instr_2^\ast\}^\ast\{\epsilon~\instr_3\ast\}^?~\val^m~\instr_1^\ast~\END)~\END \\
+   (\iff \expand_F(\X{bt}) = [t_1^m] \to [t_2^n] \land (F.\AMODULE.\MITAGS[x]=a_x)^\ast)
    \end{array}
 
 
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2987,10 +2987,13 @@ The following auxiliary rules define the semantics of entering and exiting :ref:
 
 .. _exec-handler-enter:
 
-Installing an exception handler :math:`H`
-.........................................
+Installing an exception handler :math:`H` containing :math:`\instr^\ast`
+........................................................................
 
 1. Push :math:`H` onto the stack.
+
+2. Jump to the start of the instruction sequence :math:`\instr^\ast`.
+
 
 .. note::
    No formal reduction rule is needed for installing an exception :ref:`handler <syntax-handler>`
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2635,14 +2635,23 @@ Control Instructions
 
 5. Pop the values :math:`\val^m` from the stack.
 
-6. Let :math:`H` be the :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm\{a_x~instr_2^\ast\}^\ast\{\epsilon~\instr_3^\ast\}^?`, whose mapping from tag addresses to catch bodies corresponds to the |CATCH| and |CATCHALL| clauses, where :math:`a_x` is the tag address of each tag :math:`x`.
+6. Let :math:`F` be the :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>`.
 
-   .. todo::
-       Step 6 needs proper writing out into a loop over x*, which looks up each corresponding a_x in the store.
+7. Let :math:`H` be the empty :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm`.
 
-7. :ref:`Enter <exec-instr-seq-enter>` the block :math:`H~(\val^m~\instr_1^\ast)~\END` with label :math:`L`.
+8. For each catch clause :math:`(\CATCH~x~\instr_2^\ast)` do:
 
-8. :ref:`Enter <exec-handler-enter>` :math:`\val^m~\instr_1^\ast` with exception handler :math:`H`.
+   a. Let :math:`a_x` be the tag address :math:`F.\AMODULE.\MITAGS[x]`.
+
+   b. Append :math:`\{a_x~\instr_2^\ast\}` to :math:`H`.
+
+9. If there is a catch all clause :math:`(\CATCHALL~\instr_3^\ast)`, then
+
+   a. Append :math:`\{\epsilon~\instr_3^\ast\}` to :math:`H`.
+
+10. :ref:`Enter <exec-instr-seq-enter>` the block :math:`H~(\val^m~\instr_1^\ast)~\END` with label :math:`L`.
+
+11. :ref:`Enter <exec-handler-enter>` :math:`\val^m~\instr_1^\ast` with exception handler :math:`H`.
 
 .. math::
    ~\\[-1ex]
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2999,7 +2999,7 @@ Entering :math:`\instr` with exception handler :math:`H`
 
 
 .. note::
-   No formal reduction rule is needed for installing an exception :ref:`handler <syntax-handler>`
+   No formal reduction rule is needed for entering an exception :ref:`handler <syntax-handler>`
    because it is an :ref:`administrative instruction <syntax-instr-admin>`
    that the :ref:`try <syntax-try>` instruction reduces to directly.
 
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2627,27 +2627,27 @@ Control Instructions
 
 1. Assert: due to :ref:`validation <valid-blocktype>`, :math:`\expand_F(\blocktype)` is defined.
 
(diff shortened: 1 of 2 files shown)


document/core/exec/instructions.rst
```diff
@@ -2625,8 +2625,21 @@ Control Instructions
 :math:`\TRY~\blocktype~\instr_1^\ast~(\CATCH~x~\instr_2^\ast)^\ast~(\CATCHALL~\instr_3^\ast)^?~\END`
 ....................................................................................................
 
-.. todo::
-   Add prose for the |TRY| - |CATCH| - |CATCHALL| execution step.
+1. Assert: due to :ref:`validation <valid-blocktype>`, :math:`\expand_F(\blocktype)` is defined.
+
+2. Let :math:`[t_1^n] \to [t_2^m]` be the :ref:`function type <syntax-functype>` :math:`\expand_F(\blocktype)`.
```
aheejin: Other block-like instructions (`block`, `loop`, and `if`) all use `t1^m -> t2^n` instead of `t1^n -> t2^m`, so I think it'd better be consistent. This also applies to the lines below, the preexisting equations, and the next `try`-`delegate` rule.

document/core/exec/instructions.rst
```diff
@@ -2625,8 +2625,21 @@ Control Instructions
...
+
+3. Assert: due to :ref:`validation <valid-try-catch>`, there are at least :math:`n` values on the top of the stack.
+
+4. Pop the values :math:`\val^n` from the stack.
+
+5. Let :math:`L` be the label whose arity is :math:`m` and whose continuation is the end of the |TRY| instruction.
```
aheejin: Other block-like instructions (`block`, `loop`, and `if`)'s first 5 steps are this, and we have all these steps here as well, but in a slightly different order. I think they better be the same for consistency?
```
1. Assert: due to :ref:`validation <valid-blocktype>`, :math:`\expand_F(\blocktype)` is defined.

2. Let :math:`[t_1^m] \to [t_2^n]` be the :ref:`function type <syntax-functype>` :math:`\expand_F(\blocktype)`.

3. Let :math:`L` be the label whose arity is :math:`n` and whose continuation is the end of the block.

4. Assert: due to :ref:`validation <valid-block>`, there are at least :math:`m` values on the top of the stack.

5. Pop the values :math:`\val^m` from the stack.
```

https://github.com/WebAssembly/exception-handling/blob/76732aeefb639347a253c4740d403b36cf7be6a6/document/core/exec/instructions.rst#L2537-2545

document/core/exec/instructions.rst
```diff
@@ -2625,8 +2625,21 @@ Control Instructions
...
+
+6. Let :math:`H` be the :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm\{a_x~instr_2^\ast\}^\ast\{\epsilon~\instr_3^\ast\}^?`, whose mapping from tag addresses to branch targets corresponds to the |CATCH| and |CATCHALL| clauses, where :math:`a_x` is the tag address of each tag :math:`x`.
```
aheejin: If we are going to use `a_x`, I think we should use it in the equations too. Currently the equations just use `a`.

---

It is a little confusing that we use two different `catch`es - one for the real `catch` instruction and the other for the administrative `catch` - within a single sentence. If we see this raw text it may not be obvious, but it's seen like this in the web page.
image
Should we prepend 'administrative instruction' to the first `CATCHadm`?

---

> whose mapping from tag addresses to branch targets corresponds to the |CATCH| and |CATCHALL| clauses,

What does this mean? What is the mapping from tag addresses to branch targets?

document/core/exec/instructions.rst
```diff
@@ -2625,8 +2625,21 @@ Control Instructions
...
+
+7. :ref:`Enter <exec-instr-seq-enter>` the block :math:`H~(\val^n~\instr_1^\ast)~\END` with label :math:`L`.
+
+8. :ref:`Enter <exec-handler-enter>` the exception handler `H`.
```
aheejin: Are these two different things? If so, how?

ioannad:
@aheejin Switched `n` with `m` and the ordering of the steps as suggested in your comments.

document/core/exec/instructions.rst
ioannad: @aheejin I'll change `a` to `a_x` in the equations as well, in the next commit. 

I agree that using the same word for the administrative `catch` is not very clear, but I seem to remember @rossberg thought it's clear from the context. FWIW the first `catch` is referred to as a [catching exception handler](https://webassembly.github.io/exception-handling/core/exec/runtime.html#exception-handlers) and the next `catch` as a clause of the `try-catch-catch_all` instruction. 

>  Should we prepend 'administrative instruction' to the first CATCHadm?

I meant to imply that the first `CATCHadm` is an administrative instruction by calling it a "catching exception handler". I'm not sure how to rephrase this to include "administrative instruction". Any suggestions?

> >  whose mapping from tag addresses to branch targets corresponds to the |CATCH| and |CATCHALL| clauses,
> What does this mean? What is the mapping from tag addresses to branch targets?

I was referring to the explanation given for [catching exception handlers](https://webassembly.github.io/exception-handling/core/exec/runtime.html#exception-handlers):

> Catching handlers start with the identifier and carry a mapping from tag addresses to their associated branch targets,

I'll try to think of a way to rephrase this, can't come up with something atm.

document/core/exec/instructions.rst
ioannad: I'm not sure I understand the question. Would it be clearer if there was a second step at section [exec-handler-enter](https://webassembly.github.io/exception-handling/core/exec/instructions.html#entering-an-exception-handler-h) as follows?

> 1. Push H onto the stack.
> 2. Jump to the start of the instruction sequence :math:`\instr^\ast`.

ioannad:
@aheejin I also switched `a` to `a_x` in the formal rule. For the other review comments see my replies, as I am not sure atm how to rephrase those parts.

document/core/exec/instructions.rst
aheejin: > > > whose mapping from tag addresses to branch targets corresponds to the |CATCH| and |CATCHALL| clauses,
> > > What does this mean? What is the mapping from tag addresses to branch targets?
> 
> I was referring to the explanation given for [catching exception handlers](https://webassembly.github.io/exception-handling/core/exec/runtime.html#exception-handlers):

What 'branch targets' are we talking about here? I think this is the part I was confused about. Do we have a branch target per catch clause?

document/core/exec/instructions.rst
aheejin: I think I understand 7, but I was confused by 8. What does "entering the exception handling `H`" mean? Does that mean the exception is caught? (If so, can we assume it is caught?)

document/core/exec/instructions.rst
ioannad: The term "branch" here refers to a CFG of a try-catch instruction. Such a CFG has a central [basic block](https://en.wikipedia.org/wiki/Basic_block) checking the exception's tag, and from this block there are branches to all the catch blocks of the try-catch instruction. Depending on the exception's tag, say it's `a`, the execution flow follows the branch to the block of `catch a`.

document/core/exec/instructions.rst
ioannad: I was in fact describing step 8, so the steps in my previous comment are the meaning of "entering the exception handler `H`", that is by putting the `H` (either a `catch[a? instr2*}*` or a `delegate{l}`) on the stack and starting to execute the instructions `instr*` of the try block. 

I now realised we also call this "installing an exception handler" (see [note here](https://webassembly.github.io/exception-handling/core/exec/instructions.html#entering-an-exception-handler-h)), which is perhaps a better description of what's happening, rather than "entering an exception handler". Switching to that.

ioannad: Done. @aheejin WDYT?

ioannad: Done. @aheejin WDYT?

document/core/exec/instructions.rst
aheejin: How about "catch bodies"? We don't use the term basic block in our formal spec, and we use the term "branch" only to mean `br`/`br_if`/`br_table` instructions, so I think it can be confusing.

document/core/exec/instructions.rst
```diff
@@ -2974,11 +2987,14 @@ The following auxiliary rules define the semantics of entering and exiting :ref:
 
 .. _exec-handler-enter:
 
-Entering an exception handler :math:`H`
-.......................................
+Installing an exception handler :math:`H` containing :math:`\instr^\ast`
```
rossberg: I think I'd keep calling it "entering", since it is a dynamic scope, and you exit it in the end.
```suggestion
Installing an exception handler :math:`H` for :math:`\instr^\ast`
```

document/core/exec/instructions.rst
```diff
@@ -2625,16 +2625,29 @@ Control Instructions
...
+
+4. Assert: due to :ref:`validation <valid-try-catch>`, there are at least :math:`m` values on the top of the stack.
+
+5. Pop the values :math:`\val^m` from the stack.
+
+6. Let :math:`H` be the :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm\{a_x~instr_2^\ast\}^\ast\{\epsilon~\instr_3^\ast\}^?`, whose mapping from tag addresses to branch targets corresponds to the |CATCH| and |CATCHALL| clauses, where :math:`a_x` is the tag address of each tag :math:`x`.
+
+7. :ref:`Enter <exec-instr-seq-enter>` the block :math:`H~(\val^m~\instr_1^\ast)~\END` with label :math:`L`.
+
+8. :ref:`Install <exec-handler-enter>` the exception handler `H` which contains `\val^m~\instr_1^\ast`.
```
```suggestion
8. :ref:`Install <exec-handler-enter>` the exception handler `H` for `\val^m~\instr_1^\ast`.
```

document/core/exec/instructions.rst
```diff
@@ -2625,16 +2625,29 @@ Control Instructions
...
+
+2. Let :math:`[t_1^m] \to [t_2^n]` be the :ref:`function type <syntax-functype>` :math:`\expand_F(\blocktype)`.
+
+3. Let :math:`L` be the label whose arity is :math:`n` and whose continuation is the end of the |TRY| instruction.
+
+4. Assert: due to :ref:`validation <valid-try-catch>`, there are at least :math:`m` values on the top of the stack.
+
+5. Pop the values :math:`\val^m` from the stack.
+
+6. Let :math:`H` be the :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm\{a_x~instr_2^\ast\}^\ast\{\epsilon~\instr_3^\ast\}^?`, whose mapping from tag addresses to branch targets corresponds to the |CATCH| and |CATCHALL| clauses, where :math:`a_x` is the tag address of each tag :math:`x`.
```
rossberg: I think this requires proper writing out into a loop over x*, which looks up each corresponding a_x in the store.

rossberg (changes requested):

document/core/exec/instructions.rst
rossberg: Actually, it would be more consistent with label blocks to phrase this
```
Entering :math:`\instr^\ast` with exception handler :math:`H`
```

ioannad: I like it! Switching to that.

document/core/exec/instructions.rst
ioannad: I left writing out the loop as a todo for now, either for this PR or for another one, as @aheejin prefers.

document/core/exec/instructions.rst
ioannad: done

document/core/exec/instructions.rst
ioannad: done

document/core/exec/instructions.rst
```diff
@@ -2635,11 +2635,14 @@ Control Instructions
...
-6. Let :math:`H` be the :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm\{a_x~instr_2^\ast\}^\ast\{\epsilon~\instr_3^\ast\}^?`, whose mapping from tag addresses to branch targets corresponds to the |CATCH| and |CATCHALL| clauses, where :math:`a_x` is the tag address of each tag :math:`x`.
+6. Let :math:`H` be the :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm\{a_x~instr_2^\ast\}^\ast\{\epsilon~\instr_3^\ast\}^?`, whose mapping from tag addresses to catch bodies corresponds to the |CATCH| and |CATCHALL| clauses, where :math:`a_x` is the tag address of each tag :math:`x`.
+
+   .. todo::
+       Step 6 needs proper writing out into a loop over x*, which looks up each corresponding a_x in the store.
 
 7. :ref:`Enter <exec-instr-seq-enter>` the block :math:`H~(\val^m~\instr_1^\ast)~\END` with label :math:`L`.
 
-8. :ref:`Install <exec-handler-enter>` the exception handler `H` which contains `\val^m~\instr_1^\ast`.
+8. :ref:`Enter <exec-handler-enter>` the block :math:`\val^m~\instr_1^\ast` with exception handler :math:`H`.
```
```suggestion
8. :ref:`Enter <exec-handler-enter>` :math:`\val^m~\instr_1^\ast` with exception handler :math:`H`.
```

document/core/exec/instructions.rst
ioannad: Addded a `.. todo::` section for now, but I plan to add this loop early next week.

document/core/exec/instructions.rst
ioannad: This is now reformulated to the "Enter..." version, [as agreed](https://github.com/WebAssembly/exception-handling/pull/224/files/cd6eb9d46a9cc4ddb4bc66b1fcc23612e8dbb1ed#r972825235).

document/core/exec/instructions.rst
ioannad: done

document/core/exec/instructions.rst
ioannad: done

document/core/exec/instructions.rst
ioannad: After [this suggestion](https://github.com/WebAssembly/exception-handling/pull/224#discussion_r972825235) this is now switched to the formulation "Enter instr* with exception handler H".

document/core/exec/instructions.rst
ioannad: Added the loop, everything should be ok now?

document/core/exec/instructions.rst
ioannad: done in previous commit

document/core/exec/instructions.rst
```diff
@@ -2625,16 +2625,41 @@ Control Instructions
...
+8. For each catch clause :math:`(\CATCH~x~\instr_2^\ast)` do:
+
+   a. Let :math:`a_x` be the tag address :math:`F.\AMODULE.\MITAGS[x]`.
+
+   b. Append :math:`\{a_x~\instr_2^\ast\}` to :math:`H`.
+
+9. If there is a catch all clause :math:`(\CATCHALL~\instr_3^\ast)`, then
+
+   a. Append :math:`\{\epsilon~\instr_3^\ast\}` to :math:`H`.
+
```
ioannad: @rossberg this is the newly added loop that constructs the catching exception handler `H`. I am not sure about also calling `(\CATCH~x~\instr_2^\ast)` a "catch clause", as this is currently the name we give to a `\CAUGHTADM` administrative instruction. They directly correspond to eachother though, in the sense that `\CAUGHTADM` is a selected `\CATCH` or `\CATCHALL`, so I used the same name to descibe it in this loop. WDYT?

ioannad: There was also [a suggestion](https://github.com/WebAssembly/exception-handling/pull/224#discussion_r972772672) by @aheejin to use "catch bodies", although IIUC this was originally meant to be used for the parts of the handler, so for `\{a_x~\instr_2^\ast\}` and `\{\epsilon~\instr_3^\ast\}`.

Not sure what's best for this version of the prose.

document/core/exec/instructions.rst
```diff
@@ -2635,14 +2635,23 @@ Control Instructions
...
 
-   .. todo::
-       Step 6 needs proper writing out into a loop over x*, which looks up each corresponding a_x in the store.
+7. Let :math:`H` be the empty :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm`.
 
-7. :ref:`Enter <exec-instr-seq-enter>` the block :math:`H~(\val^m~\instr_1^\ast)~\END` with label :math:`L`.
+8. For each catch clause :math:`(\CATCH~x~\instr_2^\ast)` do:
 
-8. :ref:`Enter <exec-handler-enter>` the block :math:`\val^m~\instr_1^\ast` with exception handler :math:`H`.
+   a. Let :math:`a_x` be the tag address :math:`F.\AMODULE.\MITAGS[x]`.
```
rossberg: Perhaps insert an assertion before this step that the binding exists due to validation.

document/core/exec/instructions.rst
```diff
@@ -2635,14 +2635,23 @@ Control Instructions
...
+
+   b. Append :math:`\{a_x~\instr_2^\ast\}` to :math:`H`.
+
+9. If there is a catch all clause :math:`(\CATCHALL~\instr_3^\ast)`, then
+
+   a. Append :math:`\{\epsilon~\instr_3^\ast\}` to :math:`H`.
+
+10. :ref:`Enter <exec-instr-seq-enter>` the block :math:`H~(\val^m~\instr_1^\ast)~\END` with label :math:`L`.
```
rossberg: This is a bit inconsistent with how block entering is described elsewhere (without the delimiters). The idea is that the prose does not actually write out the administrative delimiters.

document/core/exec/instructions.rst
```diff
@@ -2999,7 +3008,7 @@ Entering :math:`\instr` with exception handler :math:`H`
```
rossberg, line 3027: I believe the `\instr` on line 3002 needs to be `\instr^\ast`, too.

document/core/exec/instructions.rst
```diff
@@ -2635,14 +2635,23 @@ Control Instructions
 
 5. Pop the values :math:`\val^m` from the stack.
 
-6. Let :math:`H` be the :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm\{a_x~instr_2^\ast\}^\ast\{\epsilon~\instr_3^\ast\}^?`, whose mapping from tag addresses to catch bodies corresponds to the |CATCH| and |CATCHALL| clauses, where :math:`a_x` is the tag address of each tag :math:`x`.
+6. Let :math:`F` be the :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>`.
 
-   .. todo::
-       Step 6 needs proper writing out into a loop over x*, which looks up each corresponding a_x in the store.
+7. Let :math:`H` be the empty :ref:`catching exception handler <syntax-handler>` :math:`\CATCHadm`.
```
rossberg: `\CATCHadm` itself isn't really a legal phrase, since it merely is the header of a block-like construct.

I'd suggest something like:
```
7. For each catch clause :math:`(\CATCH~x_i~\instr_{2i}^\ast)` do:

   a. Assert: due to :ref:`validation <valid-try-catch>`, :math:`F.\AMODULE.\MITAGS[x_i]` exists.

   b. Let :math:`a_i` be the tag address :math:`F.\AMODULE.\MITAGS[x_i]`.

   c. Let :math:`H_i` be the handler :math:`\{a_i~\instr_{2i}^\ast\}`

8. Let :math:`H^\ast` be the concatenation of the handlers :math:`H_i`.

9. If there is a catch all clause :math:`(\CATCHALL~\instr_3^\ast)`, then:

    a. Let :math:`H^?` be the handler :math:`\{\epsilon~\instr_3^\ast\}`.

10. Else:

    a. Let :math:`H^?` be the empty handler :math:`\epsilon`.

11. :ref:`Enter <exec-handler-enter>` the block :math:`\val^m~\instr_1^\ast` with label :math:`L` and exception handler :math:`H`.
```
To make this work, you'll need to extend the definition of entering/exiting a handler slightly to include the label, to the form "Entering `\instr^\ast` with label L and handler H". But that would keep the abstraction level of the prose intact better.

ioannad: @rossberg , should 7.c. and 9a. say "handler" or something like "handler clause"? ATM a (catching) [handler is defined](https://webassembly.github.io/exception-handling/core/exec/runtime.html#syntax-handler) as `\CATCHadm{\tagaddr?~\instr*}*`. Which definition does allow for an empty ` \CATCHadm`. I guess I'm not sure what you mean by `\CATCHadm`  being "merely the header of a block-like construct".

document/core/exec/instructions.rst
ioannad: I think I understand now, what you mean with "header of a block-like construct". Ok, I will try to come up with a different formulation, although still I think the "parts" of a ` \CATCHadm` handler should not be also called handlers, but handler clauses or something similar. WDYT  @rossberg ?

document/core/exec/instructions.rst
ioannad: Done, together with the suggestions from the other review comment on `exec-try-catch`.

rossberg (approved):

document/core/exec/instructions.rst
```diff
@@ -2625,16 +2625,45 @@ Control Instructions
...
 
 .. math::
    ~\\[-1ex]
    \begin{array}{l}
-   F; \val^n~(\TRY~\X{bt}~\instr_1^\ast~(\CATCH~x~\instr_2^\ast)^\ast~(\CATCHALL~\instr_3^\ast)^?~\END
+   F; \val^m~(\TRY~\X{bt}~\instr_1^\ast~(\CATCH~x~\instr_2^\ast)^\ast~(\CATCHALL~\instr_3^\ast)^?~\END
    \quad \stepto \\
-   \qquad F; \LABEL_m\{\}~(\CATCHadm\{a~\instr_2^\ast\}^\ast\{\epsilon~\instr_3\ast\}^?~\val^n~\instr_1^\ast~\END)~\END \\
-   (\iff \expand_F(\X{bt}) = [t_1^n] \to [t_2^m] \land (F.\AMODULE.\MITAGS[x]=a)^\ast)
+   \qquad F; \LABEL_n\{\}~(\CATCHadm\{a_x~\instr_2^\ast\}^\ast\{\epsilon~\instr_3\ast\}^?~\val^m~\instr_1^\ast~\END)~\END \\
```
```suggestion
   \qquad F; \LABEL_n\{\epsilon\}~(\CATCHadm\{a_x~\instr_2^\ast\}^\ast\{\epsilon~\instr_3\ast\}^?~\val^m~\instr_1^\ast~\END)~\END \\
```
It looks `block`, `loop`, and `if` rules all have this. I also added this in `try`-`delegate` in #243.

document/core/exec/runtime.rst
```diff
@@ -550,10 +550,10 @@ If there is no :ref:`tag address <syntax-tagaddr>`, the instructions of that tar
      \production{(handler)} & \handler &::=& \CATCHadm\{\tagaddr^?~\instr^\ast\}^\ast &|& \DELEGATEadm\{l\}
    \end{array}
 
-Intuitively, for each target :math:`\{\tagaddr^?~\instr^\ast\}` of a |CATCHadm|, :math:`\instr^\ast` is the *continuation* to execute
-when the handler catches a thrown exception with tag |tagaddr|, or for any exception, when a target specifies no tag address.
-In that case, the exception is handled by the exception handler |CATCHadm|.
-If this list of targets is empty, or if the tag address of the thrown exception is not in the handler's mapping and there is no |CATCHALL| clause, then the exception will be rethrown.
+Intuitively, for each handler clause :math:`\{\tagaddr^?~\instr^\ast\}` of a |CATCHadm|, :math:`\instr^\ast` is the *continuation* to execute
+when the handler catches a thrown exception with tag |tagaddr|, or for any exception, when a handler clause specifies no tag address.
+In that case, we say that the exception is handled by the exception handler |CATCHadm|.
```
aheejin:
```suggestion
In that case, the exception is handled by the exception handler |CATCHadm|.
```
Do we need this "we say that"? Also I think uses of "we" are discouraged.

aheejin (approved):
LGTM % comments

PR #225, opened 2022-09-07
Add prose for reduction rule of rethrow.

ioannad:
<del>Why WIP:</del>
--------
<del>I'm not sure how to justify step 5 of the prose, that is how to assert that the lth label surrounding a rethrow instruction has a CAUGHTadm{a val*} next to it.</del>

- Also slightly changed notation to match the [formal overview document](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions-formal-overview.md).

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2719,30 +2719,13 @@ Control Instructions
 
 2. Let :math:`L` be the :math:`l`-th label appearing on the stack, starting from the top and counting from zero.
 
-3. Assert: due to :ref:`validation <valid-rethrow>`, :math:`L` is a catch label, i.e., a label of the form :math:`(\LCATCH~[t^\ast])`.
+3. Assert: due to :ref:`validation <valid-rethrow>`, :math:`L` is a catch label, i.e., a label of the form :math:`(\LCATCH~[t^\ast])`, which is a label followed by a caught exception in an active catch clause.
 
-4. Repeat :math:`l+1` times:
+4. Let :math:`\{a~\val^\ast\}` be the caught exception.
 
-   a. While the top of the stack is a value, a |handler|, or a |CAUGHTadm| instruction, do:
+5. Push the values :math:`\val^\ast` onto the stack.
 
-      i. Pop the instruction from the stack.
-
-   b. Assert: due to :ref:`validation <valid-rethrow>`, the top of the stack now is a label.
-
-   c. Pop the label from the stack.
-
-5. Assert: due to **TODO**, the last two instructions popped are a |CAUGHTadm| instruction and the :math:`l+1`-th label, in that order.
-
-.. todo::
-   Justify the existence of a |CAUGHTadm| here. Perhaps extend some typing rule?
-
-6. Let :math:`\CAUGHTadm\{a~\val^\ast\}` be the instruction after the label :math:`L`.
-
-7. Put all the popped values back on the stack, except the last instruction, the |RETHROW|.
-
-8. Push the values :math:`\val^\ast` onto the stack.
-
-8. :ref:`Throw <exec-throwadm>` an exception with :ref:`tag address <syntax-tagaddr>` :math:`a`.
+6. :ref:`Throw <exec-throwadm>` an exception with :ref:`tag address <syntax-tagaddr>` :math:`a`.
 
 .. math::
    ~\\[-1ex]
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2715,7 +2715,7 @@ Control Instructions
 :math:`\RETHROW~l`
 ..................
 
-1. Assert: due to :ref:`validation <valid-rethrow>`, the stack contans at least :math:`l+1` labels.
+1. Assert: due to :ref:`validation <valid-rethrow>`, the stack contains at least :math:`l+1` labels.
 
 2. Let :math:`L` be the :math:`l`-th label appearing on the stack, starting from the top and counting from zero.
 
@@ -2723,7 +2723,7 @@ Control Instructions
 
 4. Repeat :math:`l+1` times:
 
-   a. While the instruction at the top of the stack is a value, a |handler|, or a |CAUGHTadm| instruction, do:
+   a. While the top of the stack is a value, a |handler|, or a |CAUGHTadm| instruction, do:
 
       i. Pop the instruction from the stack.
 
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2715,14 +2715,40 @@ Control Instructions
 :math:`\RETHROW~l`
 ..................
 
+1. Assert: due to :ref:`validation <valid-rethrow>`, the stack contans at least :math:`l+1` labels.
+
+2. Let :math:`L` be the :math:`l`-th label appearing on the stack, starting from the top and counting from zero.
+
+3. Assert: due to :ref:`validation <valid-rethrow>`, :math:`L` is a catch label, i.e., a label of the form :math:`(\LCATCH~[t^\ast])`.
+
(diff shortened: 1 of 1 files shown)


document/core/exec/instructions.rst
```diff
@@ -2680,14 +2680,40 @@ Control Instructions
 :math:`\RETHROW~l`
 ..................
 
+1. Assert: due to :ref:`validation <valid-rethrow>`, the stack contans at least :math:`l+1` labels.
```
```suggestion
1. Assert: due to :ref:`validation <valid-rethrow>`, the stack contains at least :math:`l+1` labels.
```

document/core/exec/instructions.rst
```diff
@@ -2680,14 +2680,40 @@ Control Instructions
...
+
+2. Let :math:`L` be the :math:`l`-th label appearing on the stack, starting from the top and counting from zero.
+
+3. Assert: due to :ref:`validation <valid-rethrow>`, :math:`L` is a catch label, i.e., a label of the form :math:`(\LCATCH~[t^\ast])`.
+
+4. Repeat :math:`l+1` times:
+
+   a. While the instruction at the top of the stack is a value, a |handler|, or a |CAUGHTadm| instruction, do:
```
```suggestion
   a. While the top of the stack is a value, a |handler|, or a |CAUGHTadm| instruction, do:
```
Nit: It's unclear we can call values instructions, and rules for `br` and `return` also just say "While the top of the stack ..."

aheejin (commented):
We extended labels to include an optional `catch` in validation:
```
label := catch'? resulttype
```
Is this not used in execution semantics? If we use this, can we this to guarantee there's a `catch` label on the stack, which is equivalent of having `caught`? I might be completely confused about different things... in which case please let me know.

ioannad:
> We extended labels to include an optional `catch` in validation:
> 
> ```
> label := catch'? resulttype
> ```
> 
> Is this not used in execution semantics? If we use this, can we this to guarantee there's a `catch` label on the stack, which is equivalent of having `caught`? I might be completely confused about different things... in which case please let me know.

I don't think you are confused, my problem is that I don't know how to use it there. So by validation we know that the l'th label of the context C has the form `catch [t*]`, but why is the `l'th` administrative instruction followed by an administrative instruction `caught{a val*}`? The only thing I have is that validation of `caught` requires a `catch [t*]` label in the context, and it's the only instruction that requires such a label, but I can't connect the last dots after that. Is it obvious and I'm just stuck? I don't know. :sweat_smile:

document/core/exec/instructions.rst
```diff
@@ -2680,14 +2680,40 @@ Control Instructions
 :math:`\RETHROW~l`
 ..................
 
+1. Assert: due to :ref:`validation <valid-rethrow>`, the stack contains at least :math:`l+1` labels.
+
+2. Let :math:`L` be the :math:`l`-th label appearing on the stack, starting from the top and counting from zero.
+
+3. Assert: due to :ref:`validation <valid-rethrow>`, :math:`L` is a catch label, i.e., a label of the form :math:`(\LCATCH~[t^\ast])`.
+
+4. Repeat :math:`l+1` times:
```
rossberg: I suggest doing it only l times in the loop, so that you haven't skipped over the caught clause already.

document/core/exec/instructions.rst
```diff
@@ -2680,14 +2680,40 @@ Control Instructions
...
+
+   a. While the top of the stack is a value, a |handler|, or a |CAUGHTadm| instruction, do:
+
+      i. Pop the instruction from the stack.
+
+   b. Assert: due to :ref:`validation <valid-br>`, the top of the stack now is a label.
+
+   c. Pop the label from the stack.
+
+5. Assert: due to **TODO**, the last two instructions popped are a |CAUGHTadm| instruction and the :math:`l+1`-th label, in that order.
```
rossberg: I'd use typing of `rethrow` as a proxy.

document/core/exec/instructions.rst
```diff
@@ -2680,14 +2680,40 @@ Control Instructions
...
+
 .. todo::
-   Add prose for the |RETHROW| execution step.
+   Justify the existence of a |CAUGHTadm| here. Perhaps extend some typing rule?
+
+6. Let :math:`\CAUGHTadm\{a~\val^\ast\}` be the instruction after the label :math:`L`.
+
+7. Put all the popped values back on the stack, except the last instruction, the |RETHROW|.
```
rossberg: That seems really odd. As commented elsewhere, I think we should avoid this.

document/core/exec/instructions.rst
```diff
@@ -2680,14 +2680,40 @@ Control Instructions
...
+
+8. Push the values :math:`\val^\ast` onto the stack.
+
+8. :ref:`Throw <exec-throwadm>` an exception with :ref:`tag address <syntax-tagaddr>` :math:`a`.
```
rossberg: I think you somehow need to jump to the right point first.

document/core/exec/instructions.rst
```diff
@@ -2680,14 +2680,40 @@ Control Instructions
...
+
+3. Assert: due to :ref:`validation <valid-rethrow>`, :math:`L` is a catch label, i.e., a label of the form :math:`(\LCATCH~[t^\ast])`.
+
+4. Repeat :math:`l+1` times:
+
+   a. While the top of the stack is a value, a |handler|, or a |CAUGHTadm| instruction, do:
+
+      i. Pop the instruction from the stack.
+
+   b. Assert: due to :ref:`validation <valid-br>`, the top of the stack now is a label.
```
```suggestion
   b. Assert: due to :ref:`validation <valid-rethrow>`, the top of the stack now is a label.
```

document/core/exec/instructions.rst
aheejin: How do you think can we avoid this? Can we use a concept of shadow stack or something, so we don't really pop the values in 4-a-i above but just scan through it?

Should we do this looping thing in 4-a-i after all? All other branch-like instructions do it so there must be a reason, but can't we just say, because of validation, we can assert `L`th is a catch label with type `[t*]->[]` and `L-1`th is a `caught` instruction because of try-catch execution rule, and just refer to the `val`s?

document/core/exec/instructions.rst
aheejin: Can you elaborate a little more?

document/core/exec/instructions.rst
aheejin: Where is the right point? Shouldn't we throw the exception in the original location?

document/core/exec/instructions.rst
ioannad: I removed the loop entirely, trying to avoid popping and pushing back the entire block context.

document/core/exec/instructions.rst
ioannad: I removed the loop as mentioned in the previous comment, and replaced the assertion with:

```
3. Assert: due to :ref:`validation <valid-rethrow>`, :math:`L` is a catch label, i.e., a label of the form :math:`(\LCATCH~[t^\ast])`, which is a label followed by a caught exception in an active catch clause.
```

Is that formulation ok @rossberg ?

document/core/exec/instructions.rst
ioannad: As mentioned above, I just pushed an attempt to avoid all this pushing a popping. WDYT?

rossberg (approved):

aheejin (approved):

ioannad:
The force push was to rebase to the current exception-handling/main, to make sure all checks pass before merging.

rossberg:
> The force push was to rebase to the current exception-handling/main, to make sure all checks pass before merging.

FWIW, instead of rebase you can simply merge `main` into this branch, which has the same effect but keeps log and PR intact.

ioannad:
@rossberg sorry about that, it was careless, because the PR is approved.

PR #226, opened 2022-09-07
Add prose for reduction rules involving thrown exceptions,

ioannad:
i.e., prose for some execution steps for `THROWadm`, `CATCHadm`, `DELEGATEadm`.

The parts of this commit that involve `DELEGATEadm` may change once PR #220 is settled.

Nonetheless, in this commit the typing rule and the reduction rule for `DELEGATEadm` are both tentatively changed to match the current [formal overview](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions-formal-overview.md).


document/core/appendix/properties.rst
```diff
@@ -633,11 +634,11 @@ To that end, all previous typing judgements :math:`C \vdash \X{prop}` are genera
...
 .. math::
    \frac{
-     S; C,\CLABELS\,[t^\ast] \vdashinstrseq \instr^\ast : [] \to [t^\ast]
+     S; C \vdashinstrseq \instr^\ast : [] \to [t^\ast]
      \qquad
-     C.\CLABELS[l] = [t_0^\ast]
+     C.\CLABELS[l+1] = [t_0^\ast]
    }{
-     S; C,\CLABELS\,[t^\ast] \vdashadmininstr \DELEGATEadm\{l\}~\instr^\ast~\END : [] \to [t^\ast]
+     S; C \vdashadmininstr \DELEGATEadm\{l\}~\instr^\ast~\END : [] \to [t^\ast]
```
aheejin: Is this change related to the PR? If not, can we make it a separate PR or merge this into another `delegate` PR?

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
 Throwing an exception with :ref:`tag address <syntax-tagaddr>` :math:`a`
 ........................................................................
 
+When an administrative throw occurs, then values, labels, |CAUGHTadm| instructions,
+and call frames are popped if necessary, until an appropriate exception handler is found
+on the top of the stack.
+
+ 1. Assert: due to validation, :math:`S.\STAGS[a]` exists.
```
```suggestion
 1. Assert: due to :ref:`validation <valid-throw>`,, :math:`S.\STAGS[a]` exists.
```

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
...
+
+    b. Else :math:`H` is of the form :math:`\CATCHadm\{a_1^?~\instr^\ast\}\{a'^?~\instr'^\ast\}^\ast`.
+
+    c. If :math:`a_1^? = \epsilon`, then:
+
+       i. Push :math:`\CAUGHTadm\{a~\val^n\}` onto the stack.
 
+       ii. Jump to the start of the instruction sequence :math:`\instr^\ast`.
+
+    d. Else if :math:`a_1 = a`, then:
```
```suggestion
    d. Else if :math:`a_1^? = a`, then:
```
Do we not need this to be consistent with 'c'?

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
...
+
+       i. Push :math:`\CAUGHTadm\{a~\val^n\}` onto the stack.
```
aheejin: Does `val^n` mean a different thing from the line "ii" below? I guess this line doesn't mean actual values but just notes what the tag's type is... right?

aheejin:
Gentle ping :)

document/core/appendix/properties.rst
ioannad: Done.

document/core/exec/instructions.rst
ioannad: > Does `val^n` mean a different thing from the line "ii" below? 

No, these `val^n` are the same `val^n` from the line "ii" below. So a `caught{a val^n}` holds the exception while the instructions in the catch block are executed. Since this is a `catch x_a` (due to "d"), the values in the exception are pushed to the stack first.

> I guess this line doesn't mean actual values but just notes what the tag's type is... right?

In fact `caught` holds the actual exception, so its tag and its values.

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
 Throwing an exception with :ref:`tag address <syntax-tagaddr>` :math:`a`
 ........................................................................
 
+When an administrative throw occurs, then values, labels, |CAUGHTadm| instructions,
```
```suggestion
When a throw occurs, then values, labels, active catch clauses,
```

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
...
    && (\iff~(a_1^? = \epsilon \lor a_1^? = a)~\land\\
    && \ S.\STAGS[a].\TAGITYPE = [t^n]\to[]) \\
-   \LABEL_n\{\}~\XB^l[\DELEGATEadm\{l\}~\XT[(\THROWadm~a)]~\END]~\END &\stepto&
-   \XT[(\THROWadm~a)]  \\
+   \LABEL_n\{\}~\XB^l[\DELEGATEadm\{l\}~\XT[\val^n~(\THROWadm~a)]~\END]~\END &\stepto&
+   \XT[\val^n~(\THROWadm~a)]  \\
    \end{array}
 
+.. note::
+   Note that the reduction step resulting in a |CAUGHTadm| instruction is the only one that does not preserve the throw context.
```
```suggestion
   The reduction step activating a catch clause |CAUGHTadm| is the only one that does not preserve the throw context.
```

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
...
+   While a |THROWadm| propagates through the stack, the throw context |XT| is collected
+   until the exception is caught inside a |CAUGHTadm| instruction, in which point it's discarded.
```
```suggestion
   until the exception is caught inside a |CAUGHTadm| instruction, at which point it is discarded.
```

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
...
+17. Jump to the continuation of :math:`L`.
 
 .. math::
    \begin{array}{rcl}
-   \CATCHadm\{a_1^?~\instr^\ast\}\{a'^?~\instr'^\ast\}^\ast~\XT[(\THROWadm~a)]~\END &\stepto&
-   \CATCHadm\{a'^?~\instr'^\ast\}^\ast~\XT[(\THROWadm~a)]~\END  \\
+   \CATCHadm~\XT[\val^n~(\THROWadm~a)]~\END &\stepto&
+   \XT[\val^n~(\THROWadm~a)] \\
+   \CATCHadm\{a_1^?~\instr^\ast\}\{a'^?~\instr'^\ast\}^\ast~\XT[\val^n~(\THROWadm~a)]~\END &\stepto&
+   \CATCHadm\{a'^?~\instr'^\ast\}^\ast~\XT[\val^n~(\THROWadm~a)]~\END  \\
```
rossberg: I don't think this formulation works, since you don't know what `n` is without looking up a's arity in the store in every single rule. You could do that, but the previous formulation instead keeps XT around in the latter rule so that extracting the right number of values from it can be delayed until the handler is found (where you have to look it up again).

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
...
    && \ S.\STAGS[a].\TAGITYPE = [t^n]\to[]) \\
-   \LABEL_n\{\}~\XB^l[\DELEGATEadm\{l\}~\XT[(\THROWadm~a)]~\END]~\END &\stepto&
-   \XT[(\THROWadm~a)]  \\
+   \LABEL_n\{\}~\XB^l[\DELEGATEadm\{l\}~\XT[\val^n~(\THROWadm~a)]~\END]~\END &\stepto&
+   \XT[\val^n~(\THROWadm~a)]  \\
    \end{array}
 
+.. note::
+   Note that the reduction step resulting in a |CAUGHTadm| instruction is the only one that does not preserve the throw context.
+   While a |THROWadm| propagates through the stack, the throw context |XT| is collected
```
rossberg: This does not explain why, but see above.

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
...
+
+ 2. Let :math:`[t^n] \to []` be the :ref:`tag type <syntax-tagtype>` :math:`S.\STAGS[a].\TAGITYPE`.
+
+ 3. Assert: due to :ref:`validation <valid-throw>`, there are :math:`n` values on the top of the stack.
+
+ 4. Pop the :math:`n` values :math:`\val^n` from the stack.
+
+ 5. While the stack is not empty and the top of the stack is not an :ref:`exception handler <syntax-handler>`, do:
+
+    a. Pop the top element from the stack, prepending it to the :ref:`throw context <syntax-ctxt-throw>` of the exception: :math:`\XT[\val^n~(\THROWadm~a)]`.
```
rossberg: The spec so far carefully avoids to use the notion of contexts in the prose. I think you can reformulate this in a way more similar to branches, where there is no need to mention XT.

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
...
+
+       i. Push the throw context that we collected so far :math:`\XT[\val^n~(\THROWadm~a)]` onto the stack.
+
+    b. Else :math:`H` is of the form :math:`\CATCHadm\{a_1^?~\instr^\ast\}\{a'^?~\instr'^\ast\}^\ast`.
+
+    c. If :math:`a_1^? = \epsilon`, then:
+
+       i. Push :math:`\CAUGHTadm\{a~\val^n\}` onto the stack.
 
+       ii. Jump to the start of the instruction sequence :math:`\instr^\ast`.
```
rossberg: I believe this ought to be formulated in terms of a separate step of "Entering a catch clause", mirroring the exit you have already introduced. (Same below)

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
...
+
+13. Assert: due to :ref:`validation <valid-delegate-admin>`, the stack contains at least :math:`l+1` labels.
+
+14. Let :math:`L` be the :math:`l`-th label appearing on the stack, starting from the top and counting from zero.
+
+15. Repeat :math:`l+1` times:
+
+    a. While the instruction on the top of the stack is not a label, do:
+
+       i. Pop the instruction from the stack, without pushing it to |XT|.
```
rossberg: The informal stack does not contain "instructions", only value, labels, frames, and now handlers and catch clauses. We probably need to find a common name for those ("control frames"?).

Also, the formulation of `br` has to change, since it no longer just skips over values, it also pops EH-related control frames.

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
...
+
+   b. Assert: due to :ref:`validation <valid-delegate-admin>`, the top of the stack now is a label.
+
+   c. Pop the label from the stack.
+
+16. Push the throw context that we collected so far :math:`\XT[\val^n~(\THROWadm~a)]` onto the stack.
```
rossberg: Yeah, I'm not sure what that would mean. I think we need a wording without XT.

rossberg (changes requested):

document/core/exec/instructions.rst
ioannad: Oh, I see, thank you! switching to the previous throw contexts and adding your commend as explainer.

document/core/exec/instructions.rst
ioannad: I'm not sure how to search the stack for a handler without popping stuff from it. So I don't know how to describe keeping the throw context in these rules.

rossberg: Why do you need to keep the context in the prose? Above, it has already popped the operands `val^n`, and that's all you need to keep AFAICS.

ioannad: Don't we need it to return the entire stack if the exception is [not handled](https://webassembly.github.io/exception-handling/core/exec/runtime.html?highlight=exception#results)?

document/core/exec/instructions.rst
ioannad: @rossberg 
I'm not sure any more what this review comment means, in particular which formulation it refers to. I think I misunderstood when I  thought it refers to previous formulation of _throw contexts_, because IIUC any non-trivial throw context would still have the exception payload values `val^n` inside the square brackets with the throw (`\XT[val^n (throw a)]`). 

Which previous formulation are you referring to?

document/core/exec/instructions.rst
ioannad: @rossberg this comment is now removed, but I'm still not sure why the rules keep the throw context, so I added a `todo::`  to add an explanation why or what's the meaning of apparently preserving the throw context in some of the rules.

document/core/exec/instructions.rst
ioannad: @rossberg I removed this now to just popping the stack, like [above](https://github.com/WebAssembly/exception-handling/pull/226/files#r1044887407). Should the uncaught exceptions be changed to `val^n (throw a)`?

document/core/exec/instructions.rst
ioannad: Rephrased to `    a. Pop the top element from the stack.`

Should I make a new PR rephrasing `document/core/exec/runtime.rst` to add control frames anyway?

Just for the record, the formulation of `br` is now changed by merged PR #233.

document/core/exec/instructions.rst
ioannad: Removed everything after ", prepending".

document/core/exec/instructions.rst
```diff
@@ -3077,66 +3077,69 @@ on the top of the stack.
...
 
-    a. Pop the top element from the stack, prepending it to the :ref:`throw context <syntax-ctxt-throw>` of the exception: :math:`\XT[\val^n~(\THROWadm~a)]`.
+    a. Pop the top element from the stack.
 
  6. Assert: The stack is now either empty, or there is an exception handler on the top of the stack.
 
  7. If the stack is empty, then:
 
-    a. **TODO** *Return a result value representing the uncaught exception (will probably just be the same as 11.a.i. below).*
+    a. Return the uncaught exception :math:`\val^n~(\THROWadm~a)` as a value.
```
```suggestion
    a. Return the uncaught exception :math:`\val^n~(\THROWadm~a)` as a result.
```

document/core/exec/instructions.rst
```diff
@@ -3077,66 +3077,69 @@ on the top of the stack.
...
 
-.. todo::
-   After PR #221 is resolved, this step should be filled in with a PR to specify uncaught exception results.
+8. Else:
 
-8. Else there is an :ref:`exception handler <syntax-handler>` :math:`H` on the top of the stack.
+   a. Assert: there is an :ref:`exception handler <syntax-handler>` :math:`H` on the top of the stack.
```
rossberg: It seems off that this is under `else` but what follows is not.

document/core/exec/instructions.rst
```diff
@@ -3077,66 +3077,69 @@ on the top of the stack.
...
 
 9. Pop the exception handler :math:`H` from the stack.
 
-10. Assert: :math:`H` is either of the form :math:`\CATCHadm\{a^?~\instr^\ast\}^k` or :math:`\DELEGATEadm\{l\}.`
+10. Assert: :math:`H` is either of the a catching or a delegating :ref:`exception handler <syntax-handler>`.
```
rossberg: Is this assertion needed? That appears to be true by definition of exception handler.

document/core/exec/instructions.rst
```diff
@@ -3077,66 +3077,69 @@ on the top of the stack.
...
+       i. :ref:`Enter <exec-caughtadm-enter>` :math:`\val^n~\instr^\ast` with catch clause holding the caught exception :math:`\{a~\val^n\}`.
 
-       ii. Push the values :math:`\val^n` back to the stack.
+    e. Else:
 
-       iii. Jump to the start of the instruction sequence :math:`\instr^\ast`.
+       i. Let :math:`H'` be the catching exception handler with clauses :math:`\{a'^?~\instr'^\ast\}^\ast`.
 
-    e. Else, repeat step 11 for :math:`H = \CATCHadm\{a'^?~\instr'^\ast\}^\ast`.
+       ii. Repeat step 11 for :math:`H'`.
```
rossberg: Let's turn this use of goto into a structured while loop. ;)

document/core/exec/instructions.rst
```diff
@@ -3077,66 +3077,69 @@ on the top of the stack.
...
 
-12. Else the handler :math:`H` has the form :math:`\DELEGATEadm\{l\}`.
+12. Else:
 
-13. Assert: due to :ref:`validation <valid-delegate-admin>`, the stack contains at least :math:`l+1` labels.
+    a. Assert: :math:`H` is a delegating exception handler.
```
rossberg: Same nesting oddity here.

document/core/exec/instructions.rst
```diff
@@ -3077,66 +3077,69 @@ on the top of the stack.
...
 
-14. Let :math:`L` be the :math:`l`-th label appearing on the stack, starting from the top and counting from zero.
+13. Let :math:`l` be the label index of :math:`H`.
 
-15. Repeat :math:`l+1` times:
+14. Assert: due to :ref:`validation <valid-delegate-admin>`, the stack contains at least :math:`l` labels.
 
-    a. While the instruction on the top of the stack is not a label, do:
+15. Let :math:`L` be the :math:`l`-th label appearing on the stack, starting from the top and counting from zero.
```
rossberg: What is this step? `L` doesn't even seem to be used.

document/core/exec/instructions.rst
```diff
@@ -3077,66 +3077,69 @@ on the top of the stack.
...
-       i. Push :math:`\CAUGHTadm\{a~\val^n\}` onto the stack.
+    c. If :math:`a_1^? = \epsilon`, then:
 
-       ii. Jump to the start of the instruction sequence :math:`\instr^\ast`.
+       i. :ref:`Enter <exec-caughtadm-enter>` :math:`\instr^\ast` with catch clause holding the caught exception :math:`\{a~\val^n\}`.
 
     d. Else if :math:`a_1^? = a`, then:
 
-       i. Push :math:`\CAUGHTadm\{a~\val^n\}` onto the stack.
+       i. :ref:`Enter <exec-caughtadm-enter>` :math:`\val^n~\instr^\ast` with catch clause holding the caught exception :math:`\{a~\val^n\}`.
```
rossberg: You defined how to enter it, but not how it is exited.

Perhaps we don't need these auxiliary definitions for something as simple as this. Just push the value, enter the instruction sequence, pop the exception.

document/core/exec/instructions.rst
```diff
@@ -3146,20 +3149,28 @@ on the top of the stack.
...
+   The reduction step entering a catch clause |CAUGHTadm| is the only one that does not preserve the throw context.
 
 .. todo::
-   Add explainer note for the reduction of |DELEGATEadm|, when PR #221 is settled.
+   Explain why or what's the meaning of apparently preserving the throw context in the other rules.
+
+
+.. _exec-caughtadm-enter:
+
+Entering :math:`\instr^\ast` with catch clause holding a caught exception :math:`\{a~\val^n\}`
```
rossberg: See comment above.

document/core/exec/runtime.rst
```diff
@@ -550,10 +551,11 @@ If there is no :ref:`tag address <syntax-tagaddr>`, the instructions of that han
      \production{(handler)} & \handler &::=& \CATCHadm\{\tagaddr^?~\instr^\ast\}^\ast &|& \DELEGATEadm\{l\}
    \end{array}
 
-Intuitively, for each handler clause :math:`\{\tagaddr^?~\instr^\ast\}` of a |CATCHadm|, :math:`\instr^\ast` is the *continuation* to execute
-when the handler catches a thrown exception with tag |tagaddr|, or for any exception, when a handler clause specifies no tag address.
-In that case, the exception is handled by the exception handler |CATCHadm|.
-If this list of targets is empty, or if the tag address of the thrown exception is not in any of the handler's clauses and there is no |CATCHALL| clause, then the exception will be rethrown.
+Intuitively, for each catch clause :math:`\{\tagaddr^?~\instr^\ast\}` of a catching handler, :math:`\instr^\ast` is the *continuation* to execute
+when the handler catches a thrown exception with tag |tagaddr|, or for any exception, when a catch clause specifies no tag address.
```
```suggestion
when the handler catches a thrown exception with tag |tagaddr|, or for any exception, when the catch clause specifies no tag address.
```

document/core/exec/instructions.rst
rossberg: I believe you can just use `\XT[(throw a)]`, since the context will cover any val* anyway.

document/core/exec/instructions.rst
rossberg: I no longer think we need to define "control frame". Just pop the "top element", as you wrote above.

document/core/exec/instructions.rst
rossberg: See my email: you need to allow (uncaught) exceptions as stack elements and push it as that.

document/core/exec/instructions.rst
```diff
@@ -3016,26 +3016,104 @@ When the end of a :ref:`try <syntax-try>` instruction is reached without a jump,
...
+
+       iii. Jump to the start of the instruction sequence :math:`\instr^\ast`.
+
+    e. Else, repeat step 11 for :math:`H = \CATCHadm\{a'^?~\instr'^\ast\}^\ast`.
+
+12. Else the handler :math:`H` has the form :math:`\DELEGATEadm\{l\}`.
+
+13. Assert: due to :ref:`validation <valid-delegate-admin>`, the stack contains at least :math:`l+1` labels.
+
+14. Let :math:`L` be the :math:`l`-th label appearing on the stack, starting from the top and counting from zero.
```
rossberg: This would be much simpler to express after the loop below.

document/core/exec/instructions.rst
ioannad: Should I rename section `_exec-caughtadm` to `_exec-caughtadm-exit` or remove both sections `_exec-caughtadm-enter` and `_exec-caughtadm`, and merge their steps into this prose directly?

document/core/exec/instructions.rst
ioannad: I am changing this to 

```
8. Else assert: there is an :ref:`exception handler <syntax-handler>` :math:`H` on the top of the stack.
```

This way I hope it's clear that everything that follows is under the else, given that the other branch of the "If" (`7.a`) returns. 

Or would you prefer that everything that follows is nested under `8.`?

document/core/exec/instructions.rst
ioannad: Removing this line.

document/core/exec/instructions.rst
ioannad: Since the other "if" branch does not end with a clear "return", here I am putting the rest of the steps nested under the "Else".

document/core/exec/instructions.rst
ioannad: Removing.

document/core/exec/instructions.rst
ioannad: In this case, I'm not sure how to write the reduction for a `CATCHadm` with an empty list of catch clauses.
```
  \CATCHadm~\XT[\THROWadm~a]~\END & \stepto & ???
```

document/core/exec/instructions.rst
ioannad: This sounds like the answer to my question [above](https://github.com/WebAssembly/exception-handling/pull/226/files#r1101318259), is to remove both sections `_exec-caughtadm-enter` and `_exec-caughtadm`, and merge their steps into the `exec-throwadm` prose directly?

document/core/exec/instructions.rst
ioannad: Nevermind my last comment, I think I got it right with the latest commits.

document/core/exec/instructions.rst
ioannad: Thank you for the explanation via email, I included it as a note.

document/core/exec/runtime.rst
```diff
@@ -543,11 +543,11 @@ to their associated branch *targets*. A single handler is expressed syntacticall
...
 If there is no :ref:`tag address <syntax-tagaddr>`, the instructions of that handler correspond to a |CATCHALL| clause.
 
-.. todo::
-   Add prose for delegating handlers.
-
 An exception may be temporarily pushed onto the stack when it is :ref:`thrown and caught <exec-throwadm>` by a handler.
 
+A |labelidx| handler points to the outer block in which any exceptions thrown from its scope, will be rethrown (delegated to).
+This handler does not catch exceptions, but only rethrows them.
+
```
rossberg, line 548: Note: I noticed a few weeks ago that there were places in the spec where `production` accidentally has an argument with explicit parens. I fixed in the spec, but please check that that mistake isn't duplicated in this repo (like on the grammar below).

document/core/exec/runtime.rst
```diff
@@ -543,11 +543,11 @@ to their associated branch *targets*. A single handler is expressed syntacticall
 :ref:`instructions <syntax-instr>` possibly following a :ref:`tag address <syntax-tagaddr>`.
 If there is no :ref:`tag address <syntax-tagaddr>`, the instructions of that handler correspond to a |CATCHALL| clause.
 
-.. todo::
-   Add prose for delegating handlers.
-
 An exception may be temporarily pushed onto the stack when it is :ref:`thrown and caught <exec-throwadm>` by a handler.
 
+A |labelidx| handler points to the outer block in which any exceptions thrown from its scope, will be rethrown (delegated to).
```
```suggestion
A handler can also consist of a single |labelidx|, which denotes an outer block to which every caught exception will be delegated, by implicitly rethrowing inside that block.
```

document/core/exec/runtime.rst
```diff
@@ -561,9 +561,7 @@ If the list of handlers is empty, or if the tag address of the thrown exception
 When a thrown exception is caught by a handler, the caught exception is pushed onto the stack and the block of that handler's target is :ref:`entered <exec-caughtadm-enter>`.
 When exiting a block with a caught exception, the exception is discarded.
 
-
-.. todo::
-   Add prose with intuition on delegating handlers.
+A handler pointing to a |labelidx| :math:`l` can be thought of as a break to the :math:`l` th label on exception, followed by a rethrow of the exception.
```
```suggestion
A handler consisting of a |labelidx| :math:`l` can be thought of as a branch to that label that happens in case an exception occurs, immediately followed by a rethrow of the exception at the target site.
```

document/core/exec/runtime.rst
ioannad: Oh oops, I think I added explicit parens everywhere, removing them.

ioannad: Done.

document/core/exec/runtime.rst
```diff
@@ -75,14 +75,14 @@ Results
 ~~~~~~~
 
 A *result* is the outcome of a computation.
-It is either a sequence of :ref:`values <syntax-val>`, a :ref:`trap <syntax-trap>`, or an uncaught exception wrapped in its :ref:`throw context <syntax-ctxt-throw>`.
+It is either a sequence of :ref:`values <syntax-val>`, a :ref:`trap <syntax-trap>`, or an :ref:`uncaught exception <exec-throwadm>`.
```
```suggestion
It is either a sequence of :ref:`values <syntax-val>`, a :ref:`trap <syntax-trap>`, or an :ref:`exception <exec-throwadm>`.
```

document/core/exec/runtime.rst
```diff
@@ -75,14 +75,14 @@ Results
...
 
 .. math::
    \begin{array}{llcl}
    \production{result} & \result &::=&
      \val^\ast \\&&|&
      \TRAP  \\&&|&
-     \XT[\val^\ast~(\THROWadm~\tagaddr)]
+     \val^\ast~(\THROWadm~\tagaddr)
```
```suggestion
     \XT[(\THROWadm~\tagaddr)]
```
I believe we need to keep the `XT` here, unfortunately, since there is no rule that would reduce it away.

rossberg (approved):

PR #227, opened 2022-09-07
Add prose for exiting a CAUGHTadm block.

ioannad:
Also slightly changed notation to match the [formal overview document](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions-formal-overview.md).

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -3039,16 +3039,31 @@ Throwing an exception with :ref:`tag address <syntax-tagaddr>` :math:`a`
 
 .. _exec-caughtadm:
 
-Holding a caught exception with |CAUGHTadm|
-...........................................
+Exiting a |CAUGHTadm|
+.....................
 
-.. todo::
-   Add prose describing the administrative |CAUGHTadm| execution step.
+When the |END| of a |CAUGHTadm|, is reached without a jump, exception, or trap, then the following steps are performed.
+
+1. Let :math:`\val^\ast` be the values on the top of the stack.
+
+2. Pop the values :math:`\val^\ast` from the stack.
+
+3. Assert: due to :ref:`validation <valid-instr-seq>`, an administrative instruction :math:`\CAUGHTadm\{a~\val_0^\ast\}` is now on the top of the stack.
+
+4. Pop the |CAUGHTadm| from the stack.
+
+5. Push :math:`\val^\ast` back to the stack.
+
+6. Jump to the position after the |END| of the administrative instruction associated with the |CAUGHTadm| instruction.
+
 
 .. math::
    \begin{array}{rcl}
-   \CAUGHTadm\{a~\val^n\}~\val^m~\END  &\stepto& \val^m
+   \CAUGHTadm\{a~\val_0^\ast\}~\val^\ast~\END  &\stepto& \val^\ast
    \end{array}
+
+.. note::
+   An exception can only be rethrown from the scope of the |CAUGHTadm| administrative instruction holding it, i.e., from the scope of the |CATCH| or |CATCHALL| block of a :ref:`try-catch <syntax-try-catch>` instruction that caught it. Upon exit from a |CAUGHTadm|, the exception it holds is discarded.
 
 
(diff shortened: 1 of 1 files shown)


document/core/exec/instructions.rst
```diff
@@ -3039,17 +3039,32 @@ Throwing an exception with :ref:`tag address <syntax-tagaddr>` :math:`a`
...
+
+1. Let :math:`\val^\ast` be the values on the top of the stack.
+
+2. Pop the values :math:`\val^\ast` from the stack.
+
+3. Assert: due to :ref:`validation <valid-instr-seq>`, an administrative instruction :math:`\CAUGHTadm\{a~\val_0^\ast\}` is now on the top of the stack.
+
+4. Pop the |CAUGHTadm| from the stack.
+
+5. Push :math:`\val^\ast` back to the stack.
```
aheejin, line 3055: Other similar sections ('Exiting an exception handler' and 'Exiting `instr` with label `L`') use `val^m`. Is there any reason we use `val^*` here? For example, our 'Exiting an exception handler' section's prose is:
```
Exiting an exception handler                                                     
............................                                                     

When the end of a :ref:`try <syntax-try>` instruction is reached without a jump, exception, or trap, then the following steps are performed.

1. Let :math:`m` be the number of values on the top of the stack.

2. Pop the values :math:`\val^m` from the stack.

3. Assert: due to :ref:`validation <valid-instr-seq>`, the handler :math:`H` is now on the top of the stack.

4. Pop the handler from the stack.

5. Push :math:`\val^m` back to the stack.

6. Jump to the position after the |END| of the administrative instruction associated with the handler :math:`H`.
```

ioannad: @aheejin  "Exiting instr with label L" uses `val^m` because it's relevant to the rule, in that labels are of the form `label_m`. Since `caught` does not have such a subscript, I used `*`, and I think it should be the same for "Exiting an exception handler". 
 So I would be more inclined to change to `val^*` in "Exiting an exception handler" and leaving this `val*` here as it is, but I would go with whatever you prefer. WDYT?

aheejin (approved):
Hmm I see, thanks for the explanation!

document/core/exec/instructions.rst
```diff
@@ -3039,17 +3039,32 @@ Throwing an exception with :ref:`tag address <syntax-tagaddr>` :math:`a`
 
 .. _exec-caughtadm:
 
-Holding a caught exception with |CAUGHTadm|
-...........................................
+Exiting a |CAUGHTadm|
```
rossberg, line 3042:
```suggestion
Exiting a catch clause |CAUGHTadm|
```

document/core/exec/instructions.rst
```diff
@@ -3039,17 +3039,32 @@ Throwing an exception with :ref:`tag address <syntax-tagaddr>` :math:`a`
...
+.....................
+
+When the |END| of a |CAUGHTadm|, is reached without a jump, exception, or trap, then the following steps are performed.
```
rossberg, line 3045:
```suggestion
When the |END| of a catch clause |CAUGHTadm| is reached without a jump, exception, or trap, then the following steps are performed.
```

document/core/exec/instructions.rst
```diff
@@ -3039,17 +3039,32 @@ Throwing an exception with :ref:`tag address <syntax-tagaddr>` :math:`a`
...
+
+1. Let :math:`\val^\ast` be the values on the top of the stack.
+
+2. Pop the values :math:`\val^\ast` from the stack.
+
+3. Assert: due to :ref:`validation <valid-instr-seq>`, an administrative instruction :math:`\CAUGHTadm\{a~\val_0^\ast\}` is now on the top of the stack.
+
+4. Pop the |CAUGHTadm| from the stack.
```
rossberg, line 3053:
```suggestion
4. Pop the catch clause |CAUGHTadm| from the stack.
```

document/core/exec/instructions.rst
ioannad: Made PR #240 with this and the other changes requested.

PR #228, opened 2022-09-08
Add prose for typing rule of CATCHadm

ioannad:

Also adjusted notation to match the one in the [formal overview document](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions-formal-overview.md).

### DIFF
--- a/document/core/appendix/properties.rst
+++ b/document/core/appendix/properties.rst
@@ -605,26 +605,26 @@ To that end, all previous typing judgements :math:`C \vdash \X{prop}` are genera
 
 .. index:: catch, throw context
 
-:math:`\CATCHadm\{\tagaddr^?~\instr_2^\ast\}^\ast~\instr_1^\ast~\END`
+:math:`\CATCHadm\{\tagaddr^?~\instr_1^\ast\}^\ast~\instr_2^\ast~\END`
 .....................................................................
 
 * Let :math:`C'` be the same :ref:`context <context>` as :math:`C`, but with the :ref:`label type <syntax-labeltype>` :math:`[t_2^\ast]` prepended to the |CLABELS| vector.
 
 * Under context :math:`C'`,
-  the instruction sequence :math:`\instr_1^\ast` must be :ref:`valid <valid-instr-seq>` with type :math:`[] \to [t_2^\ast]`.
+  the instruction sequence :math:`\instr_2^\ast` must be :ref:`valid <valid-instr-seq>` with type :math:`[] \to [t_2^\ast]`.
 
 * Let :math:`C''` be the same :ref:`context <context>` as :math:`C`, but with the :ref:`label type <syntax-labeltype>` :math:`(\LCATCH~[t_2^\ast])` prepended to the |CLABELS| vector.
 
 * Under context :math:`C''`,
-  for every :math:`\tagaddr^?` and associated instruction sequence :math:`\instr_2^\ast`:
+  for every :math:`\tagaddr^?` and associated instruction sequence :math:`\instr_1^\ast`:
 
-  * If :math:`\tagaddr^? = \epsilon`, then :math:`\instr_2^\ast` must be :ref:`valid <valid-instr-seq>` with type :math:`[] \to [t_2^\ast]`.
+  * If :math:`\tagaddr^? = \epsilon`, then :math:`\instr_1^\ast` must be :ref:`valid <valid-instr-seq>` with type :math:`[] \to [t_2^\ast]`.
 
   * Else:
 
     * The :ref:`external tag value <syntax-externval>` :math:`\EVTAG~\tagaddr` must be :ref:`valid <valid-externval-tag>` with some :ref:`external tag type <syntax-externtype>` :math:`\ETTAG~[t_1^\ast] \to []`.
 
-    * The instruction sequence :math:`\instr_2^\ast` must be :ref:`valid <valid-instr-seq>` with type :math:`[t_1^\ast] \to [t_2^\ast]`.
+    * The instruction sequence :math:`\instr_1^\ast` must be :ref:`valid <valid-instr-seq>` with type :math:`[t_1^\ast] \to [t_2^\ast]`.
 
 * Then the compound instruction is valid under context :math:`C'` with type :math:`[] \to [t_2^\ast]`.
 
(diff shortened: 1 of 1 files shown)


document/core/appendix/properties.rst
```diff
@@ -605,21 +605,38 @@ To that end, all previous typing judgements :math:`C \vdash \X{prop}` are genera
...
 
 .. math::
    \frac{
      \begin{array}{@{}c@{}}
-     ((S \vdashexternval \EVTAG~\tagaddr : \ETTAG~[t'^\ast]\to[])^? \\
-     ~~S; C,\CLABELS\,(\LCATCH~[t^\ast]) \vdashinstrseq \instr'^\ast : [(t'^\ast)^?] \to [t^\ast])^\ast \\
-     S; C,\CLABELS\,[t^\ast] \vdashinstrseq \instr^\ast : [] \to [t^\ast] \\
+     ((S \vdashexternval \EVTAG~\tagaddr : \ETTAG~[t_1^\ast]\to[])^? \\
+     ~~S; C,\CLABELS\,(\LCATCH~[t_2^\ast]) \vdashinstrseq \instr_2^\ast : [(t_1^\ast)^?] \to [t_2^\ast])^\ast \\
+     S; C,\CLABELS\,[t_2^\ast] \vdashinstrseq \instr_1^\ast : [] \to [t_2^\ast] \\
```
aheejin: Our `try`-`catch`-`catch_all` rule has a `blocktype` `t1->t2`, so it assumes there can be an input values. (Here `t1` means a tag type, which is a different thing) Do we not assume there can be a block input type, or it is not relevant here?

ioannad: An instruction `try-catch` may take input types `val^*:t_1` but it reduces to `catch{....}* val^* instr* end`, so any input types are put in the `CATCHadm` which ends up having type `[] -> [t_2^*]`. It's similar for the [typing of the administrative label](https://webassembly.github.io/spec/core/appendix/properties.html#xref-exec-runtime-syntax-label-mathsf-label-n-xref-syntax-instructions-syntax-instr-mathit-instr-0-ast-xref-syntax-instructions-syntax-instr-mathit-instr-ast-xref-syntax-instructions-syntax-instr-control-mathsf-end). So when a [block reduces to a label](https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-control-mathsf-block-xref-syntax-instructions-syntax-blocktype-mathit-blocktype-xref-syntax-instructions-syntax-instr-mathit-instr-ast-xref-syntax-instructions-syntax-instr-control-mathsf-end), it incorporates any input values with the instructions inside the label. Does this clarify your question?

aheejin: Sorry, this can be a basic question, but I don't think I understand that administrative `label` rule either. It is written as this:
image

To my knowledge, `label_n {instr_0*} instr* end` means `instr*` runs and produces `t1^n` values, and after that the `label` is popped and `instr_0*` runs as a continuation, taking those `t1^n` values produced by `instr*`, and produces `t2*` values. Right? So an example of original wasm that would have reduced to this can be something like
```wasm
block (t1^n -> t2*)
  instr*
end
instr_0* ;; this takes t1^n and outputs t2*
```

But the second rule (= third sentence) says, `instr*` should produce `t2*` values, not `t1^n` values. Also It runs under the context `C'`, which has `t1^n` values as its return type, but the `block`'s return type is not `t1^n` but `t2*`, no?  Why does `instr*`, which is inside the `block` have to run under the context that the result label type is `t1^n`, which is the `block`'s _input type_? Shouldn't it run under the the context the result label type is the `block`'s _output type_?

aheejin (approved):
Aside form my question (https://github.com/WebAssembly/exception-handling/pull/228#discussion_r967730878), which is about the existing `label` rule, I think this LGTM.

document/core/appendix/properties.rst
```diff
@@ -605,21 +605,38 @@ To that end, all previous typing judgements :math:`C \vdash \X{prop}` are genera
...
    \end{array}
    }{
-     S; C,\CLABELS\,[t^\ast] \vdashadmininstr \CATCHadm\{\tagaddr^?~{\instr'}^\ast\}^\ast~\instr^\ast~\END : [] \to [t^\ast]
+     S; C,\CLABELS\,[t_2^\ast] \vdashadmininstr \CATCHadm\{\tagaddr^?~{\instr_2}^\ast\}^\ast~\instr_1^\ast~\END : [] \to [t_2^\ast]
```
rossberg: It's a bit confusing that `t2*` is now the type of `instr1*`. Can we switch that around?

document/core/appendix/properties.rst
rossberg: @aheejin, the continuation `instr0*` recorded in the label is the one for when the branch is taken, not for regular completion. So it is unrelated to the `t2*` in the rule, which is the type need for regular completion. These happen to be the same for block labels but not for loop labels. See the reduction rules for both:
```
val1* block (t1*->t2*) instr* end --> label{} val1* instr* end    (label of type t2*, result is t2*)
val1* loop (t1*->t2*) instr* end --> label{instr*} val1* instr* end    (label of type t1*, result is t2*)
```

document/core/appendix/properties.rst
```diff
@@ -605,21 +605,38 @@ To that end, all previous typing judgements :math:`C \vdash \X{prop}` are genera
...
 .. index:: catch, throw context
 
-:math:`\CATCHadm\{\tagaddr^?~\instr'^\ast\}^\ast~\instr^\ast~\END`
-..................................................................
+:math:`\CATCHadm\{\tagaddr^?~\instr_2^\ast\}^\ast~\instr_1^\ast~\END`
+.....................................................................
 
-.. todo::
-   Add prose.
+* Let :math:`C'` be the same :ref:`context <context>` as :math:`C`, but with the label :math:`[t_2^\ast]` prepended to the |CLABELS| vector.
```
rossberg:
```suggestion
* Let :math:`C'` be the same :ref:`context <context>` as :math:`C`, but with the :ref:`label type <syntax-labeltype>` :math:`[t_2^\ast]` prepended to the |CLABELS| vector.
```

document/core/appendix/properties.rst
```diff
@@ -605,21 +605,38 @@ To that end, all previous typing judgements :math:`C \vdash \X{prop}` are genera
...
+
+* Under context :math:`C'`,
+  the instruction sequence :math:`\instr_1^\ast` must be :ref:`valid <valid-instr-seq>` with type :math:`[] \to [t_2^\ast]`.
+
+* Let :math:`C''` be the same :ref:`context <context>` as :math:`C`, but with the label :math:`(\LCATCH~[t_2^\ast])` prepended to the |CLABELS| vector.
```
rossberg:
```suggestion
* Let :math:`C''` be the same :ref:`context <context>` as :math:`C`, but with the :ref:`label type <syntax-labeltype>` :math:`(\LCATCH~[t_2^\ast])` prepended to the |CLABELS| vector.
```

document/core/appendix/properties.rst
```diff
@@ -605,21 +605,38 @@ To that end, all previous typing judgements :math:`C \vdash \X{prop}` are genera
...
+
+* Under context :math:`C''`,
+  for every :math:`\tagaddr^?` and associated instruction sequence :math:`\instr_2^\ast`:
+
+  * If :math:`\tagaddr^? = \epsilon` then :math:`\instr_2^\ast` must be :ref:`valid <valid-instr-seq>` with type :math:`[] \to [t_2^\ast]`.
```
rossberg:
```suggestion
  * If :math:`\tagaddr^? = \epsilon`, then :math:`\instr_2^\ast` must be :ref:`valid <valid-instr-seq>` with type :math:`[] \to [t_2^\ast]`.
```

document/core/appendix/properties.rst
ioannad: done.

rossberg (approved):

PR #233, opened 2022-09-13
Update `br` stack-popping rule with EH

aheejin:
I think this should be also updated to include handlers and `caught` to match the `rethrow` rule in #225.


ioannad (approved):
Looks good, thanks!

PR #237, opened 2022-09-16
Add prose for typing rule of try-delegate.

ioannad:
Also slightly adjusted notation to match the formal overview.

<del>Why WIP:

<del>The prose is finished, but while trying to add explanatory notes, I couldn't find a reference in the spec to the validation label that the frame pushes to the context.

EDIT:
WIP is no longer necessary after addressing review comments.

### DIFF
--- a/document/core/valid/instructions.rst
+++ b/document/core/valid/instructions.rst
@@ -1415,8 +1415,14 @@ Control Instructions
 :math:`\TRY~\blocktype~\instr^\ast~\DELEGATE~l`
 ...............................................
 
-.. todo::
-   Add prose.
+* The label :math:`C.\CLABELS[l]` must be defined in the context.
+
+* The :ref:`block type <syntax-blocktype>` must be :ref:`valid <valid-blocktype>` as some :ref:`function type <syntax-functype>` :math:`[t_1^\ast] \to [t_2^\ast]`.
+
+* Let :math:`C'` be the same :ref:`context <context>` as :math:`C`, but with the :ref:`result type <syntax-resulttype>` :math:`[t_2^\ast]` prepended to the |CLABELS| vector.
+
+* Under context :math:`C'`,
+  the instruction sequence :math:`\instr^\ast` must be :ref:`valid <valid-instr-seq>` with type :math:`[t_1^\ast] \to [t_2^\ast]`.
 
 .. math::
    \frac{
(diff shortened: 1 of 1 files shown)


document/core/valid/instructions.rst
```diff
@@ -1415,20 +1415,33 @@ Control Instructions
...
    C \vdashinstrseq \TRY~\blocktype~\instr^\ast~\DELEGATE~l : [t_1^\ast]\to[t_2^\ast]
    }
 
+.. note::
+   The :ref:`notation <notation-extend>` :math:`C,\CLABELS\, [t^\ast]` inserts the new label type at index :math:`0`, shifting all others.
+
+   The :ref:`label index <syntax-labelidx>` after |DELEGATE| refers to a label surrounding the :ref:`try-delegate <syntax-try-delegate>` instruction. The furthest label it can refer to is the label inserted by the frame. For example, :math:`\TRY~\dots~\DELEGATE~0` may appear without any explicitly surrounding block, in which case the label 0 refers to the label of the frame.
+
+.. todo::
+   Add references/links to "the label of the frame": Where is this label introduced?
```
rossberg: Frames represent calls and are unrelated to labels. A function body implicitly is a block, that's why it has an outermost label.

But in fact, I'd remove this entire note. It basically just explains again how labels and functions work, which is not specific to delegate. If we did that here, then we would have to do that for every br instruction.

document/core/valid/instructions.rst
```diff
@@ -1415,20 +1415,33 @@ Control Instructions
...
      C,\CLABELS\,[t_2^\ast] \vdashinstrseq \instr^\ast : [t_1^\ast]\to[t_2^\ast]
      \qquad
-     C.\CLABELS[l] = [t^\ast]
+     C.\CLABELS[l] = [t_0^\ast]
    }{
    C \vdashinstrseq \TRY~\blocktype~\instr^\ast~\DELEGATE~l : [t_1^\ast]\to[t_2^\ast]
    }
 
+.. note::
+   The :ref:`notation <notation-extend>` :math:`C,\CLABELS\, [t^\ast]` inserts the new label type at index :math:`0`, shifting all others.
```
rossberg: Can we reuse the wording used for br and friends?

ioannad: Ok, switching to the wording of br.
FWIW I had copy pasted this from the note under if-else.

ioannad: done.

document/core/valid/instructions.rst
ioannad: done.

ioannad:
This should be now ready for merging, pending approval.

rossberg (approved):

PR #222, opened 2022-09-07
Add prose for typing rule of try-catch-catch_all.

ioannad:
Fixes related .. todo:: section.

Also changed notation to match the [formal overview document](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions-formal-overview.md).

PR #235, opened 2022-09-16
Add prose for typing rule of DELEGATEadm

ioannad:
Also switched to the current version of the formal rule, based on the formal overview document.

PR #241, opened 2022-10-21
Address additional review of #222, also fixing label types everywhere.

ioannad:
This addresses additional review comments to PR #222, that were made after it was merged. The last review comment in the discussion suggests to adjust all validation labels to use label types instead of just result types. Should address all occurrences of validation labels.

Additionally adds a boolean catch_label to control frames in the validation algorithm, and some related functionality, fixing the cases for opcodes `catch` and `catch_all`.

### DIFF
--- a/document/core/appendix/algorithm.rst
+++ b/document/core/appendix/algorithm.rst
@@ -127,7 +127,7 @@ The control stack is likewise manipulated through auxiliary functions:
      return (if frame.opcode == loop then frame.start_types else frame.end_types)
 
    func is_catch(frame : ctrl_frame) : bool =
-     return frame.catch_label
+     return frame.opcode = catch || frame.opcode = catch_all
 
    func set_catch(catch : bool) =
      error_if(ctrls.is_empty())
--- a/document/core/appendix/algorithm.rst
+++ b/document/core/appendix/algorithm.rst
@@ -16,7 +16,7 @@ Consequently, it can be integrated directly into a decoder.
 The algorithm is expressed in typed pseudo code whose semantics is intended to be self-explanatory.
 
 
-.. index:: value type, label type, stack, label, frame, instruction
+.. index:: value type, stack, label, frame, instruction
 
 Data Structures
 ~~~~~~~~~~~~~~~
@@ -51,7 +51,6 @@ the latter surrounding :ref:`structured control instructions <syntax-instr-contr
      end_types : list(val_type)
      height : nat
      unreachable : bool
-     catch_label : bool
    }
 
 For each value, the value stack records its :ref:`value type <syntax-valtype>`, or :code:`Unknown` when the type is not known.
@@ -123,15 +122,8 @@ The control stack is likewise manipulated through auxiliary functions:
      ctrls.pop()
      return frame
 
-   func label_val_types(frame : ctrl_frame) : list(val_type) =
+   func label_types(frame : ctrl_frame) : list(val_type) =
      return (if frame.opcode == loop then frame.start_types else frame.end_types)
-
-   func is_catch(frame : ctrl_frame) : bool =
-     return frame.opcode = catch || frame.opcode = catch_all
-
-   func set_catch(catch : bool) =
-     error_if(ctrls.is_empty())
-     ctrls[0].catch_label := catch
 
    func unreachable() =
      vals.resize(ctrls[0].height)
@@ -144,9 +136,7 @@ Popping a frame first checks that the control stack is not empty.
 It then verifies that the operand stack contains the right types of values expected at the end of the exited block and pops them off the operand stack.
 Afterwards, it checks that the stack has shrunk back to its initial height.
 
-The result type of the :ref:`label <syntax-label>` associated with a control frame is either that of the stack at the start or the end of the frame, determined by the opcode that it originates from.
-
-Whether a label has :ref:`label type <syntax-labeltype>` with present |LCATCH| is stored with the boolean :math:`catch_label` of the frame, which can be checked with :math:`is_catch`, or modified in the top control frame with :math:`set_catch`.
+The type of the :ref:`label <syntax-label>` associated with a control frame is either that of the stack at the start or the end of the frame, determined by the opcode that it originates from.
 
 Finally, the current frame can be marked as unreachable.
 In that case, all existing operand types are purged from the value stack, in order to allow for the :ref:`stack-polymorphism <polymorphism>` logic in :code:`pop_val` to take effect.
@@ -229,34 +219,32 @@ Other instructions are checked in a similar manner.
          error_if(frame.opcode =/= try || frame.opcode =/= catch)
          let params = tags[x].type.params
          push_ctrl(catch, params , frame.end_types)
-         if (frame.opcode =/= try) then set_catch(true)
 
        case (catch_all)
          let frame = pop_ctrl()
          error_if(frame.opcode =/= try || frame.opcode =/= catch)
          push_ctrl(catch_all, [], frame.end_types)
-         if (frame.opcode =/= try) then set_catch(true)
 
        case (br n)
          error_if(ctrls.size() < n)
-         pop_vals(label_val_types(ctrls[n]))
+         pop_vals(label_types(ctrls[n]))
          unreachable()
 
        case (br_if n)
          error_if(ctrls.size() < n)
          pop_val(I32)
-         pop_vals(label_val_types(ctrls[n]))
-         push_vals(label_val_types(ctrls[n]))
+         pop_vals(label_types(ctrls[n]))
+         push_vals(label_types(ctrls[n]))
 
        case (br_table n* m)
          pop_val(I32)
          error_if(ctrls.size() < m)
-         let arity = label_val_types(ctrls[m]).size()
+         let arity = label_types(ctrls[m]).size()
          foreach (n in n*)
            error_if(ctrls.size() < n)
-           error_if(label_val_types(ctrls[n]).size() =/= arity)
-           push_vals(pop_vals(label_val_types(ctrls[n])))
-         pop_vals(label_val_types(ctrls[m]))
+           error_if(label_types(ctrls[n]).size() =/= arity)
+           push_vals(pop_vals(label_types(ctrls[n])))
+         pop_vals(label_types(ctrls[m]))
          unreachable()
 
 .. todo::
--- a/document/core/appendix/algorithm.rst
+++ b/document/core/appendix/algorithm.rst
@@ -16,7 +16,7 @@ Consequently, it can be integrated directly into a decoder.
 The algorithm is expressed in typed pseudo code whose semantics is intended to be self-explanatory.
 
 
-.. index:: value type, stack, label, frame, instruction
+.. index:: value type, label type, stack, label, frame, instruction
 
 Data Structures
 ~~~~~~~~~~~~~~~
@@ -51,6 +51,7 @@ the latter surrounding :ref:`structured control instructions <syntax-instr-contr
      end_types : list(val_type)
      height : nat
      unreachable : bool
+     catch_label : bool
    }
 
 For each value, the value stack records its :ref:`value type <syntax-valtype>`, or :code:`Unknown` when the type is not known.
@@ -122,8 +123,15 @@ The control stack is likewise manipulated through auxiliary functions:
      ctrls.pop()
      return frame
 
-   func label_types(frame : ctrl_frame) : list(val_types) =
+   func label_val_types(frame : ctrl_frame) : list(val_type) =
      return (if frame.opcode == loop then frame.start_types else frame.end_types)
+
+   func is_catch(frame : ctrl_frame) : bool =
+     return frame.catch_label
+
+   func set_catch(catch : bool) =
+     error_if(ctrls.is_empty())
+     ctrls[0].catch_label := catch
 
    func unreachable() =
      vals.resize(ctrls[0].height)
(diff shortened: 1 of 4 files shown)


document/core/appendix/algorithm.rst
```diff
@@ -122,9 +123,16 @@ The control stack is likewise manipulated through auxiliary functions:
      ctrls.pop()
      return frame
 
-   func label_types(frame : ctrl_frame) : list(val_types) =
+   func label_val_types(frame : ctrl_frame) : list(val_type) =
```
rossberg: Why rename the function?

document/core/appendix/algorithm.rst
```diff
@@ -51,6 +51,7 @@ the latter surrounding :ref:`structured control instructions <syntax-instr-contr
      end_types : list(val_type)
      height : nat
      unreachable : bool
+     catch_label : bool
```
rossberg: I don't think we need this flag, given that we already store the opcode.

document/core/appendix/algorithm.rst
```diff
@@ -122,9 +123,16 @@ The control stack is likewise manipulated through auxiliary functions:
      ctrls.pop()
      return frame
 
-   func label_types(frame : ctrl_frame) : list(val_types) =
+   func label_val_types(frame : ctrl_frame) : list(val_type) =
      return (if frame.opcode == loop then frame.start_types else frame.end_types)
 
+   func is_catch(frame : ctrl_frame) : bool =
+     return frame.catch_label
```
```suggestion
     return frame.opcode = catch || frame.opcode = catch_all
```
We could also simply inline this condition. But this function does not currently seem to be used anywhere?

document/core/appendix/algorithm.rst
```diff
@@ -122,9 +123,16 @@ The control stack is likewise manipulated through auxiliary functions:
...
+
+   func set_catch(catch : bool) =
+     error_if(ctrls.is_empty())
+     ctrls[0].catch_label := catch
```
rossberg: Remove.

document/core/appendix/algorithm.rst
```diff
@@ -136,7 +144,9 @@ Popping a frame first checks that the control stack is not empty.
 It then verifies that the operand stack contains the right types of values expected at the end of the exited block and pops them off the operand stack.
 Afterwards, it checks that the stack has shrunk back to its initial height.
 
-The type of the :ref:`label <syntax-label>` associated with a control frame is either that of the stack at the start or the end of the frame, determined by the opcode that it originates from.
+The result type of the :ref:`label <syntax-label>` associated with a control frame is either that of the stack at the start or the end of the frame, determined by the opcode that it originates from.
+
+Whether a label has :ref:`label type <syntax-labeltype>` with present |LCATCH| is stored with the boolean :math:`catch_label` of the frame, which can be checked with :math:`is_catch`, or modified in the top control frame with :math:`set_catch`.
```
rossberg: Remove.

document/core/appendix/algorithm.rst
```diff
@@ -219,32 +229,34 @@ Other instructions are checked in a similar manner.
          error_if(frame.opcode =/= try || frame.opcode =/= catch)
          let params = tags[x].type.params
          push_ctrl(catch, params , frame.end_types)
+         if (frame.opcode =/= try) then set_catch(true)
```
rossberg: Remove.

document/core/appendix/algorithm.rst
```diff
@@ -219,32 +229,34 @@ Other instructions are checked in a similar manner.
...
 
        case (catch_all)
          let frame = pop_ctrl()
          error_if(frame.opcode =/= try || frame.opcode =/= catch)
          push_ctrl(catch_all, [], frame.end_types)
+         if (frame.opcode =/= try) then set_catch(true)
```
rossberg: Remove.

document/core/appendix/properties.rst
```diff
@@ -686,9 +686,9 @@ To that end, all previous typing judgements :math:`C \vdash \X{prop}` are genera
      \qquad
      (val : t_0)^\ast
      \qquad
-     S; C,\CLABELS\,(\LCATCH~[t^\ast]) \vdashinstrseq \instr^\ast : [] \to [t^\ast]
+     S; C,\CLABELS[0]:=(\LCATCH~[t^\ast]) \vdashinstrseq \instr^\ast : [] \to [t^\ast]
    }{
-     S; C,\CLABELS\,[t^\ast] \vdashadmininstr \CAUGHTadm\{\tagaddr~\val^\ast\}~\instr^\ast~\END : [] \to [t^\ast]
```
rossberg: Why change this formulation?

document/core/valid/instructions.rst
```diff
@@ -1458,11 +1460,11 @@ Control Instructions
 :math:`\RETHROW~l`
 ..................
 
-* The label :math:`C.\CLABELS[l]` must be defined in the context.
+* The label type :math:`C.\CLABELS[l]` must be defined in the context.
```
rossberg: Actually, it _is_ the label that must be defined (_with_ a type). So I'd revert this (similarly several occurrences below).

document/core/appendix/algorithm.rst
ioannad: It was in the spirit of now calling [such labeltypes](https://webassembly.github.io/exception-handling/core/valid/conventions.html#syntax-labeltype) "label types" , as requested [in this comment](https://github.com/WebAssembly/exception-handling/pull/222#issuecomment-1249079017) but I see now that I may have overdone it. :) Reverting/removing as suggested.

document/core/appendix/properties.rst
ioannad: This does seem misplaced. I was trying to follow the formulation for the typing of ` \DELEGATEadm` in [the formal overview](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions-formal-overview.md#typing-rules-for-administrative-instructions), but not only this change is misplaced here, but it seems unnecessary to change this in the appendix, given the [latest comment](https://github.com/WebAssembly/exception-handling/pull/235#pullrequestreview-1181041527) on the  `DELEGATEadm` typing rule. 

Reverting this change.

document/core/valid/instructions.rst
ioannad: Done, for the other occurrences as well.

document/core/appendix/properties.rst
ioannad: done

document/core/appendix/algorithm.rst
ioannad: done

document/core/appendix/algorithm.rst
ioannad: done

document/core/appendix/algorithm.rst
ioannad: done

document/core/appendix/algorithm.rst
ioannad: done

document/core/appendix/algorithm.rst
ioannad: removed

document/core/appendix/algorithm.rst
ioannad: done

document/core/appendix/properties.rst
```diff
@@ -670,25 +670,24 @@ To that end, all previous typing judgements :math:`C \vdash \X{prop}` are genera
 
 * The :ref:`values <syntax-val>` :math:`\val^\ast` must be of type :math:`[t_0^\ast]`.
 
-* The label type :math:`C.\CLABELS[0]` must be defined in the context.
-
-* Let :math:`C'` be the same :ref:`context <context>` as :math:`C`, but with the :ref:`label type <syntax-labeltype>` :math:`C.\CLABELS[0]` modified to :math:`(\LCATCH~[t^\ast])` in the |CLABELS| vector.
+* Let :math:`C'` be the same :ref:`context <context>` as :math:`C`, but with the label :math:`(\LCATCH~[t^\ast])` prepended to the |CLABELS| vector.
```
rossberg: C is an (implicit) _input_ to the rules, so you'll have to write the prose the other way round, i.e., first check that the last entry is `[t*]`, then define C' to be C with that entry updated. (That means that the C in the prose and the C in the formal rule denote different contexts, but I think that's fine.)

document/core/valid/instructions.rst
```diff
@@ -1368,23 +1368,25 @@ Control Instructions
...
 * Under context :math:`C'`,
   the instruction sequence :math:`\instr_1^\ast` must be :ref:`valid <valid-instr-seq>` with type :math:`[t_1^\ast] \to [t_2^\ast]`.
 
-* Let :math:`C''` be the same :ref:`context <context>` as :math:`C`, but with the :ref:`label <exec-label>` :math:`\LCATCH~[t_2^\ast]` prepended to the |CLABELS| vector.
+* Let :math:`C''` be the same :ref:`context <context>` as :math:`C`, but with the :ref:`label type <syntax-labeltype>` :math:`\LCATCH~[t_2^\ast]` prepended to the |CLABELS| vector.
 
-* For every :math:`(\CATCH~x~\instr_2^\ast)`:
+* For every :math:`x_i` and :math:`\instr_{2i}^\ast` in :math:`(\CATCH~x~\instr_2^\ast)^\ast`:
 
   * The tag :math:`C.\CTAGS[x]` must be defined in the context :math:`C`.
```
aheejin:
```suggestion
  * The tag :math:`C.\CTAGS[x_i]` must be defined in the context :math:`C`.
```

document/core/valid/instructions.rst
```diff
@@ -1368,23 +1368,25 @@ Control Instructions
...
 
-  * Let :math:`[t^\ast] \to []` be its :ref:`tag type <syntax-tagtype>`.
+  * Let :math:`[t_{3i}^\ast] \to [t_{4i}^\ast]` be the :ref:`tag type <syntax-tagtype>` :math:`C.\CTAGS[x]`.
```
aheejin:
```suggestion
  * Let :math:`[t_{3i}^\ast] \to [t_{4i}^\ast]` be the :ref:`tag type <syntax-tagtype>` :math:`C.\CTAGS[x_i]`.
```

document/core/appendix/properties.rst
aheejin:
```suggestion
* Let :math:`C'` be the same :ref:`context <context>` as :math:`C`, but with the label type :math:`(\LCATCH~[t^\ast])` prepended to the |CLABELS| vector.
```
Isn't `type` missing? All other similar clauses have it.

@rossberg Not sure if I understand. Also I wonder if making `C`'s meaning different from that in the equation will be confusing.. Can we change the equation to use the same symbol?

aheejin (commented):
The `caught` validation rule is shown like this now:
image

The numerator says `labels(catch [t*])`, but the denominator says `labels[t*]`. Should we change the denominator to `labels(catch [t*])` as well?

rossberg:
@aheejin:
> Should we change the denominator to `labels(catch [t*])` as well?

The rule is intended as is, the purpose being that it refines the existing label binding to a catch label, enabling its use for rethrow in the premise.

document/core/appendix/properties.rst
rossberg: Yes, we could use C' in the formal rule, though the tick would just be noise there. In general, I have opted to not artificially complicate the formal rules for the sake of the prose, though this case is minor enough.

Adding `type` sounds okay be me.

aheejin: Then can we change the symbol in the prose?

aheejin: Actually I'm not sure if I understood your comment correctly:
> C is an (implicit) _input_ to the rules, so you'll have to write the prose the other way round, i.e., first check that the last entry is `[t*]`, then define C' to be C with that entry updated. 

Can you elaborate a little more on what this means? And if the symbols don't cause much confusion, that might be fine.

rossberg: In the formal rules, the context is always explicit. We just call it C, or something else, and we can even "pattern match" it in the rules, which is what happens in the rule for caught. This implicitly imposes a constraint, because the rule is only applicable if the context has that specific form.

In the prose, the context is treated as a parameter to the "algorithm". And we keep it implicit to avoid more excessive verbosity. That implicit context is available under the fixed name C. This convention is described [here](https://webassembly.github.io/spec/core/valid/conventions.html#prose-notation). A constraint on that context (e.g., the presence of a certain entry) then has to be checked explicitly.

document/core/valid/instructions.rst
aheejin: Hmm, will fix these myself in a separate PR. Nevermind

document/core/valid/instructions.rst
aheejin: Hmm, will fix these myself in a separate PR. Nevermind

document/core/valid/instructions.rst
aheejin: Sorry, please ignore my previous (deleted) comment saying that I will fix this myself in a separate PR. I was confused and I thought this was already merged, and I didn't want to bother you to create another PR just for this, but this hasn't been merged after all...

document/core/appendix/properties.rst
ioannad: Moved these changes  to PR #244, reverting these changes to whatever the document had previously.

ioannad: Please ignore the last two commits (the second reverting the first) - the changes to the typing of CAUGHTadm were already reverted. 

Changes to the typing of CAUGHTadm are now done in PR #244.

ioannad:
@aheejin @rossberg 
All review comments should now be addressed here, except those involving the prose for the typing of `CAUGHTadm`: these changes are now moved to PR #244 as [mentioned in my previous comment](https://github.com/WebAssembly/exception-handling/pull/241#discussion_r1032777661).

rossberg (approved):

PR #245, opened 2022-12-12
Update Exceptions.md

ouillie:
Looks like an off-by-one.


aheejin:
No it's not. See the next paragraph:
> Note that `rethrow 2` is not allowed because it does not refer to a `try` label from within its catch block. Rather, it references a `block` instruction, so it will result in a validation failure.

I guess https://github.com/WebAssembly/exception-handling/issues/146#issuecomment-777714491 can possibly help understanding.

PR #251, opened 2023-02-16
[js-api] Rename variables to be consistent (NFC)

aheejin:
This renames variables to be consistent among exception JS API tests. Now `WebAssembly.Tag`s are named `tag` or prefixed with it and `WebAssembly.Exception`s are named `exn` or prefixed with it.

Currently some `WebAssembly.Tag`s are named `exn` because `Tag` was used to call `Exception` in the earlier version of the JS API, which this PR fixes.


Ms2ger (approved):
ok

PR #254, opened 2023-02-16
Adds validation for exception results.

ioannad:
Uses the "current" version of exception results, as it will be changed by #226.

### DIFF
--- a/document/core/appendix/properties.rst
+++ b/document/core/appendix/properties.rst
@@ -62,15 +62,15 @@ Results
 :ref:`Results <syntax-result>` :math:`\val^\ast~(\THROWadm~\tagaddr)`
 .....................................................................
 
-* The :ref:`tag address <syntax-tagaddr>` :math:`\tagaddr` must be in :math:`\moduleinst.\MITAGS`.
-
 * The :ref:`external value <syntax-externval>` :math:`\EVTAG~\tagaddr` must be :ref:`valid <valid-externval-tag>` with some :ref:`external type <syntax-externtype>` :math:`\ETTAG~\tagtype`.
 
-* Let :math:`[t_1^\ast]\to[t_2^\ast]` be the :ref:`tag type <syntax-tagtype>` |tagtype|.
+* Let :math:`[t^\ast]\to[{t'}^\ast]` be the :ref:`tag type <syntax-tagtype>` |tagtype|.
 
-* The values :math:`\val^\ast` must be :ref:`valid <valid-val>` with types :math:`[t_1^\ast]`.
+* For each :ref:`value <syntax-val>` :math:`\val_i` in :math:`\val^\ast`:
 
-* Then the result is valid with :ref:`result type <syntax-resulttype>` :math:`[t_2^\ast]`, for any sequence :math:`t_2^\ast` of :ref:`value types <syntax-valtype>`.
+  * The value :math:`\val_i` must be :ref:`valid <valid-val>` with :ref:`value type <syntax-valtype>` :math:`t_i`.
+
+* Then the result is valid with :ref:`result type <syntax-resulttype>` :math:`[{t'}^\ast]`, for any sequence :math:`{t'}^\ast` of :ref:`value types <syntax-valtype>`.
 
 
 .. math::
--- a/document/core/appendix/properties.rst
+++ b/document/core/appendix/properties.rst
@@ -66,7 +66,7 @@ Results
 
 * The :ref:`external value <syntax-externval>` :math:`\EVTAG~\tagaddr` must be :ref:`valid <valid-externval-tag>` with some :ref:`external type <syntax-externtype>` :math:`\ETTAG~\tagtype`.
 
-* Let :math:`[t_1^\ast]\to[]` be the :ref:`tag type <syntax-tagtype>` |tagtype|.
+* Let :math:`[t_1^\ast]\to[t_2^\ast]` be the :ref:`tag type <syntax-tagtype>` |tagtype|.
 
 * The values :math:`\val^\ast` must be :ref:`valid <valid-val>` with types :math:`[t_1^\ast]`.
 
@@ -75,11 +75,11 @@ Results
 
 .. math::
    \frac{
-     \vdashexterntype \ETTAG~[t_1^\ast]\to[]~\ok
+     S \vdashexternval \EVTAG~\tagaddr : \ETTAG~[t^\ast] \to []
      \qquad
-     (S \vdashval \val : t_1)^\ast
+     (S \vdashval \val : t)^\ast
    }{
-     S \vdashresult \val^\ast~(\THROWadm~\tagaddr) : [t_2^\ast]
+     S \vdashresult \val^\ast~(\THROWadm~\tagaddr) : [{t'}^\ast]
    }
 
 
--- a/document/core/appendix/properties.rst
+++ b/document/core/appendix/properties.rst
@@ -58,8 +58,29 @@ Results
      S \vdashresult \TRAP : [t^\ast]
    }
 
-.. todo::
-   Add validation for exception results.
+
+:ref:`Results <syntax-result>` :math:`\val^\ast~(\THROWadm~\tagaddr)`
+.....................................................................
+
+* The :ref:`tag address <syntax-tagaddr>` :math:`\tagaddr` must be in :math:`\moduleinst.\MITAGS`.
+
+* The :ref:`external value <syntax-externval>` :math:`\EVTAG~\tagaddr` must be :ref:`valid <valid-externval-tag>` with some :ref:`external type <syntax-externtype>` :math:`\ETTAG~\tagtype`.
+
+* Let :math:`[t_1^\ast]\to[]` be the :ref:`tag type <syntax-tagtype>` |tagtype|.
+
+* The values :math:`\val^\ast` must be :ref:`valid <valid-val>` with types :math:`[t_1^\ast]`.
+
+* Then the result is valid with :ref:`result type <syntax-resulttype>` :math:`[t_2^\ast]`, for any sequence :math:`t_2^\ast` of :ref:`value types <syntax-valtype>`.
+
+
+.. math::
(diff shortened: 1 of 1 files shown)


document/core/appendix/properties.rst
```diff
@@ -58,8 +58,29 @@ Results
      S \vdashresult \TRAP : [t^\ast]
    }
 
-.. todo::
-   Add validation for exception results.
+
+:ref:`Results <syntax-result>` :math:`\val^\ast~(\THROWadm~\tagaddr)`
+.....................................................................
+
+* The :ref:`tag address <syntax-tagaddr>` :math:`\tagaddr` must be in :math:`\moduleinst.\MITAGS`.
```
rossberg: That isn't right, it may come from any module. It must only exist in the store, which is already verified by the next step.

document/core/appendix/properties.rst
```diff
@@ -58,8 +58,29 @@ Results
...
+
+* The :ref:`external value <syntax-externval>` :math:`\EVTAG~\tagaddr` must be :ref:`valid <valid-externval-tag>` with some :ref:`external type <syntax-externtype>` :math:`\ETTAG~\tagtype`.
+
+* Let :math:`[t_1^\ast]\to[]` be the :ref:`tag type <syntax-tagtype>` |tagtype|.
```
rossberg:
```suggestion
* Let :math:`[t_1^\ast]\to[t_2^\ast]` be the :ref:`tag type <syntax-tagtype>` |tagtype|.
```
We don't know upfront that its return type is empty, that requires an explicit check as well.

document/core/appendix/properties.rst
```diff
@@ -58,8 +58,29 @@ Results
...
+
+* The values :math:`\val^\ast` must be :ref:`valid <valid-val>` with types :math:`[t_1^\ast]`.
```
rossberg: The sequence of values does not form a result itself here, so we shouldn't type it against a result type. Rather use the same "loop" as in the regular result case (mirroring what the formal rule does).

document/core/appendix/properties.rst
```diff
@@ -58,8 +58,29 @@ Results
...
+
+
+.. math::
+   \frac{
+     \vdashexterntype \ETTAG~[t_1^\ast]\to[]~\ok
+     \qquad
+     (S \vdashval \val : t_1)^\ast
+   }{
+     S \vdashresult \val^\ast~(\THROWadm~\tagaddr) : [t_2^\ast]
+   }
```
rossberg:
```suggestion
   \frac{
     S \vdashexternval \EVTAG~\tagaddr : \ETTAG~[t^\ast] \to []
     \qquad
     (S \vdashval \val : t)^\ast
   }{
     S \vdashresult \val^\ast~(\THROWadm~\tagaddr) : [{t'}^\ast]
   }
```

document/core/appendix/properties.rst
ioannad: Done.

document/core/appendix/properties.rst
ioannad: Done.

document/core/appendix/properties.rst
```diff
@@ -62,24 +62,24 @@ Results
 :ref:`Results <syntax-result>` :math:`\val^\ast~(\THROWadm~\tagaddr)`
 .....................................................................
 
-* The :ref:`tag address <syntax-tagaddr>` :math:`\tagaddr` must be in :math:`\moduleinst.\MITAGS`.
-
 * The :ref:`external value <syntax-externval>` :math:`\EVTAG~\tagaddr` must be :ref:`valid <valid-externval-tag>` with some :ref:`external type <syntax-externtype>` :math:`\ETTAG~\tagtype`.
 
-* Let :math:`[t_1^\ast]\to[]` be the :ref:`tag type <syntax-tagtype>` |tagtype|.
+* Let :math:`[t^\ast]\to[{t'}^\ast]` be the :ref:`tag type <syntax-tagtype>` |tagtype|.
```
rossberg: There seems to be a step missing after this one that checks that `t'*` is empty.

document/core/appendix/properties.rst
```diff
@@ -62,24 +62,24 @@ Results
...
+
+* For each :ref:`value <syntax-val>` :math:`\val_i` in :math:`\val^\ast`:
 
-* The values :math:`\val^\ast` must be :ref:`valid <valid-val>` with types :math:`[t_1^\ast]`.
+  * The value :math:`\val_i` must be :ref:`valid <valid-val>` with :ref:`value type <syntax-valtype>` :math:`t_i`.
 
-* Then the result is valid with :ref:`result type <syntax-resulttype>` :math:`[t_2^\ast]`, for any sequence :math:`t_2^\ast` of :ref:`value types <syntax-valtype>`.
+* Then the result is valid with :ref:`result type <syntax-resulttype>` :math:`[{t'}^\ast]`, for any sequence :math:`{t'}^\ast` of :ref:`value types <syntax-valtype>`.
```
rossberg, line 71: We have a name clash, this is a different `t'*`.

document/core/appendix/properties.rst
ioannad: Added assertion.

document/core/appendix/properties.rst
ioannad: Renamed the first occurrence to `t_0^\ast`.

document/core/appendix/properties.rst
```diff
@@ -58,8 +58,31 @@ Results
...
-   Add validation for exception results.
+
+:ref:`Results <syntax-result>` :math:`\val^\ast~(\THROWadm~\tagaddr)`
+.....................................................................
+
+* The :ref:`external value <syntax-externval>` :math:`\EVTAG~\tagaddr` must be :ref:`valid <valid-externval-tag>` with some :ref:`external type <syntax-externtype>` :math:`\ETTAG~\tagtype`.
+
+* Let :math:`[t^\ast]\to[t_0^\ast]` be the :ref:`tag type <syntax-tagtype>` |tagtype|.
+
+* Assert: due to :ref:`validation <valid-externval-tag>` the type sequence :math:`t_0^\ast` must be empty.
```
rossberg: This isn't actually an assertion -- this rule is the very one that validates this property for results. ;)

document/core/appendix/properties.rst
```diff
@@ -58,8 +58,31 @@ Results
...
+
+* For each :ref:`value <syntax-val>` :math:`\val_i` in :math:`\val^\ast`:
```
rossberg:
```suggestion
* For each :ref:`value <syntax-val>` :math:`\val_i` in :math:`\val^\ast` and corresponding :ref:`value type <syntax-valtype>` :math:`t_i` in :math:`t_i`:
```

document/core/appendix/properties.rst
ioannad: Doesn't the first line (L.65) do this validation? 
 
 > * The :ref:`external value <syntax-externval>` :math:`\EVTAG~\tagaddr` must be :ref:`valid <valid-externval-tag>` with some :ref:`external type <syntax-externtype>` :math:`\ETTAG~\tagtype`.

ioannad: I guess that's the justification, so removing the "Assert:"

document/core/appendix/properties.rst
```diff
@@ -66,9 +66,9 @@ Results
 
 * Let :math:`[t^\ast]\to[t_0^\ast]` be the :ref:`tag type <syntax-tagtype>` |tagtype|.
 
-* Assert: due to :ref:`validation <valid-externval-tag>` the type sequence :math:`t_0^\ast` must be empty.
+* Due to :ref:`validation <valid-externval-tag>` the type sequence :math:`t_0^\ast` must be empty.
```
rossberg:
```suggestion
* The type sequence :math:`t_0^\ast` must be empty.
```
It's not due to some other validation, this rule _is_ the validation. ;)

rossberg: To expand on this, there are several layers:

* Externval-tag validity, which you referenced here, does not actually check that the referenced tag instance is valid (and this rule does not know if the store is valid).
* Tag instance validity (as part of store validity) checks that the tag type is valid.
* Tag type validity checks that the result is empty (but see below).

You can actually combine the first three bullets in this rule into one by saying something like "The extern value TAG tagaddr must be valid with some external type TAG [t^\ast]->[]." This style is already used in some of the other rules, I believe, e.g. THROWadm.

Finally, I noticed that we are currently a bit inconsistent: tag type validity requires that the result is empty, but we are simply assuming, not asserting that in various places of the prose (e.g.,  the validation rule for THROW). My preference for fixing this would be to _not_ treat tag types with empty results as invalid themselves, but only enforce emptiness in the exception-related use site rules.

ioannad: > Externval-tag validity, which you referenced here, does not actually check that the referenced tag instance is valid (and this rule does not know if the store is valid).

Thank you, I think I missed that point. I saw that [externval-tag validation](https://webassembly.github.io/exception-handling/core/valid/types.html#xref-syntax-types-syntax-tagtype-mathsf-tag-xref-syntax-types-syntax-tagtype-mathit-tagtype) does check if the tag is valid, and I forgot that it's not aware of the store's validity.

 I will implement your second suggestion to combine the first three bullets. This way it's shorter and consistent with THROWadm validation.

> Finally, I noticed that we are currently a bit inconsistent: tag type validity requires that the result is empty, but we are simply assuming, not asserting that in various places of the prose (e.g., the validation rule for THROW). My preference for fixing this would be to not treat tag types with empty results as invalid themselves, but only enforce emptiness in the exception-related use site rules.

I will create a new PR to make these changes, to not mix up the conversations.

ioannad: @rossberg 

> I will implement your second suggestion to combine the first three bullets. This way it's shorter and consistent with THROWadm validation.

Done.

rossberg (approved):

PR #260, opened 2023-02-17
Adds validation algorithm case for `throw x`, adds missing x to 'catch'

ioannad:


document/core/appendix/algorithm.rst
```diff
@@ -227,6 +227,11 @@ Other instructions are checked in a similar manner.
          error_if(frame.opcode =/= try || frame.opcode =/= catch)
          push_ctrl(catch_all, [], frame.end_types)
 
+       case (throw x)
+          let params = tags[x].type.params
+          pop.vals(params)
```
```suggestion
          pop.vals(tags[x].type.params)
```
And the same in the catch case.

rossberg (approved):

document/core/appendix/algorithm.rst
ioannad: Done for `catch` too.

PR #262, opened 2023-02-17
Merge from upstream spec

aheejin:


dschuff (approved):
LGTM, but IMO you don't really even need a review on this kind of change

aheejin:
Currently the CI has two failures. One is 'build-core-spec' failure, which is also reproduced in the spec repo (WebAssembly/spec#1603).

The other is 'ref-interpreter' fails on `binary.wast` (https://github.com/WebAssembly/exception-handling/actions/runs/4208770248/jobs/7305156280, which I'm not sure why. This failure is not reproduced in my local machine. This is the reason why I'm now trying various random changes and reverting (by force-pushing) them.

(I'm now trying to revert commits in the merge to figure out which commit is causing the issue. All these reverting commits will be re-reverted.)

aheejin:
It looks https://github.com/WebAssembly/exception-handling/pull/262/commits/31cabf87f8e605d71933d724430831a841d2d298 (from WebAssembly/spec#1595) is causing the 'ref-interpreter' failure (https://github.com/WebAssembly/exception-handling/actions/runs/4209147102/jobs/7305884276), but I haven't figured out why or why it's not causing the same problem in the spec repo.

aheejin:
I was able to reproduce the error on my machine now. The command is
```console
(from the interpreter dir)
$ ../test/core/run.py --wasm $PWD/wasm --js node
```
And the `node` should be v18, as https://github.com/WebAssembly/exception-handling/commit/31cabf87f8e605d71933d724430831a841d2d298 says. (My local node was a newer one, v19.5, which is the current LTS. The error was not reproduced on this version.) Haven't figured out why node v18 fails only on the EH repo (and not the spec repo) yet. I didn't even know we need `node` to run the core spec tests....

rossberg:
The test likely fails because node does not yet implement/expose exception handling. Until it does, you'll have to remove the `JS=node` on line 26 of .github/workflows/main.yml to deactivate running the test suite through node in CI.

I don't know why the spec build started failing, that was news to me.

aheejin:
@Ms2ger Sorry but I'm asking this for the third time now.. Can you please let me merge my own PRs, in the WebAssembly repos where I have write access? The ref-interpreter CI was failing and I was trying to fix it before merging.

aheejin:
> The test likely fails because node does not yet implement/expose exception handling. Until it does, you'll have to remove the `JS=node` on line 26 of .github/workflows/main.yml to deactivate running the test suite through node in CI.

Node started to support EH without a flag starting v17: https://webassembly.org/roadmap/
So it's a node issue but doesn't seem to be the version issue. I'm trying to figure out what the issue is. (Node v19 doesn't have the issue anymore)

aheejin:
@rossberg 
> The test likely fails because node does not yet implement/expose exception handling. Until it does, you'll have to remove the `JS=node` on line 26 of .github/workflows/main.yml to deactivate running the test suite through node in CI.

Node started to support EH without a flag starting v17: https://webassembly.org/roadmap/
So it's a node issue but doesn't seem to be the version issue. I'm trying to figure out what the issue is. (Node v19 doesn't have the issue anymore)

By the way I thought core spec tests are run by the reference interpreter... Are they run by node instead now? What happens if I comment out that `node` line? Are they run by the interpreter or not run?

aheejin:
@Ms2ger Sorry, but I'm asking this for the third time now. Can you please let me merge my own PRs, in the WebAssembly repos where I have write access? The ref-interpreter CI was failing and I was trying to fix it before merging. As I said in my previous comments as well, I prefer merging my PRs myself, for these reasons:
- When I ask review for multiple reviewers, I want to wait for some more time for other reviewers to comment so I can address them.
- There can be some CI errors I want to address before merging.
- I want to edit my commit messages and prefer that over an automatically generated commit message, which concatenates all my individual commits in that PR.

Given that almost all people who post PRs in this repo have write access anyway, can we leave PRs to be merged by authors, maybe at least in this repo?

rossberg:
> By the way I thought core spec tests are run by the reference interpreter... Are they run by node instead now? What happens if I comment out that `node` line? Are they run by the interpreter or not run?

They are run by both. The node run is to ensure that the JS test generation of the interpreter works correctly. If you don't set the js command then it will still convert the tests to JS but not test the result.

PR #266, opened 2023-02-24
[js-api] Add more tests reusing payload

aheejin:
As suggested in
https://github.com/WebAssembly/exception-handling/pull/256#discussion_r1109501365, this adds two more tests:
- JS throws an exception, Wasm catches it and returns the payload.
- JS throws an exception, Wasm catches it and throws a new exception using that payload.


aheejin:
Friendly ping.

dschuff (approved):

PR #274, opened 2023-04-21
[interpreter] Cherry-pick tail call support

aheejin:
This cherry-picks interpreter changes from
https://github.com/WebAssembly/tail-call to run spec tests that mix `try`s with `return_call(_indirect)`s. (#275) When tail-call is merged to the main spec repo, we can revert this change and merge the upstream spec.

This also adds missing handling for `ReturningInvoke` in the new `Catch`/`Caught`/`Delegate` instructions in the evaluator.

### DIFF
--- a/interpreter/binary/decode.ml
+++ b/interpreter/binary/decode.ml
@@ -303,8 +303,13 @@ let rec instr s =
     let y = at var s in
     let x = at var s in
     call_indirect x y
+  | 0x12 -> return_call (at var s)
+  | 0x13 ->
+    let y = at var s in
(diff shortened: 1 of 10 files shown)


thibaudmichaud (approved):

rossberg (commented):
The evaluator will also need to define the interaction between ReturningInvoke and the new Catch/Caught/Delegate instructions. I believe that should merely consist of adding ReturningInvoke to the enumerated cases in lines 721, 746, and 762.

aheejin:
Thanks!

PR #275, opened 2023-04-21
[test] Add tests for return_call(_indirect) in try

### LINKED ISSUE #249: Tail call + exception handling
`return_call` and related instructions from the tail-call proposal (now phase 4) unwind the current frame before calling the target.  So if the `return_call` is enclosed by a `try` block, I expect that any exception thrown by the callee cannot be caught by the `try` block:

```wasm
try
  return_call $fn
catch_all
  ;; Unreachable, this frame is gone so the exception is "rethrown" to the caller.
end
```

Does this match everyone's expectations? Should we add some spec text and tests to cover this case? Are there other ambiguous cases with tail-call+EH?

aheejin:
This adds tests of `return_call(_indirect)`s within `try`s. Because this repo's interpreter doesn't have the tail call support, this requires #274 to be merged before merging. But currently even with #274 the
tests here don't run. When a `return_call(_indirect)` runs within a `try`, it errors out with
```
runtime crash: undefined frame
```
I think we need to fix the interpreter first before merging this.

Closes #249.


thibaudmichaud (approved):

thibaudmichaud:
I would also suggest adding a test that return_calls an import, and maybe one that return_calls a host function in the JS tests.

rossberg (approved):
Yes, the crash is unsurprising and easily fixed, see my comment on #274.

.github/workflows/ci-interpreter.yml
```diff
@@ -31,4 +31,4 @@ jobs:
       - name: Build interpreter
         run: cd interpreter && opam exec make
       - name: Run tests
-        run: cd interpreter && opam exec make JS=node ci
+        run: cd interpreter && opam exec make JS='node --experimental-wasm-return_call' ci
```
aheejin, line 34: This was necessary to make the interpreter tests pass, because we not only run them with ocaml but also with JS. I think we can revert this later once the tail call is merged to the main spec and is enabled without a flag.

PR #278, opened 2023-04-26
[docs] Fix typo

harry900831:
If I understand correctly, enter the block with caught exception { $val^n$ } may push the value on stack so we don't have to enter the block $val^n$ $instr^*$.


harry900831:
It seems that I misunderstood the spec. Sorry for bothering, I'm closing this PR.

PR #283, opened 2023-10-14
[spec] Update spec for option B'

rossberg:
This updates the spec document to option B'.

@ioannad, would you be interested in reviewing this change? I know it is rather large, but it's deleting more than it's adding. :)


conrad-watt:
@rossberg I can probably take a look at this while I'm in the air on Sunday.

titzer (approved):
I'd be OK merging this PR into `exnref-1b` before review and then reviewing the merged branch as a PR on the main branch. (AFAICT there is no PR to merge `exnref-1b` into the main branch yet, and I already noticed a typo there).

rossberg:
> AFAICT there is no PR to merge exnref-1b into the main branch yet

There is: #285 ;)

Happy to do it in either order.

rossberg:
Thinking again, it probably makes sense not to block #285, so people see an up-to-date explainer and tests etc. I rather hold off on merging this one until @conrad-watt or somebody else did a proper review.

conrad-watt:
Apologies for the delay, on this now.

conrad-watt:
EDIT: I originally complained about a build warning here, but I was building on the wrong branch

document/core/appendix/index-instructions.py
```diff
@@ -103,14 +103,14 @@ def Instruction(name, opcode, type=None, validation=None, execution=None, operat
...
-    Instruction(r'\CATCHALL', r'\hex{19}', None, r'valid-try-catch', r'exec-try-catch'),
+    Instruction(None, r'\hex{18}'),
+    Instruction(None, r'\hex{19}'),
     Instruction(r'\DROP', r'\hex{1A}', r'[t] \to []', r'valid-drop', r'exec-drop'),
     Instruction(r'\SELECT', r'\hex{1B}', r'[t~t~\I32] \to [t]', r'valid-select', r'exec-select'),
     Instruction(r'\SELECT~t', r'\hex{1C}', r'[t~t~\I32] \to [t]', r'valid-select', r'exec-select'),
     Instruction(None, r'\hex{1D}'),
     Instruction(None, r'\hex{1E}'),
-    Instruction(None, r'\hex{1F}'),
+    Instruction(r'\TRYTABLE~\X{bt}', r'\hex{1F}', r'[t_1^\ast] \to [t_2^\ast]', r'valid-try_table', r'exec-try_table'),
```
conrad-watt, line 113: Are we no longer listing the `catch` opcodes separately?

document/core/appendix/properties.rst
```diff
@@ -59,16 +59,12 @@ Results
    }
 
 
-:ref:`Results <syntax-result>` :math:`\val^\ast~(\THROWadm~\tagaddr)`
+:ref:`Results <syntax-result>` :math:`\XT[(\REFEXNADDR~a)~\THROWREF]`
```
conrad-watt, line 62: It feels kind of nasty to classify an instruction inside a context as a "result". Can you walk me through why you don't want to use something like `\THROWadm` any more?

document/core/util/macros.def
```diff
@@ -1280,6 +1290,7 @@
 .. |vdashglobalinst| mathdef:: \xref{appendix/properties}{valid-globalinst}{\vdash}
```
conrad-watt, line 1290: line 1283 (above): `.. |vdashval| mathdef:: \xref{appendix/properties}{valid-val}{\vdash}`

should be `exec/modules` rather than `appendix/properties` (I think)

document/core/exec/instructions.rst
```diff
@@ -2683,123 +2683,184 @@ Control Instructions
...
-   \quad \stepto \\
-   \qquad F; \LABEL_n\{\epsilon\}~(\HANDLERadm_n\{(a_x~\instr_2^\ast)^\ast~(\epsilon~\instr_3^\ast)^?\}~\val^m~\instr_1^\ast~\END)~\END \\
-   (\iff \expand_F(\X{bt}) = [t_1^m] \to [t_2^n] \land (F.\AMODULE.\MITAGS[x]=a_x)^\ast)
+   \begin{array}{lclr@{\qquad}l}
+   S; F; \val^n~(\THROW~x) &\stepto& S'; F; (\REFEXNADDR~|S.\SEXNS|)~\THROWREF &
+     (\begin{array}[t]{@{}r@{~}l@{}}
+      \iff & F.\AMODULE.\MITAGS[x] = a \\
+      \land & S.\STAGS[a].\TAGITYPE = [t^n] \toF [] \\
+      \land & \X{exn} = \{ \EITAG~a, \EIFIELDS~\val^n \} \\
+      \land & S' = S \with \SEXNS = S.\SEXNS~\X{exn} ) \\
```
conrad-watt, line 2752: Is it necessary to allocate exceptions in the store? Since they're immutable, can the abstract semantics not just treat them as ordinary values?

document/core/appendix/index-instructions.py
rossberg: They are no longer in the space of instruction opcodes, but just flag bytes in a vector.

document/core/appendix/properties.rst
rossberg: The introduction of the context is actually unrelated to replacing THROWadm, its absence was simply wrong before. We need it here since there is no reduction rule that can eliminate it on the outermost level (outside a function). We can't simply add a rule XT[exn] ⇝ exn, since that would introduce overlapping rules.

I see two ways to fix this in the future: (1) switching to a bubbling semantics for exceptions (avoiding overlap), or (2) introducing a separate global reduction relation like we have with threads. The former should go along with changes to other parts of the semantics to stay coherent, the latter I'd rather do after we have merged with threads. Hence I don't want to do either in this PR.

document/core/exec/instructions.rst
rossberg: References need internal identity, because it can become observable in the host, e.g. in JS. This is analogous to functions, which are likewise immutable.

(The lack of exn identity was one of the problems we faced with the JS API spec in the old proposal.)

document/core/util/macros.def
rossberg: Indeed, fixed.

rossberg (commented):
Thanks for the review!

conrad-watt (approved):
In the longer term, I'd like to revisit the way the exception "value" is formalised, but beyond that things seem fine.

Issue #287, opened 2024-01-24
exception lifetime question


yamt:
reading the latest overview https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions.md
to me, it seems that some kind of GC is assumed to track the lifetime of an exception referenced by exnref.
is it right?

keithw:
I don't think so -- my understanding is that in a GC-less profile, exnref will be constrained so that runtimes can represent it as a copyable value type of reasonable size (a sum type / variant of any possible exception), rather than as a reference.

In particular, in the GC-less profile, it will be malformed for a tag type to include an exnref param, and there will be an implementation-defined limit on the number of parameters in a tag type (distinct from the [existing limit on the number of parameters in a function type](https://webassembly.github.io/spec/core/appendix/implementation.html#structure)) that could be set small enough so that a max-size serialized exception is small enough to keep as a value and copy as needed.

Discussion here: https://github.com/WebAssembly/exception-handling/issues/280#issuecomment-1664945711

yamt:
ok.
then, GC-less runtimes need to be able to handle huge values, right?
(at least far larger than v128)

keithw:
I think that's probably right, although the rules haven't been written down formally yet afaik. I think plausibly the spec could say that the implementation-defined limit is on the total size of values in a tag type (rather than the number of values), and a reasonable value for the limit could be 16 bytes (one v128, two i64's, or four i32's) without breaking real modules in practice.

titzer:
What @keithw said. Note that his suggestion is just one possible implementation strategy. It would also be possible to reference-count them, since they cannot form cycles. A tracing of GC would of course be a correct implementation. Since Wasm up-until-now has no other GC requirement, adding GC of just `exnref` could be done with a stack scan looking for roots, with no transitive closure required.

With Wasm GC being now Phase 4, the tracing option would only apply in the no-gc profile that will be later defined.

yamt:
>  It would also be possible to reference-count them, since they cannot form cycles.

my understanding is that, funcref can already form cycles and exnref can be a part of such cycles.
am i missing something?

yamt:
> I don't think so -- my understanding is that in a GC-less profile, exnref will be constrained so that runtimes can represent it as a copyable value type of reasonable size (a sum type / variant of any possible exception), rather than as a reference.
> 
[...]

PR #293, opened 2024-01-29
Mention a few more possible implementation limits

yamt:
cf. https://github.com/WebAssembly/exception-handling/issues/287#issuecomment-1908694098

### DIFF
--- a/document/core/appendix/implementation.rst
+++ b/document/core/appendix/implementation.rst
@@ -46,6 +46,8 @@ An implementation may impose restrictions on the following dimensions of a modul
 * the number of results in a :ref:`function type <syntax-functype>`
 * the number of parameters in a :ref:`block type <syntax-blocktype>`
 * the number of results in a :ref:`block type <syntax-blocktype>`
+* the number of parameters in a :ref:`tag type <syntax-tagtype>`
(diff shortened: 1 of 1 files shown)


document/core/appendix/implementation.rst
```diff
@@ -46,6 +46,8 @@ An implementation may impose restrictions on the following dimensions of a modul
 * the number of results in a :ref:`function type <syntax-functype>`
 * the number of parameters in a :ref:`block type <syntax-blocktype>`
 * the number of results in a :ref:`block type <syntax-blocktype>`
+* the number of parameters in a :ref:`tag type <syntax-tagtype>`
+* the total size of parameters in a :ref:`tag type <syntax-tagtype>`
```
rossberg: The "size" of parameters is not a well-defined notion for the spec, since e.g. reference types have no observable size. However, we can omit this item, since it is already subsumed by the number of parameters (for which the limit may vary with their types) – the same is already assumed for function parameters, locals, etc.

yamt: ok. i removed the item.

rossberg (approved):
Thanks!

PR #298, opened 2024-02-20
Add noexn type to overview

rossberg:
Explain how exnref generalises when combined with the enriched type system from the GC proposal.


aheejin (approved):
Thanks!

PR #300, opened 2024-03-26
Fix the description of the stack value for `catch` and `catch_all`

JSMonk:
Maybe I missed something, but it seems like the description of the stack value has a typo, and the `catch_all` should be instead of the `catch_ref`. If I'm wrong, feel free to close the PR 🤷‍♂️


rossberg:
No, the text is correct. Both catch and catch_ref push the exception arguments to the stack, the other two don't; catch_ref pushes _both_ arguments and exnref, catch_all neither.

JSMonk:
@rossberg , yeap, I've already re-read it and understand what I understood wrong. But thank you for the clarification 🙏

PR #303, opened 2024-05-23
Revive legacy JS API and restructure legacy spec directory

aheejin:
This adds `core` and `js-api` directory in `document/legacy/exceptions`, moves all core spec files into `core/`, and restores the deleted legacy JS API files in `js-api/`.

The core legacy spec only contains the EH instructions, which is easier to view, but it is hard to precisely carve out the modified part from the JS API file, so this adds the whole file for the legacy spec.


rossberg (approved):
Personally, I'd prefer keeping it as just "exceptions" (since it's shorter and because it's not just the catch mechanism). But I'm fine either way.

dschuff (approved):
LGTM, don't really have any opinion on the naming.

aheejin:
I'm not strongly opinionated on the directory name either. The main reason I renamed it was in case there will be other proposals part of which may end up in this `legacy` directory, it might be better if you conform to the original proposal names to prevent any confusion.... But I guess there will likely be no proposals going through a "journey" like this, after all? 🤷🏻

aheejin:
Will revert back to `exceptions` for now then.

rossberg:
Yeah, I sincerely hope this doesn't become a regular thing. :)  FWIW, I don't think proposal names are highly relevant for later spec matters, since they are historic and sometimes become inaccurate when a proposal evolves, see e.g. "bulk memory ops", "threads", etc.

PR #312, opened 2024-06-21
Fix typo in explainer

### LINKED ISSUE #310: Issues with the README
The README says this:

> In case of `catch` or `catch_ref`, the arguments of the exception are pushed back onto the stack. For `catch_ref` and `catch_all_ref`, an exception reference is then pushed to the stack, which represents the caught exception.

I may be wrong, but this just seemed inconsistent. I would assume it should say this:

> In case of `catch` or *`catch_all`*, the arguments of the exception are pushed back onto the stack. For `catch_ref` and `catch_all_ref`, an exception reference is then pushed to the stack, which represents the caught exception.

There's also a minor typo in the following line (it misspells *caught* as *cought*):

> All forms have a label which is branched to when an exception is cought (see below).

I didn't want to become a "contributor" for such a trivial pull request, so just opened an issue here instead.

aheejin:
Fixes #310.

### DIFF
--- a/proposals/exception-handling/Exceptions.md
+++ b/proposals/exception-handling/Exceptions.md
@@ -36,8 +36,8 @@ succeeding instructions to process the data.
 A WebAssembly exception is created when you throw it with the `throw`
 instruction. Thrown exceptions are handled as follows:
 
-1. They can be caught by one of the *catch clauses* in an enclosing try
-   block of a function body.
+1. They can be caught by one of the *catch clauses* in an enclosing try block of
+   a function body.
 
 1. Throws not caught within a function body continue up the call stack, popping
    call frames, until an enclosing try block is found.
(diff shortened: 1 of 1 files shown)


proposals/exception-handling/Exceptions.md
```diff
@@ -164,10 +164,10 @@ after possible block parameters were popped.
 
 Then catch clauses are tried in the order they appear in the catching try block,
 until one matches. If a matching catch clause is found, control is transferred
-to the label of that catch clause. In case of `catch` or `catch_ref`, the
-arguments of the exception are pushed back onto the stack. For `catch_ref` and
-`catch_all_ref`, an exception reference is then pushed to the stack, which
-represents the caught exception.
+to the label of that catch clause. In case of `catch`, the arguments of the
```
sjrd: This change is not correct. Now the phrasing says that `catch_ref` pushes *only* the ref. But it pushes the arguments *and* the ref.

aheejin: You're right; I was confused and the original text was correct.. Will revert this part.

rossberg (approved):

PR #313, opened 2024-06-21
[js-api] Update wasm-module-builder.js for exnref

aheejin:
This pulls in some exnref-related changes from
https://github.com/v8/v8/blob/main/test/mjsunit/wasm/wasm-module-builder.js but does not sync with it completely. That v8 version contains a lot of features from new proposals that are not really relevant to EH and exnref, so this pulls only the relevant EH parts in.

This also renames `kWasmStmt` to `kWasmVoid` as in the V8 version, because it looks more intuitive. Also renames `kTagAttribute` to `kExceptionAttribute`.

### DIFF
--- a/test/js-api/wasm-module-builder.js
+++ b/test/js-api/wasm-module-builder.js
@@ -90,8 +90,8 @@ let kDeclFunctionImport = 0x02;
 let kDeclFunctionLocals = 0x04;
 let kDeclFunctionExport = 0x08;
 
-// Local types
-let kWasmStmt = 0x40;
+// Value types and related
+let kWasmVoid = 0x40;
 let kWasmI32 = 0x7f;
 let kWasmI64 = 0x7e;
 let kWasmF32 = 0x7d;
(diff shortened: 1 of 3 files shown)


test/js-api/wasm-module-builder.js
```diff
@@ -217,10 +221,9 @@ let kExprIf = 0x04;
 let kExprElse = 0x05;
 let kExprTry = 0x06;
 let kExprCatch = 0x07;
-let kExprCatchAll = 0x19;
```
aheejin, line 220: These old instructions were not deleted; they were just moved below to maintain the sorted order of opcodes.

(This is not actually the same as what v8 file does; it refactored the old opcodes into a list, which I didn't pull in here: https://github.com/v8/v8/blob/8b52e5f026dc859cab827ec584f3eedbc0510459/test/mjsunit/wasm/wasm-module-builder.js#L261-L463. But the sorted order is maintained in this list too.)

aheejin:
Sorry, will reopen this PR in the upstream branch (to make stacked PRs)
Reopened as #314.

PR #321, opened 2024-07-04
[interpreter] Fix crashes on try_table with parameters

yamt:
The test case was ported from toywasm.


rossberg (approved):
Oh, thanks!

This proposal could definitely use better test coverage.

yamt:
> Oh, thanks!
> 
> This proposal could definitely use better test coverage.

maybe we can have a copy of block.wast with `s/block/try_table/g` as try_table w/o catch clauses is semantically identical to block.

PR #329, opened 2024-09-12
Correct definition of instruction block execution semantics

CharlieTap:
I found this whilst implementing the spec, I could tell something was off as the definition of exiting the instruction block mentions you pop the label first and then the handler which was inconsistent.


rossberg (approved):
Thanks!

PR #330, opened 2024-09-13
Fix ThrowRef execution semantics

CharlieTap:
This is another one I found whilst implementing the spec. A rather confusing one to be honest 😓 

Effectively if you take the frame at the beginning of the instruction, you may be executing inside an imported function and thus have a different module instance in your stack frame. As part of finding the handler, you unwind the stack and exit this imported function but you are still referencing its module instance. When you then perform the check that attempts to match the tag you would  be looking up tag within the wrong module instance.

This test helped me find it:

```wat
(module
  (func $imported-throw (import "test" "throw"))
  (tag $e0)

  (func (export "imported-mismatch") (result i32)
    (block $h
      (try_table (result i32) (catch_all $h)
        (block $h0
          (try_table (result i32) (catch $e0 $h0)
            (i32.const 1)
            (call $imported-throw)
          )
          (return)
        )
        (i32.const 2)
      )
      (return)
    )
    (i32.const 3)
  )
)
```

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -2757,57 +2757,57 @@ Control Instructions
 :math:`\THROWREF`
 .................
 
-1. Let :math:`F` be the :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>`.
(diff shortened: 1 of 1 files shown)


rossberg (approved):
Good catch!

PR #332, opened 2024-09-22
[legacy] Add setjmp bait

SoniEx2:
sadly UBSAN doesn't appear to detect this (yet).

### DIFF
--- a/test/legacy/exceptions/core/try_catch.wast
+++ b/test/legacy/exceptions/core/try_catch.wast
@@ -175,6 +175,20 @@
   (func (export "break-try-catch_all")
     (try (do (br 0)) (catch_all))
   )
+
+  (func $longjmp-bait (throw $e0))
+  (func (export "setjmp-bait") (param $return-early i32) (result i32)
+    (local $value i32)
+    (try $try
+      (do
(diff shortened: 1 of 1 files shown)


rossberg:
Shouldn't we have a corresponding test for the non-legacy instruction?

SoniEx2:
yes, eventually

wasm2c only supports EHL at this point so having the test for EHv4 isn't that critical yet.

SoniEx2:
surprisingly, this test also appears to break wabt interpreter.

for context, the interpreter isn't even using setjmp.

aheejin:
Sorry I may lack some context, but what is this test about? They say `setjmp` and `longjmp` but here they are just normal functions. Does this test have any specific characteristics or does this just happen to be the case the wabt interpreter fails on? Apparenly this PR doesn't have any changes to the spec interpreter so that is fine, right?

SoniEx2:
the test targets wasm2c specifically, where exceptions are implemented using setjmp/longjmp, but also happens to trigger a bug in wabt interpreter (which is mostly unrelated to wasm2c, despite also being part of wabt).

the wabt interpreter is unrelated to the spec interpreter, the spec interpreter is probably fine.

PR #334, opened 2024-10-07
Handle negative type codes in wasm-module-builder.js

thibaudmichaud:
This is a follow-up from:
https://github.com/WebAssembly/exception-handling/commit/901a9b0e33501535691707fe79115b14de9f457c

In particular `kWasmAnyFunc` was emitted as 0xf0 instead of 0x70 because of the missing Leb encoding, and the `module/exports` test failed in V8 because of this.

### DIFF
--- a/test/js-api/wasm-module-builder.js
+++ b/test/js-api/wasm-module-builder.js
@@ -981,7 +981,7 @@ class WasmModuleBuilder {
             section.emit_u32v(imp.initial); // initial
             if (has_max) section.emit_u32v(imp.maximum); // maximum
           } else if (imp.kind == kExternalTable) {
-            section.emit_u8(imp.type);
+            section.emit_type(imp.type);
             var has_max = (typeof imp.maximum) != "undefined";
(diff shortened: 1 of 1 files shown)


Ms2ger (approved):
Oh no. Could you propagate to the base spec repo too?

thibaudmichaud:
The base spec is fine, it still uses the plain leb-encoded type code:
https://github.com/WebAssembly/spec/blob/9c69d19f429fb3766fe60cd63accf7578fe16235/test/js-api/wasm-module-builder.js#L101

Ms2ger:
Oh, and it doesn't have `emit_type` at all.

aheejin (approved):

aheejin:
Hmm, we now use `main` for the main branch. Can you merge this into `main` branch too? Or are `main` and `master` synced automatically? (I think `master` just catches up to `main`, but I'm not really sure if the other way around happens)

thibaudmichaud:
Sorry about that, I've reopened a PR for the main branch.

